Issue #8, opened 2023-12-07
Implement receiving new EC-compatible chains


anorth:
The protocol should update its view of incentive-compatible chains that it will switch to while executing, if proposed by another node. There's a design question whether this should be pushed in from the host, or pulled from within. Pull-based will avoid any concurrency etc. but push-based might be easier for the host.

A simple decision might be to follow whatever pattern is used for F3 learning about new chains in the first place, for proposing as values to finalise.

At least keep the concurrency out of GossipPBFT, but it might be ok for the containing F3 implementation to deal with potentially-concurrent pushes.

anorth:
Following discussion with @Kubuxu, a likely path here is that the internal GossipBFT will pull from the F3 layer when it needs these, but that F3 layer may receive them as pushes from the node.

This issue is about the internal bit, so
- Add a new ReceiveECChain method to the ECReceiver interface
- Check it extends the input
- Use this longer chain in `isAcceptable()`

anorth:
Closed by #19

PR #19, opened 2023-12-15
Implement receiving new EC-compatible chains

ranchalp:
Solves #8

### DIFF
--- a/f3/api.go
+++ b/f3/api.go
@@ -5,6 +5,8 @@ type ChainReceiver interface {
 	// Receives a chain appropriate for use as initial proposals for a Granite instance.
 	// The chain's base is assumed to be an appropriate base for the instance.
 	ReceiveCanonicalChain(chain ECChain, power PowerTable, beacon []byte)
+
+	ReceiveECChain(chain ECChain)
 }
 
 // A consensus message.
--- a/f3/api.go
+++ b/f3/api.go
@@ -6,6 +6,8 @@ type ChainReceiver interface {
 	// The chain's base is assumed to be an appropriate base for the instance.
 	ReceiveCanonicalChain(chain ECChain, power PowerTable, beacon []byte)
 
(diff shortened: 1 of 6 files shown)


f3/api.go
```diff
@@ -5,6 +5,8 @@ type ChainReceiver interface {
 	// Receives a chain appropriate for use as initial proposals for a Granite instance.
 	// The chain's base is assumed to be an appropriate base for the instance.
 	ReceiveCanonicalChain(chain ECChain, power PowerTable, beacon []byte)
+
```
anorth, line 8: Please add documentation

f3/participant.go
```diff
@@ -49,6 +49,12 @@ func (p *Participant) ReceiveCanonicalChain(chain ECChain, power PowerTable, bea
 	}
 }
 
+func (p *Participant) ReceiveECChain(chain ECChain) {
+	if p.granite != nil && chain.HasPrefix(p.granite.input) {
+		p.granite.heaviestChain = chain
```
anorth, line 54: Please don't mutate the granite instance externally. Instead, add a method to receive a new chain and enforce the prefix logic there.

The predicate isn't right. It must check that the current heaviest acceptable chain is a prefix, not just the input. Otherwise this can switch chains mid-execution.

f3/granite.go
```diff
@@ -73,6 +73,8 @@ type instance struct {
 	// State for each round of phases.
 	// State from prior rounds must be maintained to provide justification for values in subsequent rounds.
 	rounds map[int]*roundState
+	// Heaviest acceptable chain
+	heaviestChain ECChain
```
anorth, line 77: I would just call this "acceptable". The granite code never looks at the weight or compares it to another chain. The "heaviest" nature is something external. For this code, it's simply a predicate of what it will accept.

ranchalp:
Thanks for the comments! I have just addressed them.

anorth (approved):

PR #28, opened 2023-12-20
VRF implementation (#6)

### LINKED ISSUE #6: Implement non-fake VRF ticket generation and validation
Tickets are currently fake. Implement real ones based on signatures. There should be prior art for this in the Lotus codebase.

Depends on #4 for signatures.

ranchalp:
Closes #6

### DIFF
--- a/f3/vrf.go
+++ b/f3/vrf.go
@@ -2,6 +2,7 @@ package f3
 
 import (
 	"bytes"
+	"encoding/binary"
 	"fmt"
 )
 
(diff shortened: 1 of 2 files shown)


ranchalp:
(Fixed lint error)

anorth (commented):
If the real VRF is cheap enough to evaluate (with fake sigs) to use in simulation then I think we can/should just delete FakeVRF entirely. So please do that.

ranchalp:
Done, thanks!

anorth (approved):

PR #29, opened 2023-12-22
Decide phase

### LINKED ISSUE #2: Implement the DECIDE phase with strong quorum
The protocol is to be updated by making the DECIDE phase mandatory, completing only after receiving a strong quorum of DECIDE messages.

- Add the DECID phase, moving the decision point of the instance until after a strong quorum of such messages are gathered
- Re-use the existing quorum state helpers

matejpavlovic:
Resolves #2.

Implements the DECIDE phase in Granite.
As created, this PR is incomplete. The round numbers of the DECIDE messages should be consolidated. Currently, 0 is always output as the decision round.

### DIFF
--- a/f3/granite.go
+++ b/f3/granite.go
@@ -103,6 +103,10 @@ type instance struct {
 	rounds map[uint32]*roundState
 	// Acceptable chain
 	acceptable ECChain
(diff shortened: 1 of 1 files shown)


ranchalp:
Continued in #30. This is deprecated.

PR #30, opened 2024-01-03
Decide phase

### LINKED ISSUE #2: Implement the DECIDE phase with strong quorum

ranchalp:
Extending #29, (new PR as I have no pushing rights on @matejpavlovic 's fork). From @matejpavlovic :
> Resolves #2 .
> 
> Implements the DECIDE phase in Granite.
> As created, this PR is incomplete. The round numbers of the DECIDE messages should be consolidated. Currently, 0 is always output as the decision round (which is the most probable reason for the tests failing).

### DIFF
--- a/f3/granite.go
+++ b/f3/granite.go
@@ -329,6 +329,9 @@ func (i *instance) isJustified(msg *GMessage) bool {
 		// COMMIT for bottom is always justified.
 		round := i.roundState(msg.Round)
 		return msg.Value.IsZero() || round.prepared.HasStrongQuorumAgreement(msg.Value.HeadCIDOrZero())
+	} else if msg.Step == DECIDE {
+		// DECIDE needs no justification
+		return !msg.Value.IsZero()
 	}
 	return false
 }
--- a/f3/granite.go
+++ b/f3/granite.go
@@ -492,7 +492,7 @@ func (i *instance) tryDecide(value ECChain) {
 	}
 
 	if strongQuorum {
-		i.decide(value, 0)
+		i.decide(value, i.round)
 	}
 }
 
--- a/f3/granite.go
+++ b/f3/granite.go
@@ -264,7 +264,7 @@ func (i *instance) receiveOne(msg *GMessage) {
 	// Every COMMIT phase stays open to new messages even after the protocol moves on to
 	// a new round. Late-arriving COMMITS can still (must) cause a local decision, *in that round*.
 	// DECIDE messages are also independent of current phase or round.
-	if msg.Step == COMMIT {
+	if msg.Step == COMMIT && i.phase != DECIDE {
 		i.tryCommit(msg.Round)
 	} else if msg.Step == DECIDE {
 		i.tryDecide(msg.Value)
--- a/f3/granite.go
+++ b/f3/granite.go
@@ -452,7 +452,7 @@ func (i *instance) tryCommit(round uint32) {
 	timeoutExpired := i.host.Time() >= i.phaseTimeout
 
 	if len(foundQuorum) > 0 && !foundQuorum[0].IsZero() && i.phase != DECIDE && i.phase != TERMINATED {
-		// A participant may be forced to terminate a value that's not its preferred chain.
+		// A participant may be forced to decide a value that's not its preferred chain.
 		// The participant isn't influencing that decision against their interest, just accepting it.
 		i.value = foundQuorum[0]
 		i.beginDecide()
--- a/f3/granite.go
+++ b/f3/granite.go
@@ -459,7 +459,7 @@ func (i *instance) tryCommit(round uint32) {
 	} else if i.round == round && i.phase == COMMIT && timeoutExpired && committed.ReceivedFromStrongQuorum() {
 		// Adopt any non-empty value committed by another participant (there can only be one).
 		// This node has observed the strong quorum of PREPARE messages that justify it,
-		// and mean that some other nodes may terminate that value (if they observe more COMMITs).
+		// and mean that some other nodes may decide that value (if they observe more COMMITs).
 		for _, v := range committed.ListAllValues() {
 			if !v.IsZero() {
 				if !i.isAcceptable(v) {
--- a/f3/granite.go
+++ b/f3/granite.go
@@ -103,6 +103,10 @@ type instance struct {
 	rounds map[uint32]*roundState
 	// Acceptable chain
 	acceptable ECChain
+	// Decision state. Collects DECIDE messages until a decision can be made, independently of protocol phases/rounds.
+	decision *quorumState
+	// Flag set when the participant sends a DECIDE message. Used to send it only once.
+	decideSent bool
 }
 
 func newInstance(
@@ -136,6 +140,8 @@ func newInstance(
 			0: newRoundState(powerTable),
 		},
 		acceptable: input,
+		decision:   newQuorumState(powerTable),
+		decideSent: false,
 	}
 }
 
@@ -246,6 +252,8 @@ func (i *instance) receiveOne(msg *GMessage) {
 		round.prepared.Receive(msg.Sender, msg.Value)
 	case COMMIT:
 		round.committed.Receive(msg.Sender, msg.Value)
+	case DECIDE:
+		i.decision.Receive(msg.Sender, msg.Value)
 	default:
 		i.log("unexpected message %v", msg)
 	}
@@ -253,8 +261,11 @@ func (i *instance) receiveOne(msg *GMessage) {
 	// Try to complete the current phase.
 	// Every COMMIT phase stays open to new messages even after the protocol moves on to
 	// a new round. Late-arriving COMMITS can still (must) cause a local decision, *in that round*.
+	// DECIDE messages are also independent of current phase or round.
 	if msg.Step == COMMIT {
 		i.tryCommit(msg.Round)
+	} else if msg.Step == DECIDE {
+		i.tryDecide(msg.Value)
 	} else {
 		i.tryCompletePhase()
 	}
@@ -435,10 +446,11 @@ func (i *instance) tryCommit(round uint32) {
 	foundQuorum := committed.ListStrongQuorumAgreedValues()
 	timeoutExpired := i.host.Time() >= i.phaseTimeout
 
-	if len(foundQuorum) > 0 && !foundQuorum[0].IsZero() {
+	if len(foundQuorum) > 0 && !foundQuorum[0].IsZero() && !i.decideSent {
 		// A participant may be forced to decide a value that's not its preferred chain.
 		// The participant isn't influencing that decision against their interest, just accepting it.
-		i.decide(foundQuorum[0], round)
+		i.decideSent = true
+		i.broadcast(DECIDE, foundQuorum[0], nil)
 	} else if i.round == round && i.phase == COMMIT && timeoutExpired && committed.ReceivedFromStrongQuorum() {
 		// Adopt any non-empty value committed by another participant (there can only be one).
 		// This node has observed the strong quorum of PREPARE messages that justify it,
@@ -457,6 +469,25 @@ func (i *instance) tryCommit(round uint32) {
 
 		}
 		i.beginNextRound()
+	}
+}
+
+func (i *instance) tryDecide(value ECChain) {
+	weakQuorum := i.decision.HasWeakQuorumAgreement(value.HeadCIDOrZero())
+	strongQuorum := i.decision.HasStrongQuorumAgreement(value.HeadCIDOrZero())
+
+	// If a weak quorum of DECIDE messages has been received, at least one correct participant must have sent one
+	// (and there exists a proof of it, albeit not necessarily locally present).
+	// That means that this value must be decided by every correct participant.
+	// Broadcasting a DECIDE message is necessary for liveness if the broadcast primitive is only best-effort broadcast.
+	// If broadcast() implemented reliable broadcast, this step would not be necessary.
+	if weakQuorum && !i.decideSent {
+		i.broadcast(DECIDE, value, nil)
+		i.decideSent = true
+	}
+
+	if strongQuorum {
+		i.decide(value, 0)
 	}
 }
 
--- a/f3/granite.go
+++ b/f3/granite.go
@@ -264,7 +264,7 @@ func (i *instance) receiveOne(msg *GMessage) {
 	// Try to complete the current phase.
 	// Every COMMIT phase stays open to new messages even after the protocol moves on to
 	// a new round. Late-arriving COMMITS can still (must) cause a local decision, *in that round*.
-	if msg.Step == COMMIT && i.phase != TERMINATED {
+	if msg.Step == COMMIT && i.phase != DECIDE {
 		i.tryCommit(msg.Round)
 	} else {
 		i.tryCompletePhase()
--- a/f3/granite.go
+++ b/f3/granite.go
@@ -223,6 +223,10 @@ func (i *instance) tryPendingMessages() {
 
 // Processes a single message.
 func (i *instance) receiveOne(msg *GMessage) {
+	if i.phase == TERMINATED {
+		return // No-op
+	}
+
 	// Drop any messages that can never be valid.
 	if !i.isValid(msg) {
 		i.log("dropping invalid %s", msg)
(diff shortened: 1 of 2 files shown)


ranchalp:
I considered multiple options for the round of DECIDE. Ultimately I decided to dedicate a special InstanceRound field that is only used for debugging purposes and is always set to i.round. Other options I considered: 

1. Directly checking if message step is step == DECIDE and then setting the round = 0 only for the signing process. But this adds logic to the broadcast function.

2. Create a new function broadcastDecide that is called only when broadcasting DECIDE and that does the same as 1 (set round = 0 for the signature only)

3. Refactor broadcast by extracting two new aux. functions, sign and broadcastGMessage, and have either the same as (2.) but calling the auxiliary functions, or directly calling them at the point where a decide is sent. 

I think what we have currently is the cleanest as it has no special treatment of DECIDE messages during broadcast (other than when calling the broadcast function the round parameter of the function is set to 0 instead of i.round), but of course it incurs an additional field of the `GMessage` struct.

ranchalp:
Also, I am not sure what we settled on returning as the round number now that we established that ‘decision round’ is a vague concept. I set it to the current instance round in which the process decides (as I am assuming this is useful purely for debugging purposes as well).

f3/granite.go
```diff
@@ -246,15 +254,20 @@ func (i *instance) receiveOne(msg *GMessage) {
...
 	}
 
 	// Try to complete the current phase.
 	// Every COMMIT phase stays open to new messages even after the protocol moves on to
 	// a new round. Late-arriving COMMITS can still (must) cause a local decision, *in that round*.
+	// DECIDE messages are also independent of current phase or round.
 	if msg.Step == COMMIT {
 		i.tryCommit(msg.Round)
+	} else if msg.Step == DECIDE {
+		i.tryDecide(msg.Value)
```
anorth, line 270: This case here, and the implementation of tryDecide, are mixing the Bracha amplification in with the core state transitions, which confuses things. Please separate them. Do the amplification from a separate place in this method, independent of the core state machine. I'm not convinced we will want/use the amplification if we switch to explicit evidence, but the rest of the logic should stay the same.

- Add `i.tryDecide` down at 289. It should have no parameter but follow the same pattern as `tryCommit`.
- Add a new `i.maybeAmplify(msg.Value)` call here if `msg.step == DECIDE` (but not an _else_ if - should still `tryCompletePhase()`).

f3/granite.go
```diff
@@ -424,22 +440,23 @@ func (i *instance) tryPrepare() {
...
 	// Unlike all other phases, the COMMIT phase stays open to new messages even after an initial quorum is reached,
 	// and the algorithm moves on to the next round.
 	// A subsequent COMMIT message can cause the node to decide, so there is no check on the current phase.
 	committed := i.roundState(round).committed
-	foundQuorum := committed.ListQuorumAgreedValues()
+	foundQuorum := committed.ListStrongQuorumAgreedValues()
 	timeoutExpired := i.host.Time() >= i.phaseTimeout
 
-	if len(foundQuorum) > 0 && !foundQuorum[0].IsZero() {
+	if len(foundQuorum) > 0 && !foundQuorum[0].IsZero() && !i.decideSent {
```
anorth, line 454: This `!decideSent` is not in the algorithm [here](https://hackmd.io/YDW-1jd7Tzy8pp2kb6T7zQ). It's not obvious what to do here - the Bracha amplification introduces a problem. Intuitively a strong quorum of COMMIT is better than a weak quorum of DECIDE, but here it is losing out (obviously with some adversarial equivocation)!

Leaving this condition here will prevent the phase transition resulting from the strong quorum of commits. For now I think it would be better to remove the condition, but perhaps file an issue for it. (One alternative to note in the issue: we could push the condition down into the `beginDecide` method, allowing the state transition but not the broadcast. This would put a node in a weird state where it can't contribute to the strong quorum of DECIDE for the value it COMMITTED, and is waiting for a strong quorum of DECIDE for a different value).

I hope we can just remove the amplification.

f3/granite.go
```diff
@@ -460,6 +477,25 @@ func (i *instance) tryCommit(round uint32) {
...
+	// That means that this value must be decided by every correct participant.
+	// Broadcasting a DECIDE message is necessary for liveness if the broadcast primitive is only best-effort broadcast.
+	// If broadcast() implemented reliable broadcast, this step would not be necessary.
+	if weakQuorum && !i.decideSent {
+		i.broadcast(DECIDE, value, nil, 0)
+		i.decideSent = true
+	}
+
+	if strongQuorum {
+		i.decide(value, i.round)
```
anorth, line 495: We might want to rename this method  since "deciding" (entering the DECIDE phase) is no longer the terminating step, which is what this method is about. Perhaps `complete()` or `terminate()`?

f3/granite.go
```diff
@@ -460,6 +477,25 @@ func (i *instance) tryCommit(round uint32) {
 	}
 }
 
+func (i *instance) tryDecide(value ECChain) {
```
anorth, line 480: As noted above, this method is coupling the state machine with amplification. Please separate them so this method can follow the same pattern as all the other state transitions. It should have no parameter, but be a pure function of state.

anorth (commented):
Thanks, this is close, but I think should be extended to implement the DECIDE phase more fully as an equal of the others. This means adding a new phase name, e.g. COMPLETED to represent the termination of the algorithm, and then entering/exiting the DECIDE phase as part of the state machine (even though it's permanently open until termination).

The Bracha amplification complicates things. I would be happy if you removed it and deferred to be implemented later (create an issue for it). If we do explicit validation, we probably won't need it. But if you want to leave it in, please fully decouple it from the state transition logic.

f3/granite.go
ranchalp: A strong quorum of COMMIT of any round (and noting that all honest nodes send a COMMIT before they move on to the next round) is actually as good as a weak quorum of DECIDE without rounds (and noting that honest nodes *only* send DECIDE when they have received a strong quorum of COMMITs). 

Strictly speaking they are just as good: an adversary controlling less than a weak quorum cannot fake either and an adversary controlling a weak quorum or more can fake both. 

However, even without Bracha, I think `!i.decideSent` makes sense to prevent nodes from broadcasting the same DECIDE twice (even without bracha they may send it after receiving a strong quorum of DECIDE, which may be delivered before a strong quorum of COMMIT). 

The modification of the algorithm that includes i.decideSent (and the Bracha amplification) is the one in the [PR of the FIP](https://github.com/filecoin-project/FIPs/blob/9ad97ce63e4d37b60ed9077ecc0941a08b521e39/FIPS/fip-xxxx.md?plain=1#L407). AFAIK this is the most up to date version of the pseudocode and the only one that is currently being updated with new comments.

ranchalp: Actually, since we are repurposing `i.phase = DECIDE` and introducing `i.phase = TERMINATED`, we can instead use `i.phase != DECIDE && i.phase != TERMINATED` and completely forget about `i.decideSent`. (Done in the upcoming pushed changes).

f3/granite.go

f3/granite.go
ranchalp: And actually `i.phase != TERMINATED` is only necessary for clarity, as tryCommit will only happen if `i.phase != TERMINATED` in the first place with the current code. Happy to remove it but leaving it for clarity for now.

f3/granite.go
ranchalp: done (`terminate()`)

f3/granite.go

ranchalp:
Thanks for the comments! Pushed my attempt at resolving them :).

f3/granite.go
```diff
@@ -43,6 +43,8 @@ type GMessage struct {
 	Ticket Ticket
 	// Signature by the sender's public key over Instance || Round || Step || Value.
 	Signature []byte
+	// InstanceRound is only used for debugging purposes, to check the round in which the process decided, as DECIDE messages set the round field to 0 for aggregation
+	InstanceRound uint32
```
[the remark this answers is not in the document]
anorth: Yes, if the tests are passing I think it's ok to forget about the instance round. If we need it, put it in state (and update tests to look for it there).

f3/granite.go
anorth: Right, i.decideSent is redundant because the phase transitions ensure each messages is only sent once. Defensive programming of extra, redundant checks will confuse things.

f3/granite.go
```diff
@@ -246,14 +250,21 @@ func (i *instance) receiveOne(msg *GMessage) {
...
+	// Bracha amplification for DECIDE phase (instead of justifying DECIDE messages)
+	if msg.Step == DECIDE {
+		i.maybeAmplify(msg.Value)
+	}
+
 	// Try to complete the current phase.
 	// Every COMMIT phase stays open to new messages even after the protocol moves on to
 	// a new round. Late-arriving COMMITS can still (must) cause a local decision, *in that round*.
-	if msg.Step == COMMIT {
+	if msg.Step == COMMIT && i.phase != TERMINATED {
```
anorth, line 267:
```suggestion
	if msg.Step == COMMIT && i.phase != DECIDE {
```

Please move the `TERMINATED` check to the top of `receiveOne`, and just return as a no-op. This is possible if the inbox has more than one message in it, and the first one results in a termination. Otherwise, receiving a new message while TERMINATED already results in a rejection.

But here we still need a `!= DECIDE` check. If this node has entered DECIDE phase because it has observed a strong quorum of COMMIT, it can't make a phase change based on more of them so we should not process it.

f3/granite.go
```diff
@@ -264,6 +275,8 @@ func (i *instance) receiveOne(msg *GMessage) {
 func (i *instance) tryCompletePhase() {
 	i.log("try step %s", i.phase)
 	switch i.phase {
+	case TERMINATED:
+		return // No-op
```
anorth, line 279: Nit: move to be the last case, following their sequence.

f3/granite.go
```diff
@@ -424,25 +440,26 @@ func (i *instance) tryPrepare() {
...
 	i.phaseTimeout = i.alarmAfterSynchrony(PREPARE)
-	i.broadcast(COMMIT, i.value, nil)
+	i.broadcast(i.round, COMMIT, i.value, nil)
 }
 
 func (i *instance) tryCommit(round uint32) {
 	// Unlike all other phases, the COMMIT phase stays open to new messages even after an initial quorum is reached,
 	// and the algorithm moves on to the next round.
-	// A subsequent COMMIT message can cause the node to decide, so there is no check on the current phase.
+	// A subsequent COMMIT message can cause the node to terminate, so there is no check on the current phase.
```
anorth, line 449:
```suggestion
	// A subsequent COMMIT message can cause the node to decide, so there is no check on the current phase.
```

A bit too happy with the find-and-replace. Do you use an IDE for such renames?

f3/granite.go
```diff
@@ -424,25 +440,26 @@ func (i *instance) tryPrepare() {
...
 	committed := i.roundState(round).committed
-	foundQuorum := committed.ListQuorumAgreedValues()
+	foundQuorum := committed.ListStrongQuorumAgreedValues()
 	timeoutExpired := i.host.Time() >= i.phaseTimeout
 
-	if len(foundQuorum) > 0 && !foundQuorum[0].IsZero() {
-		// A participant may be forced to decide a value that's not its preferred chain.
+	if len(foundQuorum) > 0 && !foundQuorum[0].IsZero() && i.phase != DECIDE && i.phase != TERMINATED {
```
anorth, line 454: Both of these new conditions are redundant. They are confusing because the reader is now left working out why they are here, how the code might get into such a state. Please remove.

f3/granite.go
```diff
@@ -424,25 +440,26 @@ func (i *instance) tryPrepare() {
...
+		// A participant may be forced to terminate a value that's not its preferred chain.
```
anorth, line 455:
```suggestion
		// A participant may be forced to decide a value that's not its preferred chain.
```

f3/granite.go
```diff
@@ -424,25 +440,26 @@ func (i *instance) tryPrepare() {
...
 		// The participant isn't influencing that decision against their interest, just accepting it.
-		i.decide(foundQuorum[0], round)
-	} else if i.round == round && i.phase == COMMIT && timeoutExpired && committed.ReceivedFromQuorum() {
+		i.value = foundQuorum[0]
+		i.beginDecide()
+	} else if i.round == round && i.phase == COMMIT && timeoutExpired && committed.ReceivedFromStrongQuorum() {
 		// Adopt any non-empty value committed by another participant (there can only be one).
 		// This node has observed the strong quorum of PREPARE messages that justify it,
-		// and mean that some other nodes may decide that value (if they observe more COMMITs).
+		// and mean that some other nodes may terminate that value (if they observe more COMMITs).
```
anorth, line 462:
```suggestion
		// and mean that some other nodes may decide that value (if they observe more COMMITs).
```

f3/granite.go
```diff
@@ -460,6 +477,32 @@ func (i *instance) tryCommit(round uint32) {
 	}
 }
 
+// maybeAmplify implements Bracha amplification:
+// If a weak quorum of DECIDE messages has been received, at least one correct participant must have sent one
+// (and there exists a proof of it, albeit not necessarily locally present).
+// That means that this value must be terminated by every correct participant.
+// Broadcasting a DECIDE message is necessary for liveness if the broadcast primitive is only best-effort broadcast.
+// If broadcast() implemented reliable broadcast, this step would not be necessary.
+func (i *instance) maybeAmplify(value ECChain) {
```
anorth, line 486: Nit: please move this method lower down so it does not interrupt the sequence of try/beginPhase methods.

f3/granite.go
```diff
@@ -460,6 +477,32 @@ func (i *instance) tryCommit(round uint32) {
...
+	weakQuorum := i.decision.HasWeakQuorumAgreement(value.HeadCIDOrZero())
+	if weakQuorum && i.phase != DECIDE && i.phase != TERMINATED {
+		i.value = value
+		i.beginDecide()
```
anorth, line 490: I don't agree that receiving a DECIDE message should update the node's value nor cause a state transition into the DECIDE phase. The FIP PR does not do this either. This should be _only_ Bracha amplification.

Taking that as an assumption, the phase checks here are unwanted, but you will need some `decideSent` state to avoid a rebroadcast. The complexity of https://github.com/filecoin-project/go-f3/pull/30#discussion_r1445459148 will still be present, though. Could you please add a comment when decideSent is set true about the implications for the node arriving at its own decision from COMMITs?

Again, I hope we can just delete this after explicit validation.

anorth (commented):
Looking good except for the Bracha amplification.. If you want to bump that to a follow-up PR, it'll be easy to land the rest of this one.

f3/granite.go
ranchalp: Using GoLand these days, though I have a background as an emacs lover.

f3/granite.go
ranchalp: As a matter of fact. We can still keep DECIDE messages completely unjustified (not verifying any justification), and still remove the Bracha amplification. 

The reasoning is that eventually all (honest) nodes will receive enough COMMITs to send their decide, enter then the DECIDE phase, and then they can just check the locally delivered (unjustified) DECIDE messages to check for a strong quorum. 

Therefore, I have completely remove `i.decideSent` and `i.maybeAmplify`, but still left DECIDE with no justification whatsoever. Delivered DECIDE messages are immediately saved in quorumState but only when the node receives enough COMMITs to transit its state to DECIDE will it even check for the received DECIDE (and no Byzantine behavior changes anything because we check for a strong quorum of DECIDE).

ranchalp:
Addressed the comments! You must be happy that we can remove Bracha amplification with very minor implications to the code. It was really just kind of an alternative path to termination (but an unnecessary one and not even necessarily a faster-path).

f3/granite.go
ranchalp: EDIT: We can forget about justifying DECIDE messages now that we have implicit justification and assume reliable broadcast. Without assuming reliable broadcast then we would need explicit justification and justify DECIDE messages (no bracha).

f3/granite.go
```diff
@@ -665,22 +702,28 @@ func (q *quorumState) HasAgreement() bool {
...
-// Checks whether a chain (head) has reached quorum.
-func (q *quorumState) HasQuorumAgreement(cid TipSetID) bool {
+// Checks whether a chain (head) has reached a strong quorum.
+func (q *quorumState) HasStrongQuorumAgreement(cid TipSetID) bool {
+	cp, ok := q.chainPower[cid]
+	return ok && cp.hasStrongQuorum
+}
+
+// Checks whether a chain (head) has reached weak quorum.
+func (q *quorumState) HasWeakQuorumAgreement(cid TipSetID) bool {
```
anorth, line 716: This is now unused, but I guess we can leave it here for a while until we're sure it won't be needed.

anorth (approved):
Great, thank you.

PR #31, opened 2024-01-03
Explicit message justification (Opt: generalization of quorumState)

### LINKED ISSUE #16: Implement signed evidence-based message justification
Implement the specified BLS-aggregate evidence-based version of the protocol. Messages should be justified just once and either processed or dropped. ~Expect constant memory complexity.~

We might hold this code in a branch while we continue to evaluate which justification method to use.

Requires #4 for signatures.

ranchalp:
Resolves #16.

Incomplete, and so far mostly created just to allow for discussion of design decisions.


ranchalp:
More refactoring to do, but explicit and implicit both work by now.

f3/granite.go
```diff
@@ -43,6 +43,33 @@ type GMessage struct {
...
+type AggEvidence struct {
+	// Enumeration of QUALITY, PREPARE, COMMIT, CONVERGE, DECIDE
+	Step string
+	// Chain of tipsets proposed/voted for finalisation in this instance.
+	// Non-empty: the first entry is the base tipset finalised in instance-1
+	Value ECChain
+	// GossiPBFT instance number
+	Instance uint32
+	// GossiPBFT round
+	Round uint32
```
anorth, line 60: Put these in the same order as they are in GMessage. Remove the copy-pasta comments.

f3/granite.go
```diff
@@ -43,6 +43,33 @@ type GMessage struct {
...
+	// Indexes in the base power table of the signers (bitset)
+	Signers BitSet
+	// BLS aggregate signature of signers
+	Signature []byte
+}
+
+func ZeroAggEvidence() AggEvidence {
+	return AggEvidence{}
+}
```
anorth, line 69: I don't think we need this method if the canonical Go zero value is what we want.

f3/granite.go
```diff
@@ -43,6 +43,33 @@ type GMessage struct {
...
+
+func (a AggEvidence) isZero() bool {
+	return a.Step == "" && a.Value.IsZero() && a.Instance == 0 && a.Round == 0 && a.Signers.Size() == 0 && len(a.Signature) == 0
```
anorth, line 72: Just check for `a == AggEvidence{}`. It's debatable whether you need this method, since the caller can just do that.

f3/granite.go
```diff
@@ -728,6 +1039,99 @@ func (c *convergeState) findMinTicketProposal() ECChain {
 	return minValue
 }
 
+// BitSet is a bitset implementation.
+type BitSet struct {
```
anorth, line 1043: Please import and use https://github.com/filecoin-project/go-bitfield/blob/master/bitfield.go instead

f3/granite.go
```diff
@@ -728,6 +1039,99 @@ func (c *convergeState) findMinTicketProposal() ECChain {
...
+	for i := uint64(0); i < b.size; i++ {
+		if b.IsSet(i) {
+			callback(i)
+		}
+	}
+}
+
+// SenderIndex maps ActorID to a unique index in the range [0, len(powerTable.Entries)).
+// This is used to index into a BitSet. index2Actor is the reverse mapping.
+type SenderIndex struct {
```
anorth, line 1104: Rather than this we probably want to build a stronger PowerTable type that is an ordered map, so we can just look up entries by index. The host should build it for us. I think defining and implementing that should be a PR of its own that you can do before this one.

f3/granite.go
```diff
@@ -728,6 +1039,99 @@ func (c *convergeState) findMinTicketProposal() ECChain {
...
+	}
+	keys := make([]ActorID, 0, len(table.Entries))
+
+	// Extract keys from the map
+	for senderID := range table.Entries {
+		keys = append(keys, senderID)
+	}
+
+	// Sort the keys
+	sort.Slice(keys, func(i, j int) bool { return keys[i] < keys[j] })
```
anorth, line 1122: This probably isn't the ordering we will want - probably want descending by power instead.

anorth (commented):
This interface-based way of doing both at once here has introduced a lot of code, some copy paste, and plumbing of values around more than necessary. The message justificaition check should be performed at the same place the implicit justification is checked. Everything after that should just be able to assume justified messages.

I'm still not that confident that supporting both at once in the codebase is reasonable. This method is not one that's just an easy removal of a single call and the block of code that it calls – it's had a large impact on the existing code structure.

This has identified the need for an improved PowerTable type, though, so let's break that out into a PR ahead of this one.

f3/granite.go
ranchalp: AggEvidence contains fields that make it not comparable with `==` such as `Signers` (bitfield.Bitfield), `Signature` ([] byte) and `Value` (ECChain, which is `[]Tipset`).

sim/network.go
```diff
@@ -135,6 +136,62 @@ func (n *Network) Verify(sender f3.ActorID, msg, sig []byte) bool {
 	return true
 }
 
+func (n *Network) Aggregate(sig []byte, actorID f3.ActorID, aggSignature []byte, signers *f3.BitSet, actor2Index map[f3.ActorID]uint64) ([]byte, *f3.BitSet) {
+	// Fake implementation.
+	// Just appends signature to aggregate signature.
+	// This fake aggregation is not commutative (order matters), unlike the real one.
```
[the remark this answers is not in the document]
ranchalp: The problem with XOR is that the verifier would have to somehow get the individual signatures in order to verify the aggregate, changing the interface of the `aggregator` w.r.t. the non-fake implementation. I have instead ensured determinism of the aggregate and this determinism is checked in the `verifyAggregate`, which is essentially equivalent for the purposes of the code outside the aggregator (the resulting signature is the same for everyone from aggregating and the verify only is correct if the signature is the same). 

Left a comment in the Aggregate function.

f3/granite.go
ranchalp: done.

f3/granite.go
ranchalp: done.

sim/network.go
```diff
@@ -135,6 +136,62 @@ func (n *Network) Verify(sender f3.ActorID, msg, sig []byte) bool {
...
+	if signers.Size() < actor2Index[actorID] {
+		//TODO failure (return false or panic?)
+		panic("signers size is less than actor2Index")
+	}
+
+	signers.Set(actor2Index[actorID])
```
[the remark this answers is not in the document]
ranchalp: Done (in dedicated PR #35 )

PR #32, opened 2024-01-10
Implement PowerTable as ordered map

ranchalp:
EDIT: builds upon #34 , so merge that one first.

Implements `PowerTable` as an ordered map. To be used by explicit justification to create/verify bitsets (see [here](https://github.com/filecoin-project/go-f3/pull/31#discussion_r1446689822)) and, even if using implicit justification, by the creation/verification of PoFs as part of the catch-up mechanism.


f3/powertable.go
```diff
@@ -1,22 +1,81 @@
...
+	}
+}
+
+// Add adds a new entry to the PowerTable. It is less efficient than NewPowerTableFromMap,
+// as it needs to sort the entries after each addition, but allows adding entries one by one.
 func (p *PowerTable) Add(id ActorID, power uint) {
-	if p.Entries[id] != 0 {
+	if _, ok := p.Lookup[id]; ok {
 		panic("duplicate power entry")
 	}
```
anorth, line 59: You can remove this, because you can check for duplication after doing the `sort.Search`.

f3/powertable.go
```diff
@@ -1,22 +1,81 @@
...
+	ID    ActorID
+	Power uint
+}
+
+// PowerTable maps ActorID to a unique index in the range [0, len(powerTable.Entries)).
+// Entries is the reverse mapping to a PowerEntry.
 type PowerTable struct {
-	Entries map[ActorID]uint
+	Entries []PowerEntry    // Slice to maintain the order
+	Lookup  map[ActorID]int // Map for quick index lookups
```
anorth, line 16: Please be more clear about what this is indexing. "Maps actor ID to the index of the associated entry in Entries"

f3/powertable.go
```diff
@@ -1,22 +1,81 @@
...
+
+	return PowerTable{
+		Entries: entries,
+		Lookup:  lookup,
+		Total:   total,
+	}
+}
+
+// Add adds a new entry to the PowerTable. It is less efficient than NewPowerTableFromMap,
+// as it needs to sort the entries after each addition, but allows adding entries one by one.
```
anorth, line 55:
```suggestion
// Add adds a new entry to the PowerTable. 
// This is linear in the table size, so adding many entries this way is inefficient.
```

Which is ok because this is only for testing.

f3/powertable.go
```diff
@@ -1,22 +1,81 @@
...
+	p.Entries[index] = entry
+
+	p.Lookup[id] = index
+	for i := index + 1; i < len(p.Entries); i++ {
+		p.Lookup[p.Entries[i].ID] = i
+	}
 	p.Total += power
 }
+
+func (p *PowerTable) GetPower(id ActorID) uint {
```
anorth, line 76: Suggest making this `Get(id ActorID) (Power, []byte)` and return both the power and private key, when you add that. And maybe even add the `ok bool` return here too (otherwise add a `Has()` method and panic here if !ok)

anorth (commented):
We really need two things here:
1. A data structure that can be serialized as the canonical power table, giving a CID
2. A runtime structure for looking up power values and keys

This is good for (2), but we won't want to serialise the lookup table. For now lets ignore that problem. We can chose later to add a new type for the serializable form, or manually implement serialisation here, including only the wanted data.

f3/granite.go
```diff
@@ -44,6 +45,24 @@ type GMessage struct {
 	Ticket Ticket
 	// Signature by the sender's public key over Instance || Round || Step || Value.
 	Signature []byte
+
+	Evidence AggEvidence
```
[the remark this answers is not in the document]
ranchalp: Done by rebasing from #34

f3/powertable.go
```diff
@@ -1,22 +1,81 @@
...
+type PowerEntry struct {
+	ID    ActorID
+	Power uint
+}
+
+// PowerTable maps ActorID to a unique index in the range [0, len(powerTable.Entries)).
+// Entries is the reverse mapping to a PowerEntry.
 type PowerTable struct {
-	Entries map[ActorID]uint
+	Entries []PowerEntry    // Slice to maintain the order
```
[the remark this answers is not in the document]
ranchalp: It's actually in order by (Power descending, ID ascending), but done otherwise.

f3/powertable.go
```diff
@@ -1,22 +1,81 @@
 package f3
 
-// A power table maps participant IDs to power values.
+import "sort"
+
+// PowerEntry represents a single entry in the PowerTable.
+// It includes an ActorID and its Weight
+type PowerEntry struct {
+	ID    ActorID
+	Power uint
```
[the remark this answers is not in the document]
ranchalp: Done.

f3/powertable.go
```diff
@@ -1,22 +1,81 @@
...
+}
```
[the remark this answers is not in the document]
ranchalp: done

f3/powertable.go
```diff
@@ -1,22 +1,81 @@
...
-		Entries: map[ActorID]uint{},
-		Total:   0,
+		Entries: make([]PowerEntry, 0),
+		Lookup:  make(map[ActorID]int),
 	}
 }
 
+// NewPowerTableFromMap creates a new PowerTable from a map of ActorID to weight.
+// It is more efficient than Add, as it only needs to sort the entries once.
+func NewPowerTableFromMap(entriesMap map[ActorID]uint) PowerTable {
```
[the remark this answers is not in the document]
ranchalp: I changed it to `[]PowerEntry`.

f3/powertable.go
ranchalp: Done.

f3/powertable.go
ranchalp: Done.

f3/powertable.go
ranchalp: Done.

ranchalp:
Thanks! addressed the comments :)

f3/granite.go
```diff
@@ -651,29 +563,36 @@ func (q *quorumState) Receive(sender ActorID, value ECChain) {
...
+	strongThreshold := new(StoragePower).Mul(q.powerTable.Total, two)
+	strongThreshold.Div(strongThreshold, three)
+	if candidate.power.Cmp(strongThreshold) > 0 {
 		candidate.hasStrongQuorum = true
 	}
 
-	weakThreshold := q.powerTable.Total * 1 / 3
-	if candidate.power > weakThreshold {
+	one := NewStoragePower(1)
+	weakThreshold := new(StoragePower).Mul(q.powerTable.Total, one)
```
anorth, line 593: No need to multiply by one

f3/granite.go
```diff
@@ -703,7 +622,13 @@ func (q *quorumState) HasAgreement() bool {
...
 // Checks whether at least one message has been received from a strong quorum of senders.
 func (q *quorumState) ReceivedFromStrongQuorum() bool {
-	return q.sendersTotalPower > q.powerTable.Total*2/3
+	two := NewStoragePower(2)
+	three := NewStoragePower(3)
+
+	threshold := new(StoragePower).Mul(q.powerTable.Total, two)
+	threshold.Div(threshold, three)
+
+	return q.sendersTotalPower.Cmp(threshold) > 0
```
anorth, line 631: Can you factor out a `hasStrongQuorum(part, total *StoragePower) bool` pure function to use here and above? And maybe do a `hasWeakQuorum` too for symmetry.

f3/granite.go
```diff
@@ -608,21 +520,21 @@ type quorumState struct {
 	// The power supporting each chain so far.
 	chainPower map[TipSetID]chainPower
 	// Total power of all distinct senders from which some chain has been received so far.
-	sendersTotalPower uint
+	sendersTotalPower StoragePower
```
anorth, line 523: It's tricky to know how to reference Go big.Ints properly. Is there a good reason this one is not a pointer but the others are? Using a pointer seems to be canonical so I'd suggest doing that here too, unless there's a good reason not to.

f3/powertable.go
```diff
@@ -1,22 +1,75 @@
...
+	Total   *StoragePower
 }
 
-func NewPowerTable() PowerTable {
-	return PowerTable{
-		Entries: map[ActorID]uint{},
-		Total:   0,
+// NewPowerTable creates a new PowerTable from a slice of PowerEntry .
+// It is more efficient than Add, as it only needs to sort the entries once.
+func NewPowerTable(entries []PowerEntry) *PowerTable {
```
anorth, line 25: Please note the function takes ownership of the slice - it must not be modified afterwards.

f3/powertable.go
```diff
@@ -1,22 +1,75 @@
...
+	sort.Slice(entries, func(i, j int) bool {
+		return entries[i].Power.Cmp(entries[j].Power) > 0 ||
```
anorth, line 27: Extract the repeated `entries[i].Power.Cmp(entries[j].Power)` to a temp.

f3/powertable.go
```diff
@@ -1,22 +1,75 @@
...
 
-func (p *PowerTable) Add(id ActorID, power uint) {
-	if p.Entries[id] != 0 {
+// Add adds a new entry to the PowerTable.
+// This is linear in the table size, so adding many entries this way is inefficient.
+func (p *PowerTable) Add(id ActorID, power *StoragePower, pubKey []byte) {
+	entry := PowerEntry{ID: id, Power: power, PubKey: pubKey}
+	index := sort.Search(len(p.Entries), func(i int) bool {
+		return p.Entries[i].Power.Cmp(power) > 0 ||
+			(p.Entries[i].Power.Cmp(power) == 0 && p.Entries[i].ID < id)
```
anorth, line 51: Actually please factor out this whole comparison function so it can be re-used between methods.

f3/powertable.go
```diff
@@ -1,22 +1,75 @@
...
+	})
+
+	// Check for duplication at the found index or adjacent entries
+	if index < len(p.Entries) && (p.Entries[index].ID == id || (index > 0 && p.Entries[index-1].ID == id)) {
```
anorth, line 55: You only need to check the exact entry pointed to.

> Search uses binary search to find and return the **smallest** index i in [0, n) at which f(i) is true

f3/powertable.go
```diff
@@ -1,22 +1,75 @@
...
+		return entries[i].Power.Cmp(entries[j].Power) > 0 ||
+			(entries[i].Power.Cmp(entries[j].Power) == 0 && entries[i].ID < entries[j].ID)
+	})
+
+	lookup := make(map[ActorID]int, len(entries))
+	total := NewStoragePower(0)
+	for i, entry := range entries {
+		lookup[entry.ID] = i
+		total.Add(total, entry.Power)
+	}
```
anorth, line 36: Check the map size here to ensure there were no duplicates in the input.

f3/powertable.go
```diff
@@ -1,22 +1,75 @@
...
+	p.Entries[index] = entry
+
+	p.Lookup[id] = index
+	for i := index + 1; i < len(p.Entries); i++ {
+		p.Lookup[p.Entries[i].ID] = i
+	}
+	p.Total.Add(p.Total, power)
+}
+
+func (p *PowerTable) Get(id ActorID) (*StoragePower, []byte, bool) {
```
anorth, line 70: Having seen this in use (especially ignoring the bool return), I have a further suggestion.

The existing code used the zero value map lookup to silently accept messages from participants with no power. Now is the time to change that. I recommend:
- Add a `Has(id)` method here, and call it in `isValid` to reject messages from participants absent or with zero power
- Remove the bool return here and just panic if not found. All the call sites should be able to rely on the prior filtering of valid messages.

f3/types.go
```diff
@@ -1,11 +1,50 @@
 package f3
 
 import (
+	"math/big"
 	"strings"
 )
 
 type ActorID uint64
 
+type StoragePower big.Int
```
anorth, line 10: Make this a type alias and then you can remove all the wrapper methods.

`type StoragePower  = big.Int`

This is slightly less "safe" in that it doesn't distinguish between different things a big.Int might be representing, but I think it's fine here (and Lotus does too).

sim/sim.go
```diff
@@ -33,12 +33,12 @@ func NewSimulation(simConfig Config, graniteConfig f3.GraniteConfig, traceLevel
...
 
 	// Create participants.
-	genesisPower := f3.NewPowerTable()
+	genesisPower := f3.NewPowerTable(make([]f3.PowerEntry, 0))
 	participants := make([]*f3.Participant, simConfig.HonestCount)
 	for i := 0; i < len(participants); i++ {
 		participants[i] = f3.NewParticipant(f3.ActorID(i), graniteConfig, ntwk, vrf)
 		ntwk.AddParticipant(participants[i])
-		genesisPower.Add(participants[i].ID(), 1)
+		genesisPower.Add(participants[i].ID(), f3.NewStoragePower(1), make([]byte, 0))
```
anorth, line 41: Let's put a non-empty key in here so that the fake signature verification at least checks it's the right signature. Serialize the participant ID into a byte array, and maybe add some magic bytes to distinguish it from other serialized integers.

f3/granite.go
ranchalp: Oops, AI help was too obvious here 😅

f3/granite.go
ranchalp: done.

f3/granite.go
ranchalp: You are absolutely right. I started with no pointers but missed this one when switching to pointers.

ranchalp: Done

f3/powertable.go
ranchalp: Correct, but not yet an issue. I reckon we leave as is?

f3/powertable.go
ranchalp: done.

f3/powertable.go
ranchalp: Done.

f3/types.go
ranchalp: Ah ok, I thought about this too but I did not do it precisely to distinguish between different things. But agreed that helps remove boilerplate. Done.

f3/powertable.go
ranchalp: Done.

sim/sim.go
ranchalp: Done.

ranchalp:
Just addressed the latest comments

f3/powertable.go
```diff
@@ -1,22 +1,101 @@
...
+	Power  *StoragePower
+	PubKey []byte
+}
+
+// PowerTable maps ActorID to a unique index in the range [0, len(powerTable.Entries)).
+// Entries is the reverse mapping to a PowerEntry.
 type PowerTable struct {
-	Entries map[ActorID]uint
-	Total   uint
+	Entries []PowerEntry    // Slice to maintain the order. Meant to be maintained in order in order by (Power descending, ID ascending)
```
Kubuxu, line 19: Q: Why do we build the power table sorted power?
Usage-wise, we will almost always query it by ActorID; events SPs appearing and disappearing are also rarer than an order of the power changing (especially in the long tail).

In general, it seems to me that having the power table have a different representation for granite format and different in the wire format (finality certificates) makes sense. Especially since we will need to optimize the wire format for usage in smart contracts.

Kubuxu: Ahh, ok, I saw it was also part of granite. It needs to communicate which participants signed it.
One option could be communicating that in the form of RLE+ bitfield of ActorIDs, keeping the power table format specifics out of the granite itself.
https://github.com/filecoin-project/go-bitfield/

f3/granite.go
anorth: Remember to replace the code here with a call to your new method

f3/powertable.go
anorth: I see what you're saying. Note that the TipSet structure here will have a PowerTable CID added to it, which must match the serialized format used in certificates. However that CID will come from outside, attached to the tipsets pushed in.

The signers will certainly be an RLE+ bitfield (it's present in subsequent PRs). A bitfield over powertable indices will be more dense than over ActorIDs, and I think the density does matter here, as bytes in the messages (of which there are quadratic in the participant count). The sorting by power is also a heuristic to increase density.

These are good questions, lets defer to an issue if you still think it's worth consideration. We can move on here with this, and come back to revise this structure if warranted.

f3/powertable.go
anorth: Sorry, by note I meant "write a comment"

go.mod
```diff
@@ -2,9 +2,18 @@ module github.com/filecoin-project/go-f3
...
 require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/ipfs/go-cid v0.0.5 // indirect
+	github.com/minio/blake2b-simd v0.0.0-20160723061019-3f5f724cb5b1 // indirect
+	github.com/minio/sha256-simd v0.1.1-0.20190913151208-6de447530771 // indirect
+	github.com/mr-tron/base58 v1.1.3 // indirect
+	github.com/multiformats/go-base32 v0.0.3 // indirect
+	github.com/multiformats/go-multibase v0.0.1 // indirect
+	github.com/multiformats/go-multihash v0.0.13 // indirect
+	github.com/multiformats/go-varint v0.0.5 // indirect
```
anorth, line 16: It looks like there's some noise crept in here, please normalise before merging.

anorth (approved):
LGTM after nits

f3/powertable.go
ranchalp: We have not gotten there yet (See TODOs in #37's description), but having them sorted by power makes it easy for the creator of a message to optimize message size when creating a justification (i.e. the creator needs to just iterate until the strong quorum is reached when creating the justification). Not visible at the moment even in #37 because all received values that justify the same message are added to the justification (so in many cases this will be more added signatures than required for a strong quorum).

Also the cost of having a sorted power table is negligible compared to running Granite.

f3/granite.go
ranchalp: Good catch! Done (upcoming commit)

go.mod
ranchalp: Done.

f3/powertable.go
ranchalp: Done

ranchalp:
Addressed comments from today :)

f3/powertable.go
Kubuxu: Not delaying this PR due to this, as @anorth let's defer this discussion to when we get to finality certificate

PR #34, opened 2024-01-12
Remove implicit justification

ranchalp:


f3/granite.go
```diff
@@ -319,17 +320,14 @@ func (i *instance) isJustified(msg *GMessage) bool {
 		if msg.Round == 0 || msg.Value.IsZero() {
 			return false
 		}
-		prevRound := i.roundState(msg.Round - 1)
-		return prevRound.prepared.HasStrongQuorumAgreement(msg.Value.Head().CID) ||
-			prevRound.committed.HasStrongQuorumAgreement(ZeroTipSetID())
+		return true
```
anorth, line 323: The comments on this method at 311 and some cases like 317 and 328 are now out of date. They're abstractly correct, but don't match the code so are confusing here. To be even more clear, how about:
- Move the validity tests (Round and Value.IsZero for some phases) to isValid
- Change this method to just `return true` (to be replaced by explicit justification later)

f3/granite.go
```diff
@@ -44,6 +45,24 @@ type GMessage struct {
 	Ticket Ticket
 	// Signature by the sender's public key over Instance || Round || Step || Value.
 	Signature []byte
+
+	Evidence AggEvidence
```
anorth, line 49: Can you please remove this from this PR and add it only when it will be used? That should revert dependency changes too.

anorth (commented):
Looks good, but adding the evidence here is confusing things.

f3/granite.go
ranchalp: Done!

f3/granite.go
ranchalp: Done

ranchalp:
Thanks! Addressed the comments :)

go.mod
```diff
@@ -2,10 +2,26 @@ module github.com/filecoin-project/go-f3
 
 go 1.20
 
-require github.com/stretchr/testify v1.8.2
+require (
+	github.com/filecoin-project/go-bitfield v0.2.4
```
anorth, line 6: Please delete this requirement before merging, and ensure the indirect requirements and go.sum are reverted too.

anorth (approved):

go.mod
ranchalp: Done. Merging.

PR #35, opened 2024-01-12
Add Aggregator interface and implement fake aggregator

ranchalp:
Builds upon #32, which in turn builds upon #34, so merge those first.


f3/api.go
```diff
@@ -51,11 +51,19 @@ type Signer interface {
 	Verify(sender ActorID, msg, sig []byte) bool
 }
 
+type Aggregator interface {
+	// Aggregates signatures from a participant to an existing signature.
+	Aggregate(sig []byte, aggSignature []byte) []byte
+	// VerifyAggregate verifies an aggregate signature.
+	VerifyAggregate(payload, aggSig []byte, signers map[ActorID]struct{}) bool
```
anorth, line 58: This API seems to assume that the implementor knows how to get a public key for an ActorID somewhere. It works for the fake impl, but I don't think will for the real one. I think instead `signers` should just be a list of public keys, so that it becomes a pure function.

ranchalp: Done.

ranchalp:
Thanks again! addressed the comment

f3/api.go
matejpavlovic: Actually, I think it would be most natural for aggregate signature itself to encode the list of signers (e.g., as a bitmask with respect to the publicly known power table).

Kubuxu: f3 will need to know about public keys either way, as it needs to store them in the power table/state. Yes, we could abstract pubkeys out in this interface, but then it means that the implementer of the API has to have access to the context-sensitive (per granite instance) power table

ranchalp: Implemented the verification of the proper public keys as listed [suggested in a previous PR](https://github.com/filecoin-project/go-f3/pull/32#discussion_r1453955206)

anorth (commented):
Let's merge the Signer.Verify method into the Aggregator interface (but rename it to SignatureVerifier), and then the SIgner interface only does actual signing (in non-fake this will delegate all the way out to Lotus).

Change the signing scheme to be concat the pubkey to the message, and verification (which should take the pubkey as a parameter instead of ActorID) to check that concatenation.

For aggregation, output payload ++ pubkeys (you can read the pubkey out of the input signature). Verify the new signature is unique before aggregation (error if not). Then verifying the aggregate is easy because the parameters are exactly payload and pubkeys.

There is coupling between the signing algorithm and verification, but that's the case for real signatures too!

ranchalp:
@Kubuxu new commits in the PR removed code from previous commits. Probably much better to check files changed instead of reviewing commit by commit.

sim/network.go
```diff
@@ -107,31 +111,94 @@ func (n *Network) SetAlarm(sender f3.ActorID, payload string, at float64) {
...
+				} else if err != nil {
+					panic(err) // Error in reading the signature.
+				}
+			}
+			pubKeys = append(pubKeys, existingPubKey)
+		}
+	}
+
+	for i := 0; i < len(sigs); i++ {
+		if !reflect.DeepEqual(msg, sigs[i][pubKeyLen:]) {
```
Kubuxu, line 164: Use `bytes.Equals`

sim/network.go
```diff
@@ -107,31 +111,94 @@ func (n *Network) SetAlarm(sender f3.ActorID, payload string, at float64) {
...
+		return bytes.Compare(signers[i], signers[j]) > 0
+	})
+
+	signersConcat := make([]byte, 0)
+	for _, signer := range signers {
+		signersConcat = append(signersConcat, signer...)
 	}
-	return true
+
+	aux := append([]byte(nil), payload...)
```
Kubuxu, line 199:
```suggestion
	aux := append([]byte{}, payload...)
```

sim/network.go
```diff
@@ -107,31 +111,94 @@ func (n *Network) SetAlarm(sender f3.ActorID, payload string, at float64) {
...
+	aux := append([]byte(nil), pubKey...)
+	aux = append(aux, msg...)
+	return bytes.Equal(aux, sig)
+}
+
+func (n *Network) Aggregate(sigs [][]byte, aggSignature []byte) []byte {
+	// Fake implementation.
+	// Just appends signature to aggregate signature.
+	// This fake aggregation is not commutative (order matters)
+	// But determinism is preserved by sorting by weight both here and in VerifyAggregate
```
Kubuxu, line 131: IMO: the deterministic sorting should happen outside the interface to match real signature schemes which can also have this behaviour.

f3/api.go
```diff
@@ -47,15 +47,23 @@ type Clock interface {
...
 	// Signs a message for the given sender ID.
 	Sign(sender ActorID, msg []byte) []byte
+}
+
+type Verifier interface {
 	// Verifies a signature for the given sender ID.
-	Verify(sender ActorID, msg, sig []byte) bool
+	Verify(pubKey PubKey, msg, sig []byte) bool
+	// Aggregates signatures from a participant to an existing signature.
+	Aggregate(sig [][]byte, aggSignature []byte) []byte
```
Kubuxu, line 56: I see that you are assuming that aggregate signatures can be combined. I think the conclusion is that it is not true, but we can fix that later.

sim/network.go
ranchalp: But that would mean forcing outside the interface a sorting that may not be required by *all* signature schemes. Would it not be better to have it inside the signature schemes that need it only?

ranchalp: (I get optimizing if the real BLS is not commutative, and coupling in that particular instantiation of the interface, but I also mean in particular in the case of the sorting required by this general-purpose (or so-intending) fake aggregation and verification scheme, that just happens to need to sort)

Kubuxu: I see your point, and I don't know which approach is better at this moment. I mentioned it because this requirement was unclear to me, and I immediately highlighted it as a possible difference between test/sim and reality.

We can keep order independence internal, but then we should highlight it in the interface documentation.

ranchalp: > We can keep order independence internal, but then we should highlight it in the interface documentation.

Sounds good, I suggest going back to this when we discuss interfaces in #42 not to duplicate discussions.

Kubuxu (approved):

PR #37, opened 2024-01-15
Add explicit justification of COMMIT and CONVERGE

ranchalp:
This PR Adds explicit justification, to be merged after #35 .

`quorumState` needs to store signatures (hard requirement). I also added a bitfield per tipsetID so that it easily identifies (and sorts by weight) the signers of a value (soft requirement). 

Additionally, `quorumState` needs to store messages that justify a value, for the case when a message triggers a value for the next step but the actual values that justify that message (and the message of the next step) have not been received outside the justification (hard requirement). An example is receiving a COMMIT with PREPARE in the COMMIT's evidence, but not having received the individual PREPAREs themselves. The participant may need then to use the same justification from the COMMIT directly in the CONVERGE.

TODO: - stop adding signatures once enough power has been found to reach the threshold.  
EDIT: the TODO is addressed in #50


f3/granite.go
```diff
@@ -606,23 +673,26 @@ type quorumState struct {
...
+	power uint //TODO this is unnecessary since quorumState has powerTable. Remove.
 }
 
-// A chain value and the total power supporting it.
-type chainPower struct {
+// A chain value and the total power supporting it
+type chainSupport struct {
 	chain           ECChain
 	power           uint
+	signers         bitfield.BitField
```
anorth, line 695: It'll probably be easier to understand if this just a `map[ActorID]{}` here. Or `map[int]`, one of the problems is I'm not sure what the values represent. `Bitfield` just has that internally anyway. We can build a Bitfield for serialization when necessary.

f3/granite.go
```diff
@@ -44,6 +45,24 @@ type GMessage struct {
 	Ticket Ticket
 	// Signature by the sender's public key over Instance || Round || Step || Value.
 	Signature []byte
+
+	Evidence AggEvidence
+}
+
+// Aggregated list of GossiPBFT messages with the same instance, round and value. Used as evidence for justification of messages
+type AggEvidence struct {
```
anorth, line 53: I think either `Evidence` or `Justification` would be better names here. It's always aggregated. Calling it `justification` would align with all the uses of that word throughout the code.

anorth (commented):
I've just skimmed this, and will come back to review properly once all the upstream PRs are landed. I understand the need abstractly to store the signatures of each value that must be subsequently used to create an evidence packet when the quorum is reached, and quorum state is probably a reasonable place.

I don't understand or am not convinced about the others, but they might make more sense as we iterate toward a clear expression here. I do get that if a node receives COMMIT with evidence, it will need to send its own COMMIT and copy the same evidence, but that doesn't create a need to store it. I don't think evidence ever propagates further than the immediate next phase.

f3/granite.go
ranchalp: Used `Justification` in the end

f3/granite.go
ranchalp: used `map[ActorID]struct{}`

ranchalp:
I addressed almost all comments (simultaneously with rebasing). I realize maybe the first (new) commit of this PR might be a bit happy. I can split it into parts again.

I think you are probably right about not having to store every single value, but I will think of a smarter way of doing this probably tomorrow, since quorumState should not necessarily know which of the values will be used if there is more than one perhaps? perhaps not, I will go through the pseudocode tomorrow and update here. Will ping!

ranchalp:
> I do get that if a node receives COMMIT with evidence, it will need to send its own COMMIT and copy the same evidence, but that doesn't create a need to store it. I don't think evidence ever propagates further than the immediate next phase.

Actually, Given the way in which we chose the proposal for the next round, we do not know which value will be chosen as the proposal for the next round. Any of the chose values might only be justified by the explicit justification with which it was sent, and not with locally received PREPAREs. Therefore, at least for the next round, we need to store one justification for each valid COMMIT with a non-zero value received. The map of `justifiedMessages` can be cleaned after sending the CONVERGE of the immediately following round yes (concretely right before returning at `beginConverge`), but until that is executed we need to store the justifiedMessages received (one per non-zero value as mentioned).

Not sure if it is an overkill to cleanup the map after sending the CONVERGE, would you like me to add that line? @anorth

ranchalp:
@Kubuxu Same recommendation to just check files changed (and probably for all PRs of the chain)

f3/granite.go
```diff
@@ -84,12 +103,12 @@ func (m GMessage) String() string {
 }
 
 // Computes the payload for a GMessage signature.
-func SignaturePayload(instance uint32, round uint32, step Phase, value ECChain) []byte {
+func SignaturePayload(instance uint32, round uint32, step string, value ECChain) []byte {
```
Kubuxu, line 106: Keep this as Phase and use `.String()` internally.

f3/granite.go
```diff
@@ -76,6 +77,24 @@ type GMessage struct {
...
+	Justification Justification
+}
+
+// Aggregated list of GossiPBFT messages with the same instance, round and value. Used as evidence for justification of messages
+type Justification struct {
+	Instance uint32
+
+	Round uint32
+
+	Step string
```
Kubuxu, line 90: Keep this as Phase

f3/granite.go
Kubuxu: We will probably switch to encoding the Phase as int in the future (if I were to guess).

f3/granite.go
```diff
@@ -350,8 +369,101 @@ func (i *instance) isValid(msg *GMessage) bool {
 	return true
 }
 
-// TODO Checks whether a message is justified by prior messages.
+// Checks whether a message is justified.
 func (i *instance) isJustified(msg *GMessage) bool {
+
```
Kubuxu, line 374: My initial reaction was that this is hard to read. I will elaborate.

Kubuxu: My suggestion would be to use errors instead of logs (no error = justified).

f3/granite.go
```diff
@@ -350,8 +369,101 @@ func (i *instance) isValid(msg *GMessage) bool {
...
+		if msg.Justification.Round != prevRound {
+			i.log("dropping CONVERGE %v with evidence from wrong round %d", msg.Round, msg.Justification.Round)
+			return false
+		}
+
+		if msg.Justification.Step == PREPARE_PHASE.String() {
+			if msg.Value.HeadCIDOrZero() != msg.Justification.Value.HeadCIDOrZero() {
+				i.log("dropping CONVERGE for value %v with PREPARE evidence for a different value: %v", msg.Value, msg.Justification.Value)
+				return false
+			} else if msg.Value.IsZero() {
```
Kubuxu, line 394: using `if` instead of the `else if` when the previous statement returns would also make it easier to read

f3/granite.go
```diff
@@ -76,6 +77,24 @@ type GMessage struct {
 	Ticket Ticket
 	// Signature by the sender's public key over Instance || Round || Step || Value.
 	Signature []byte
+
+	Justification Justification
+}
+
+// Aggregated list of GossiPBFT messages with the same instance, round and value. Used as evidence for justification of messages
+type Justification struct {
```
Kubuxu, line 85: Justification, in general, has to be exactly the same as the signed message it is based on. This means that the GMessage struct is:
```go
type GMessage struct {
     Sender ActorID
     Current SignedMessage
     CurrentSig Signature
     Ticket Ticket
     Justification SignedMessage
     JustificationSignature Signature
}
```

where SignedMessage contains

f3/granite.go
```diff
@@ -397,7 +509,49 @@ func (i *instance) beginConverge() {
...
+			aggSignature = i.host.Aggregate([][]byte{sig}, aggSignature)
+		}
+		justification = Justification{
+			Instance:  i.instanceID,
+			Round:     i.round - 1,
+			Step:      COMMIT_PHASE.String(),
+			Value:     value,
+			Signers:   signers,
+			Signature: aggSignature,
+		}
```
Kubuxu, line 531: This should disappear if we adopt the GMessage changes I suggested.

f3/granite.go
ranchalp: Done.

f3/granite.go
ranchalp: Done.

f3/granite.go
ranchalp: Done, but I swapped your suggestion: because some conditions of semantic validation do not require (or have) any actual aggregate to verify, I instead have the `isJustified` perform semantic validation and call a new function `verifyJustification` at the end (if it has not returned with nil or error already before).

f3/granite.go
ranchalp: I am not quite sure how this would "disappear"? The justification still needs to be created and the signatures aggregated.

f3/granite.go
ranchalp: Makes sense. I instead made `Justification` have two fields `JustificationSignature JustificationSignature` and `Payload SignedMessage`, so that it is easy to carry both around (many times they go together).

ranchalp:
Thanks @Kubuxu , just addressed all the comments.

f3/granite.go
```diff
@@ -350,9 +372,99 @@ func (i *instance) isValid(msg *GMessage) bool {
...
+
+	if !hasStrongQuorum(power, i.powerTable.Total) {
+		return fmt.Errorf("dropping message as no evidence from a strong quorum: %v", msg.Justification.QuorumSignature.Signers)
+	}
+
+	payload := SignaturePayload(msg.Justification.Payload.Instance, msg.Justification.Payload.Round, msg.Justification.Payload.Step, msg.Justification.Payload.Value)
+	signers := make([]PubKey, 0)
+	_ = msg.Justification.QuorumSignature.Signers.ForEach(func(bit uint64) error {
+		if int(bit) >= len(i.powerTable.Entries) {
+			return nil //TODO handle error
```
Kubuxu, line 391: Please check the error on the outside and error out here. RLE+ has a bit of a sharp edge where it is easy to express "all bits between 0 and 2^64-1" which makes ForEach equivalent to an infinite loop. This is why many methods on it take `max` parameters (if the fact that bit-by-bit iteration happens is unclear). Another way to guard against it is to compare `rle.Count()` against some constant.
We might have to define a constant of `MaxQuorum = 10e6` or something like that (10e6 is arbitrarily large).

f3/granite.go
```diff
@@ -350,9 +372,99 @@ func (i *instance) isValid(msg *GMessage) bool {
...
 }
 
-// TODO Checks whether a message is justified by prior messages.
-func (i *instance) isJustified(msg *GMessage) bool {
-	return true
+func (i *instance) VerifyJustification(msg *GMessage) error {
+
+	power := NewStoragePower(0)
+	_ = msg.Justification.QuorumSignature.Signers.ForEach(func(bit uint64) error {
+		power.Add(power, i.powerTable.Entries[bit].Power)
```
Kubuxu, line 379: Probably combine this into one iteration, as well as range checks.

f3/granite.go
Kubuxu: by that, I meant that the SignedMessage could have been stored in the prevRoundState but it is not that important

f3/granite.go
```diff
@@ -609,68 +799,117 @@ type chainPower struct {
...
+	head := value.HeadCIDOrZero()
+	signatures := make([][]byte, 0)
+	if err := signers.ForEach(func(i uint64) error {
+		if signature, ok := q.received[q.powerTable.Entries[i].ID].heads[head]; ok {
+			if len(signature) == 0 {
+				panic("signature is 0")
+			}
+			signatures = append(signatures, signature)
+		} else {
+			panic("QuorumSignature not found")
```
Kubuxu, line 890: Please don't introduce new panics, let's use error propagation

ranchalp: Separate PR it is then.

f3/granite.go
```diff
@@ -350,9 +372,99 @@ func (i *instance) isValid(msg *GMessage) bool {
 	return true
 }
 
-// TODO Checks whether a message is justified by prior messages.
-func (i *instance) isJustified(msg *GMessage) bool {
-	return true
+func (i *instance) VerifyJustification(msg *GMessage) error {
```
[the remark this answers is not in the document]
ranchalp: Good point, I changed the parameter to accept the justification only.

f3/granite.go
ranchalp: I changed all `ForEach` to check the bit against the powertable entries and return an error if too big (and propagated the error on the outside). Not sure if anything else is required for the `ForEach` given how we use it.

f3/granite.go
ranchalp: See above comment.

ranchalp:
Thanks for the comments! addressed

Kubuxu (approved):
Let's get this merged to unblock (and so it doesn't get out of date).

PR #39, opened 2024-01-16
Replace panics with errors

### LINKED ISSUE #11: Replace panics with errors, catch remaining panics at library boundary

### LINKED ISSUE #38: No-op or fail in unexpected current phase?
The code panics, while the comment says that the situation is treated as a no-op. Treating the situation as an error seems more intuitive, so probably the comment just needs to be updated. Just double-checking.

https://github.com/filecoin-project/go-f3/blob/9f7aed2927690a371e92e3ea9d30295e9ad22bfe/f3/granite.go#L349-L353C3

matejpavlovic:
Resolves #11
Resolves #38

### DIFF
--- a/f3/api.go
+++ b/f3/api.go
@@ -8,7 +8,7 @@ type ChainReceiver interface {
 
 	// Receives a new EC chain, and notifies the current instance if it extends its current acceptable chain.
 	// This modifies the set of valid values for the current instance.
-	ReceiveECChain(chain ECChain)
+	ReceiveECChain(chain ECChain) error
 }
 
 // A consensus message.
--- a/f3/api.go
+++ b/f3/api.go
@@ -4,7 +4,7 @@ package f3
 type ChainReceiver interface {
 	// Receives a chain appropriate for use as initial proposals for a Granite instance.
 	// The chain's base is assumed to be an appropriate base for the instance.
-	ReceiveCanonicalChain(chain ECChain, power PowerTable, beacon []byte)
+	ReceiveCanonicalChain(chain ECChain, power PowerTable, beacon []byte) error
 
 	// Receives a new EC chain, and notifies the current instance if it extends its current acceptable chain.
 	// This modifies the set of valid values for the current instance.
@@ -19,8 +19,8 @@ type Message interface{}
 type MessageReceiver interface {
 	// Receives a message from another participant.
 	// No validation may be assumed to have been performed on the message.
-	ReceiveMessage(msg *GMessage)
-	ReceiveAlarm(payload string)
+	ReceiveMessage(msg *GMessage) error
+	ReceiveAlarm(payload string) error
 }
 
 // Interface which network participants must implement.
--- a/f3/granite.go
+++ b/f3/granite.go
@@ -120,9 +120,9 @@ func newInstance(
 	instanceID uint32,
 	input ECChain,
 	powerTable PowerTable,
-	beacon []byte) *instance {
+	beacon []byte) (*instance, error) {
 	if input.IsZero() {
-		panic("input is empty")
+		return nil, fmt.Errorf("input is empty")
 	}
 	return &instance{
 		config:        config,
@@ -143,7 +143,7 @@ func newInstance(
 		},
 		acceptable: input,
 		decision:   newQuorumState(powerTable),
-	}
+	}, nil
 }
 
 type roundState struct {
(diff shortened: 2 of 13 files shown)


Kubuxu:
Can you add errors to the Signer interface as well?

f3/api.go
```diff
@@ -4,7 +4,7 @@ package f3
 type ChainReceiver interface {
 	// Receives a chain appropriate for use as initial proposals for a Granite instance.
 	// The chain's base is assumed to be an appropriate base for the instance.
-	ReceiveCanonicalChain(chain ECChain, power PowerTable, beacon []byte)
+	ReceiveCanonicalChain(chain ECChain, power PowerTable, beacon []byte) error
 
 	// Receives a new EC chain, and notifies the current instance if it extends its current acceptable chain.
 	// This modifies the set of valid values for the current instance.
```
anorth, line 10: I suggest adding an error return here too, even if currently it always succeeds. For consistency and preparation for future changes that may need an error.

f3/granite.go
```diff
@@ -157,36 +157,39 @@ func newRoundState(powerTable PowerTable) *roundState {
...
 	i.enqueueInbox(msg)
-	i.drainInbox()
+	return i.drainInbox()
 }
 
-func (i *instance) ReceiveAlarm(_ string) {
-	i.tryCompletePhase()
+func (i *instance) ReceiveAlarm(_ string) error {
+	err := i.tryCompletePhase()
+	if err != nil {
```
anorth, line 185: Nit: use the shorthand
`if err := i.tryCompletePhase(); err != nil { ... }`

(elsewhere too)

f3/granite.go
```diff
@@ -266,28 +276,32 @@ func (i *instance) receiveOne(msg *GMessage) {
...
+		return i.tryQuality()
 	case CONVERGE:
-		i.tryConverge()
+		return i.tryConverge()
 	case PREPARE:
-		i.tryPrepare()
+		return i.tryPrepare()
 	case COMMIT:
 		i.tryCommit(i.round)
+		return nil
```
anorth, line 297: Make both tryCommit and tryDecide return `error` too, even if always nil. We'll probably need it later and can avoid this noise when adding some specific error.

f3/granite.go
```diff
@@ -347,9 +361,9 @@ func (i *instance) beginQuality() {
 
 // Attempts to end the QUALITY phase and begin PREPARE based on current state.
 // No-op if the current phase is not QUALITY.
-func (i *instance) tryQuality() {
+func (i *instance) tryQuality() error {
 	if i.phase != QUALITY {
-		panic(fmt.Sprintf("unexpected phase %s", i.phase))
+		return fmt.Errorf("unexpected phase %s", i.phase)
```
anorth, line 366: Note the expected phase, QUALITY, in the message (elsewhere too)

anorth (commented):
This mostly looks great.

For the Simulator, WDYT about just panicking there rather than propagating the error. This is currently only used for tests, and  removing the error-handling noise from the tests will make them easier to read and write. (The error return from `Run` is good though).

f3/api.go
matejpavlovic: Makes sense. Done.

f3/granite.go
matejpavlovic: ✓

f3/granite.go
matejpavlovic: Yes I was actually thinking about exactly that, wondering which option is better. This is exactly the nudge that was missing to actually do it :-)

f3/granite.go
matejpavlovic: ✓

matejpavlovic:
About the simulator, I agree. It makes the tests simpler. The code now panics directly in the simulator.

matejpavlovic:
For the signer interface, @Kubuxu, I agree, but I'd like to discuss the signing/verifying/aggregation API first. I'd find it more intuitive to model it a bit differently, as I propose in [this commit](https://github.com/filecoin-project/go-f3/pull/42/commits/e0eda3a2aa81c1da46d5f0434a126c5d86515f41) (with the corresponding BLS-based implementation [here](https://github.com/filecoin-project/go-f3/pull/42/commits/2006ec500939c8bbe90ae60871512a8696207a7b)).

I'd suggest merging this and dealing with the signing API separately.

anorth (approved):

Issue #40, opened 2024-01-16
Move message validation and signature checking to be synchronous with Receive


anorth:
The network layer will want to avoid propagating invalid messages, so it needs to synchronously know the validation result.

Move the syntactic message validation (i.e. independent of any state) to be synchronous with Receive, and return a boolean that we can propagate to the network layer.

anorth:
@Kubuxu informs me that in libp2p, message validation is separate from receiving. We can't inform libp2p that a message is invalid _after_ receiving it. In libp2p, validation happens, then some other stuff, then the message is received. So in fact we need a larger refactor to split message validation out to an entry point. Note also that this may be expected to be concurrent.

PR #41, opened 2024-01-16
Move message signature validation into instance, and document the message attestation properties.

anorth:
The public key to use to verify signatures will come from the power table being used as the base for the specific Granite instance. So the verification needs to be done there.

See also #40 for propagating validation results back out to the network layer.

### DIFF
--- a/f3/granite.go
+++ b/f3/granite.go
@@ -28,6 +28,12 @@ const TERMINATED = "TERMINATED"
 
 const DOMAIN_SEPARATION_TAG = "GPBFT"
 
+// A message in the Granite protocol.
+// The same message structure is used for all rounds and phases.
+// Note that the message is self-attesting so no separate envelope or signature is needed.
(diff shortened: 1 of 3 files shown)


Kubuxu (approved):

f3/granite.go
```diff
@@ -303,6 +309,12 @@ func (i *instance) isValid(msg *GMessage) bool {
 			return false
 		}
 	}
+
+	sigPayload := SignaturePayload(msg.Instance, msg.Round, msg.Step, msg.Value)
+	if !i.host.Verify(msg.Sender, sigPayload, msg.Signature) {
```
Kubuxu, line 314: This is unrelated, but Verify should probably return an error for debugablity: signature mismatch, sender ID not found, etc.

Kubuxu: Commented on an issue adding errors to other APIs

ranchalp (commented):
I am late to the party but LGTM!

PR #44, opened 2024-01-18
Initial scaffolding of the f3 module

Kubuxu:


Kubuxu:
It is a very rough initial scaffold; please take a look and possibly merge it such that feedback from the lotus side can influence it.

f3/certstore.go
```diff
@@ -0,0 +1,189 @@
...
+	"sync"
+
+	"github.com/ipfs/go-datastore"
+	"github.com/ipfs/go-datastore/namespace"
+
+	"github.com/Kubuxu/go-broadcast"
+	"golang.org/x/xerrors"
+)
+
+// TODO: setup cbor-gen
```
anorth, line 17: I think TODOs in code like this are generally unhelpful, unless they describe clearly why the thing cannot be done yet. They're generally not something that helps the reader who runs into them while spelunking.  For representing a task list, use an issue.

For early prototyping I'd be pretty relaxed, but this is going into a repo that we're working hard to pull _out_ of that prototype phase and into production code. So I think we should just hold the high standards everywhere.

f3/certstore.go
```diff
@@ -0,0 +1,189 @@
...
+	"context"
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"sync"
+
+	"github.com/ipfs/go-datastore"
+	"github.com/ipfs/go-datastore/namespace"
+
+	"github.com/Kubuxu/go-broadcast"
```
anorth, line 13: I haven't looked at this library, but deps on individually-maintained things in high-value projects like this make me very nervous. Is the repo something appropriate to move into filecoin-project? Or if it's just a little code, can we copy-paste it into this one?

f3/certstore.go
```diff
@@ -0,0 +1,189 @@
...
+func NewCertStore(ctx context.Context, ds datastore.Datastore) (*CertStore, error) {
+	cs := &CertStore{
+		ds: namespace.Wrap(ds, datastore.NewKey("/certstore")),
+	}
+	latestCert, err := cs.loadLatest(ctx)
+	if err != nil {
+		return nil, xerrors.Errorf("loading latest cert: %w", err)
+	}
+	if latestCert != nil {
+		cs.busCerts.Publish(*latestCert)
```
anorth, line 53: How could anyone possibly have subscribed to this yet? 

I suggest moving the side-effects out of the constructor into a new method (that can be called after other things have been constructed and hooked up so they might subscribe here).

f3/certstore.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+	var c Cert
+	err = c.UnmarshalBinary(cb)
+	if err != nil {
+		return nil, xerrors.Errorf("unmarshalling latest cert: %w", err)
+	}
+	return &c, err
+}
+
+func (cs *CertStore) Latest() *Cert {
```
anorth, line 77: Please document the public methods, even if they're kinda obvious.

f3/certstore.go
```diff
@@ -0,0 +1,189 @@
...
+	var c Cert
+	if err := c.UnmarshalBinary(b); err != nil {
+		return nil, xerrors.Errorf("unmarshalling cert: %w", err)
+	}
+	return &c, err
+}
+
+// GetRange returns a range of certs from start to end inclusive.
+// If it encouters missing cert, it returns a wrapped ErrCertNotFound and the available certs
+func (cs *CertStore) GetRange(ctx context.Context, start uint64, end uint64) ([]Cert, error) {
```
anorth, line 102: What do start and end represent? Indices? Epoch numbers?
In which order are they traversed? Can they be traversed the other way?

(Please expand docs with all the semantics)

f3/certstore.go
```diff
@@ -0,0 +1,189 @@
...
+	certs := make([]Cert, 0, len(bCerts))
+	for i := 0; i < len(bCerts); i++ {
+		err := certs[i].UnmarshalBinary(bCerts[i])
+		if err != nil {
+			return nil, xerrors.Errorf("unmarshalling a cert at i=%d, instance %d: %w", i, start+uint64(i), err)
+		}
+	}
+
+	if uint64(len(certs)) < start-end {
+		return certs, xerrors.Errorf("cert at %d: %w", i, ErrCertNotFound)
```
anorth, line 133: What enforces that this range is fully populated? Should the store enforce that? If yes, how can the user ensure they use the API correctly? If no, how can the user understand the gaps?

f3/module.go
```diff
@@ -0,0 +1,140 @@
...
+
+	topic *pubsub.Topic
+}
+
+// NewModule creates and setups new libp2p f3 module
+// context is used for initialization not runtime
+// TODO notification about new EC chains
+// TODO FCEX
+// TODO flesh out EC notifications
+// TODO flesh out signing backend notifications
```
anorth, line 45: Ditto about moving TODOs into issues. Document what the current code _does_ do, not what it doesn't.

f3/module.go
```diff
@@ -0,0 +1,140 @@
...
+	return pubsub.ValidationAccept
+}
+
+type ECBackend interface{}
+
+type TODOVerifier struct{}
+
+// Verifies a signature for the given sender ID.
+func (v *TODOVerifier) Verify(pubKey PubKey, msg []byte, sig []byte) bool {
+	panic("not implemented") // TODO: Implement
```
anorth, line 129: "not implemented" says it all, the TODO doesn't add anything

anorth (commented):
Thanks. Yes I'm very keen that we merge incremental progress here so the constraints from both sides can be worked out early. That said, we're trying to turn this into production rather than prototype code so I think we should maintain production standards from the beginning.

This first review is mainly a high level about "where do we put the code", and some API questions. I hope @Stebalien can offer a detailed review of the implementation.

I think these new bits should be in a different module, possibly two different modules.

`f3` is a poor name for the existing module. Let's rename it `gpbft` soon and constrain its scope to to the things approximately already there.

The whole certificate infra and types could probably be its own `certs` module.

Then the thing currently in `module.go` can go in a separate module too, that can depend on the other ones. We could use the name `f3` for that when available, or `module` or `service` or similar.

A subsequent `standalone` module can host the standalone binary.

f3/certstore.go
Kubuxu: They were more of notes/TODOs for myself.

f3/certstore.go
Kubuxu: It is 84 lines of code for something that @Stebalien and I looked for and could not find. It is a valuable library in general, but we can copy-paste it as well.

f3/certstore.go
Kubuxu: the broadcast maintains last published entry, this primes it

f3/certstore.go
```diff
@@ -0,0 +1,189 @@
...
+		return xerrors.Errorf("checking existence of cert: %w", err)
+	}
+	if exists {
+		return ErrCertAlreadyExists
+	}
+
+	if err := cs.ds.Put(ctx, key, certBytes); err != nil {
+		return xerrors.Errorf("putting the cert: %w", err)
+	}
+	if cert.Instance > cs.Latest().Instance {
```
Stebalien, line 173: Don't we need `latest + 1`? Or are we allowed to skip? `GetRange` doesn't allow skipping.

Stebalien (commented):
First pass.

f3/certstore.go
```diff
@@ -0,0 +1,189 @@
...
+func (cs *CertStore) GetRange(ctx context.Context, start uint64, end uint64) ([]Cert, error) {
+	if start > end {
+		return nil, xerrors.Errorf("start is larger then end: %d > %d", start, end)
+	}
+	bCerts := make([][]byte, 0, start-end+1)
+	i := start
+
+	cs.dsLk.Lock()
+	for ; i <= end; i++ {
+		b, err := cs.ds.Get(ctx, cs.keyForInstance(i))
```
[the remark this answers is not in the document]
Kubuxu: We probably can, but I looked at query definitions of datastore and decided that this is way easier, and I don't expect a significant performance bottleneck here. We can go back to this if it turns out to be.

f3/certstore.go
```diff
@@ -0,0 +1,189 @@
...
+	return &c, err
+}
+
+func (cs *CertStore) Latest() *Cert {
+	return cs.busCerts.Last()
+}
+
+func (cs *CertStore) Get(ctx context.Context, instance uint64) (*Cert, error) {
+	cs.dsLk.Lock()
+	b, err := cs.ds.Get(ctx, cs.keyForInstance(instance))
```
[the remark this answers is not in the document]
Kubuxu: It depends on whether the data store is thread-safe or not.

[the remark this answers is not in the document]
Stebalien: I'd just document "must be thread safe" (most of our code assumes thread-safe datastores and, really, we should just make that a requirement at this point.

f3/certstore.go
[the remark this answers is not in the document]
Stebalien: Sg. Also, the query logic in the go-datastore API was _painful_ to implement and I'm still not sure it's correct.

f3/certstore.go
Kubuxu: We should enforce it, I will add it.

f3/certstore.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+var ErrCertAlreadyExists = errors.New("certificate already exists in the CertStore")
+var ErrCertNotFound = errors.New("certificate not found")
+
+// CertStore is responsible for storying and relaying information about new finality certificates
+type CertStore struct {
+	dsLk     sync.Mutex
+	ds       datastore.Datastore
+	busCerts broadcast.Channel[Cert]
```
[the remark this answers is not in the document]
Kubuxu: What do you mean by?
> keep using the "zero" value as the last value

f3/certstore.go
Kubuxu: Added check earlier in the codepath

f3/certstore.go
```diff
@@ -0,0 +1,189 @@
...
+
+	certs := make([]Cert, 0, len(bCerts))
+	for i := 0; i < len(bCerts); i++ {
+		err := certs[i].UnmarshalBinary(bCerts[i])
+		if err != nil {
+			return nil, xerrors.Errorf("unmarshalling a cert at i=%d, instance %d: %w", i, start+uint64(i), err)
+		}
+	}
+
+	if uint64(len(certs)) < start-end {
```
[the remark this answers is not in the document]
Kubuxu: Switched to cap() comparison

f3/certstore.go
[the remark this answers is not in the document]
Stebalien: I mean, in go-broadcast, use `T` to record the last value (where the initial value is the zero value of T), not `*T`. Then expect users to use `*T` where they actually want pointers.

Stebalien:
@Kubuxu I have some PRs in https://github.com/Kubuxu/go-broadcast/pulls that you should merge.

Kubuxu:
Please re-review.
Regarding the package renaming, let's do it separately in one large go whenever we can merge it right away. Otherwise, it will be a merge conflict party.

f3/certstore.go
```diff
@@ -0,0 +1,193 @@
...
+		}
+
+		bCerts = append(bCerts, b)
+	}
+
+	certs := make([]Cert, len(bCerts))
+	for j, bCert := range bCerts {
+		err := certs[j].UnmarshalBinary(bCert)
+		if err != nil {
+			return nil, xerrors.Errorf("unmarshalling a cert at i=%d, instance %d: %w", i, start+uint64(i), err)
```
Stebalien, line 126:
```suggestion
			return nil, xerrors.Errorf("unmarshalling a cert at i=%d, instance %d: %w", j, start+uint64(j), err)
```

f3/certstore.go
```diff
@@ -0,0 +1,193 @@
...
+		}
+	}
+
+	if len(certs) < cap(bCerts) {
+		return certs, xerrors.Errorf("cert at %d: %w", i, ErrCertNotFound)
```
Stebalien, line 131:
```suggestion
		return certs, xerrors.Errorf("cert at %d: %w", len(certs), ErrCertNotFound)
```

f3/certstore.go
```diff
@@ -0,0 +1,193 @@
...
+	if cs.Latest() != nil && cs.Latest().Instance+1 != cert.Instance {
+		return xerrors.Errorf("attempted to add cert at %d but the previous one is %d",
+			cert.Instance, cs.Latest().Instance)
+	}
+
+	certBytes, err := cert.MarshalBinary()
+	if err != nil {
+		return xerrors.Errorf("marshalling cert instance %d: %w", cert.Instance, err)
+	}
+	if err := cs.putInner(ctx, cert, certBytes); err != nil {
```
Stebalien, line 162: I'd inline this function (can remove the duplicate `keyForInstance` and this is the only caller).

f3/module.go
```diff
@@ -0,0 +1,155 @@
...
+		ds:     ds,
+		host:   h,
+		pubsub: ps,
+		verif:  verif,
+		sigs:   sigs,
+		ec:     ec,
+		log:    log,
+	}
+
+	if err := m.setupPubsub(); err != nil {
```
Stebalien, line 64: Why not put this in `Run`?

f3/module.go
```diff
@@ -0,0 +1,155 @@
...
+		return xerrors.Errorf("subscribing to topic: %w", err)
+	}
+
+	for {
+		msg, err := sub.Next(ctx)
+		if err != nil {
+			m.log.Errorf("pubsub subscription.Next() returned an error: %+v", err)
+			break
+		}
+		_ = msg
```
Stebalien, line 106: Add an XXX/TODO here?

Stebalien (commented):
More nits, but honestly, I think we should merge this then start iterating on master.

f3/certstore.go
Kubuxu: this can get even simpler as now we don't hold the lock getting, so I can get and immediately unmarshal

f3/certstore.go
Kubuxu: left over from more complex putInner which I back then extracted to `defer dsLk.Unlock()`, thanks for second eyes

f3/certstore.go
Kubuxu: but I I will keep it the way it is

f3/certstore.go
```diff
@@ -0,0 +1,187 @@
...
+// It errors if the certificate is already in the store
+func (cs *CertStore) Put(ctx context.Context, cert *Cert) error {
+	key := cs.keyForInstance(cert.Instance)
+
+	exists, err := cs.ds.Has(ctx, key)
+	if err != nil {
+		return xerrors.Errorf("checking existence of cert: %w", err)
+	}
+	if exists {
+		return ErrCertAlreadyExists
```
Stebalien, line 154: nit: any reason to return an error rather than making this idempotent?

Stebalien: Also, not thread safe.

f3/certstore.go
```diff
@@ -0,0 +1,187 @@
...
+		return xerrors.Errorf("attempted to add cert at %d but the previous one is %d",
+			cert.Instance, cs.Latest().Instance)
+	}
+
+	certBytes, err := cert.MarshalBinary()
+	if err != nil {
+		return xerrors.Errorf("marshalling cert instance %d: %w", cert.Instance, err)
+	}
+	cs.writeLk.Lock()
+	defer cs.writeLk.Unlock()
```
Stebalien, line 167: To be safe, these checks should likely go under the lock, right?

f3/module.go
```diff
@@ -0,0 +1,148 @@
...
+	sigs   Signer
+	ec     ECBackend
+	log    Logger
+
+	// topic is populated after Run is called
+	topic *pubsub.Topic
+}
+
+// NewModule creates and setups new libp2p f3 module
+// The context is used for initialization not runtime.
```
Stebalien, line 34: This should _probably_ move to lotus eventually, IMO.

Stebalien (commented):
Ok, one concern with respect to concurrency.

f3/certstore.go
Kubuxu: You are right that making it idempotent is a better pattern

PR #54, opened 2024-01-24
Switch to uint64 for Instance and Round

Kubuxu:


Stebalien (approved):

Kubuxu:
This will probably cause merge conflicts for many PRs. IDK what we want to do first.

Kubuxu:
I will merge it as I have another branch depending on it; sorry for the conflict.

anorth:
Why?

Kubuxu:
1. uint32 is not supported by cbor gen
2. Although 2^32 instances probably won't happen, it is only a few orders of magnitude away.

8bytes in signature hashing won't save us, and for propagation, we will most likely use CBOR, which will encode in 4 bytes.

Issue #48, opened 2024-01-22
Optimize explicit message justification


ranchalp:
At the moment (#37 #47), explicit justification takes all messages that justify the decision and aggregates them into the justification. This is an overkill: only a number of messages that form a strong quorum needs to be aggregated. Optimize this.

ranchalp:
Optimizing only creation as optimizing verification without actually enforcing "minimal" justifications (minimal meaning as few bits set as necessary to reach a strong quorum) is an overkill as well: we need to aggregate the public keys associated to each set bit anyways in order to verify the aggregation.

PR #55, opened 2024-01-25
Optimize justification creation

ranchalp:
#48

### DIFF
--- a/f3/granite.go
+++ b/f3/granite.go
@@ -373,6 +373,7 @@ func (i *instance) isValid(msg *GMessage) bool {
 
 func (i *instance) VerifyJustification(justification Justification) error {
 <<<<<<< HEAD
+<<<<<<< HEAD
 
 	power := NewStoragePower(0)
 	signers := make([]PubKey, 0)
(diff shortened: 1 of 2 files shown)


ranchalp:
The multiplication in `hasStrongQuorum` can be modified to only be done once to further optimize, but not sure if that necessary. I think the most important of this PR is that it makes the creator of a justification to only use the minimum amount of signatures required to make a strong quorum, which I believe eventually we should enforce on the receiver side (invalidate an aggregate signature if it contains more signatures than required), to prevent the creator to helplessly keep aggregating public keys of potentially even all nodes and only at then realize that the signature failed.

ranchalp:
Had #11 be fixed here directly in the latest commit.

anorth:
This PR is doing too many things, not all of which I think we are aligned on. Can you please scope it down to _either_ what the tile says, or refactors, or error handling.

When "optimising" here, I strongly suggest against doing anything with any complexity cost. I do think stopping signature aggregation/verification once a strong quorum is reached is reasonable (but not at all necessary), but making the code messy by inflating bitfields into slices before iterating them is not.

I do not think we should enforce it. Confidence could be improved if there is more-than-necessary evidence for a decision. Parties may decide to ignore it, but I don't think we should prevent it. (Please open an issue if you want to further explore this - it would be a spec change).

ranchalp:
Fair enough, I will make a separate PR for #11  (except for the one that I introduced in this new code and that does not reflect an invariant violation, which is in a new commit in this PR).

ranchalp:
On optimising without adding complexity, and not enforcing "minimal" justifications (minimal meaning as few bits set as necessary to reach a strong quorum): 

- I have changed `bitfield.All()` calls to `.ForEach` that return an `EOF`-like error when strong quorum is reached, so that the optimization does not come at the cost of any complexity cost.

- I have removed any modification to the verification of justifications (if we do not want to enforce minimal certificates we might as well not have any optimization, since the optimization is negligible: we need to aggregate the public keys associated to each set bit anyways).

ranchalp:
Thanks for the comments @anorth

f3/granite.go
```diff
@@ -921,6 +926,41 @@ func (q *quorumState) getSigners(value ECChain) bitfield.BitField {
...
+		return strongQuorumSigners, nil
+	}
+
+	// if we didn't find a strong quorum,
+	// return an empty bitfield, and whatever
+	// error that was returned by ForEach (if any)
+	return bitfield.New(), err
+}
+
+type StrongQuorumReachedError struct{}
```
Kubuxu, line 958: You can use errors.New, defined as a var, and then errors.Is to detect it.

anorth (commented):
Let's wait for resolution on #11. I am not happy with the error returns resulting from state invariant violations.

anorth (commented):
As for https://github.com/filecoin-project/go-f3/pull/59#issuecomment-1915630222, can we make this error go away? The code seems to be inspecting the same collection twice, and the error resulting from the probably-impossible removal of some entry between the inspections. Change HasStrongQuorumAgreement to FindStrongQuorumAgreement that returns the signatures at the same time?

ranchalp:
Will address @anorth comments once we merge #59 and after rebasing (following whatever decision we commit to according to #59 and #11 ).

ranchalp:
As promised, #59 is merged so rebased and applied same methodology here. Feel free to review @anorth

anorth:
@ranchalp I have pushed changes to simplify things a bit, and give a smaller diff. As a bonus, the change is now a net reduction in code despite doing more!

anorth (approved):
@ranchalp please review my version yourself, but then good to merge if all ok

PR #58, opened 2024-01-26
Remove unnecessary wrapping of GMessage fields.

anorth:
Removes the `SignedMessage` type which wrapped the non-self-certifying fields of GMessage. I think this was introduced to foster some sort of re-use, but it resulted in a lot of noise of `.Current` and `.Payload` throughout all code that looks at messages. It turns out to be more code to handle the wrapping than just use the four fields directly.

### DIFF
--- a/adversary/equiv.go
+++ b/adversary/equiv.go
@@ -56,44 +56,36 @@ func (w *WitholdCommit) Begin() {
 	// All victims need to see QUALITY and PREPARE in order to send their COMMIT,
 	// but only the one victim will see our COMMIT.
 	w.host.BroadcastSynchronous(w.id, f3.GMessage{
-		Sender: w.id,
-		Current: f3.SignedMessage{
-			Instance: 0,
-			Round:    0,
-			Step:     f3.QUALITY_PHASE,
-			Value:    w.victimValue,
-		},
+		Sender:    w.id,
+		Instance:  0,
+		Round:     0,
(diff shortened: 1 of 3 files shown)


Kubuxu:
I preferred this scheme as this allowed us to put SignaturePayload type on top of the `SignedMessage`, which then is used across justification and the message itself. It essentially delimits what is signed by the signature and what is not.

ranchalp:
The organization of having `SingedMessage` makes sense to me, but at the moment the benefit does not seem to outweight the cost of having to reference to `.Current` and `.Payload`. 

I think the same can be said about `QuorumSignature`, though. 

Otherwise, LGTM!

ranchalp (approved):

anorth:
@Kubuxu I don't quite understand your proposal. If you would like me to hold off merging this, could you perhaps sketch it out in pseudocode here, or make an alternative PR?

Kubuxu:
I can. An alternative would be to utilize the rarely used struct embedding for these fields (but I'm unsure if cbor-gen supports it).

anorth:
Aside: SignedMessage was a weird name for something without a signature, so if re-introduced please pick a different name.

anorth:
Ping @Kubuxu

Kubuxu:
See https://github.com/filecoin-project/go-f3/pull/65

anorth:
Closing in favour of #65 for now. We can revisit later if worthwhile.

PR #59, opened 2024-01-26
Replace panics where relevant

### LINKED ISSUE #11: Replace panics with errors, catch remaining panics at library boundary

ranchalp:
See #11 

Notice that the removed errors/panics are because (by the execution of the code within the same function) it is impossible for the error/panic to ever occur (either because of the size of slices created or because the error has been checked for previously in the function). 

Also, left panics that reflect a situation of the state of f3 that should never happen (and thus a panic is what suits the code IMO).

### DIFF
--- a/f3/granite.go
+++ b/f3/granite.go
@@ -386,7 +386,7 @@ func (i *instance) VerifyJustification(justification Justification) error {
 	}); err != nil {
 		return fmt.Errorf("failed to iterate over signers: %w", err)
 	}
-	
+
 	payload := SignaturePayload(justification.Payload.Instance, justification.Payload.Round, justification.Payload.Step, justification.Payload.Value)
 	_ = justification.QuorumSignature.Signers.ForEach(func(bit uint64) error {
 		return nil
@@ -818,8 +818,8 @@ func (i *instance) log(format string, args ...interface{}) {
 // which values have reached a strong quorum of support.
 // Supports receiving multiple values from each sender, and hence multiple strong quorum values.
 type quorumState struct {
-	// CID of each chain received, by sender. Allows detecting and ignoring duplicates.
-	received map[ActorID]senderSent
+	//// CID of each chain received, by sender. Allows detecting and ignoring duplicates.
+	received map[ActorID]struct{}
 	// The power supporting each chain so far.
 	chainSupport map[TipSetID]chainSupport
 	// Total power of all distinct senders from which some chain has been received so far.
@@ -830,17 +830,11 @@ type quorumState struct {
 	justifiedMessages map[TipSetID]Justification
 }
 
-// The set of chain heads from one sender and associated signature, and that sender's power.
-type senderSent struct {
-	heads map[TipSetID][]byte
-	power *StoragePower
-}
-
 // A chain value and the total power supporting it
 type chainSupport struct {
 	chain           ECChain
 	power           *StoragePower
-	signers         map[ActorID]struct{}
+	signers         map[ActorID][]byte
 	hasStrongQuorum bool
 	hasWeakQuorum   bool
 }
@@ -848,7 +842,7 @@ type chainSupport struct {
 // Creates a new, empty quorum state.
 func newQuorumState(powerTable PowerTable) *quorumState {
 	return &quorumState{
-		received:          map[ActorID]senderSent{},
+		received:          map[ActorID]struct{}{},
 		chainSupport:      map[TipSetID]chainSupport{},
 		sendersTotalPower: NewStoragePower(0),
 		powerTable:        powerTable,
@@ -858,41 +852,38 @@ func newQuorumState(powerTable PowerTable) *quorumState {
 
 // Receives a new chain from a sender.
 func (q *quorumState) Receive(sender ActorID, value ECChain, signature []byte, justification Justification) {
-	head := value.HeadCIDOrZero()
-	fromSender, ok := q.received[sender]
 	senderPower, _ := q.powerTable.Get(sender)
-	sigCopy := make([]byte, len(signature))
-	copy(sigCopy, signature)
+	
+	// Add sender's power to total the first time a value is received from them.
+	if _, ok := q.received[sender]; !ok {
+		q.received[sender] = struct{}{}
+		q.sendersTotalPower.Add(q.sendersTotalPower, senderPower)
+	}
+
+	head := value.HeadCIDOrZero()
+	candidate, ok := q.chainSupport[head]
 	if ok {
 		// Don't double-count the same chain head for a single participant.
-		if _, ok := fromSender.heads[head]; ok {
+		if _, ok := candidate.signers[sender]; ok {
 			return
 		}
-		fromSender.heads[head] = sigCopy
 	} else {
-		// Add sender's power to total the first time a value is received from them.
-		q.sendersTotalPower.Add(q.sendersTotalPower, senderPower)
-		fromSender = senderSent{
-			heads: map[TipSetID][]byte{
-				head: sigCopy,
-			},
-			power: senderPower,
+		candidate = chainSupport{
+			chain:           value,
+			power:           NewStoragePower(0),
+			signers:         map[ActorID][]byte{},
+			hasStrongQuorum: false,
+			hasWeakQuorum:   false,
 		}
+		q.chainSupport[value.HeadCIDOrZero()] = candidate
 	}
-	q.received[sender] = fromSender
 
-	candidate := chainSupport{
-		chain:           value,
-		power:           senderPower,
-		signers:         make(map[ActorID]struct{}),
-		hasStrongQuorum: false,
-		hasWeakQuorum:   false,
-	}
-	if found, ok := q.chainSupport[head]; ok {
-		candidate.power.Add(candidate.power, found.power)
-		candidate.signers = found.signers
-	}
-	candidate.signers[sender] = struct{}{}
+	sigCopy := make([]byte, len(signature))
+	copy(sigCopy, signature)
+	candidate.signers[sender] = sigCopy
+
+	// Add sender's power to total the first time a value is received from them.
+	candidate.power.Add(candidate.power, senderPower)
 
 	candidate.hasStrongQuorum = hasStrongQuorum(candidate.power, q.powerTable.Total)
 	candidate.hasWeakQuorum = hasWeakQuorum(candidate.power, q.powerTable.Total)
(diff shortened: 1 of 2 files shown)


Stebalien:
No, I'm going to have to hard-disagree. Panicing is nice in theory but:

1. You can't crash the entire application, especially if it's something like a blockchain.
2. There needs to be a recovery path.

You _can_ panic up to some failure boundary, but then you need to do _something_ there to recover and/or permanently stop if we really can't recover. But you need to try _very hard_ to recover before giving up, this is an automated system.

To be clear, every single uncaught panic is a security bug waiting to happen. If you're worried about invisible errors, log.

So, this would fine _if_ we have a recover at the top that:

1. First tries really hard to recover. If everyone just gives up on a round if there's a single bug, that's makes that bug a potential attack vector.
2. Next, marks the instance as "unrecoverable" and simply waits for a finality certificate. This is a last resort as it's still an attack vector and you need to hope that the bug can't be reliably exploited.

Stebalien:
Well, ok. Yes, you can panic if you can say "this situation is impossible, I just checked". I'm commenting on the "unrecoverable" cases. Nothing short of memory corruption is truly unrecoverable.

ranchalp:
The 4 panics that we have in `granite.go` at the moment are akin to an index out of bounds error on a slice that we created to have the appropriate size to access within the created bounds. And we do not currently recover from a potential panic from accessing a slice out of bounds in all cases I think. But if we want to make our panics in `granite.go` recoverable let us address them case by case (and make sure we agree on the handling):

- `panic("beginConverge called but no evidence found")` instead of panicking, we could just omit sending the CONVERGE ourselves. This means that eventually we could receive CONVERGE from others and move on without panicking. Notice however that we should *never* be in a situation where this panic occurs (if we are in the next round then we should have a strong quorum of prepares or of commits, so IMO this should be, if not a panic, a very clearly marked error because our code would have a big bug). Also, if one node can find itself in this situation, then all can, if all do then there isn't a single CONVERGE message being sent and behavior of the protocol then is undefined.

- `panic("beginDecide called but no evidence found")` treat analogously to previous panic: do not send DECIDE, clearly flag an error of this situation that should never happen, and await for a lucky strong quorum of DECIDE from others. Same as before: if one can fail then all can fail and then the protocol is waiting forever for messages that will not arrive.

- `panic("signature is 0")` and `panic("QuorumSignature not found")`: These should also never happen except for a big bug. We could just skip this signer and continue with the next. We should also probably remove the signer from the map so that the error does not happen again. 

If this seems reasonable I will code it into this PR.

Stebalien:
Those all seem reasonable. And I agree on panicing on "locally-scoped bugs" (i.e., index out of bounds where we check within the same function). I've just seen lots of "this shouldn't be possible" issues change into "actually, this is now possible" issues over time (code evolves).

And yeah, I agree those would all be bugs and shouldn't be possible. But, even if granite stops working, lotus can still continue validating the main chain (without finality). My goal is to avoid stopping the network unless we're at serious risk of corrupting the state and/or losing funds.

ranchalp:
Actually, for `panic("signature is 0")` and `panic("QuorumSignature not found")`, it is much cleaner for the code to just return the error, not attempt to send the message and, like I suggested in the other ways to handle the existing panics, hope that a strong quorum of other nodes do not fail in the same way. I did this instead. Being this a situation that should never happen, code too much into attempting to work around a stored signature not being valid seems an overkill and noise in the code.

Feel free to argue otherwise though. At the moment the returned errors result in a panic at line 88 of `sim.go`, but it will be the job of the SMR (once we code it) to handle them and not propagate them into the rest of the execution.

With the latest commit, there are now no panics in `granite.go`.

f3/granite.go
```diff
@@ -377,7 +377,7 @@ func (i *instance) VerifyJustification(justification Justification) error {
 	power := NewStoragePower(0)
 	if err := justification.QuorumSignature.Signers.ForEach(func(bit uint64) error {
 		if int(bit) >= len(i.powerTable.Entries) {
-			return fmt.Errorf("invalid signer index: %d", bit)
+			return fmt.Errorf("invalid signer bit: %d", bit)
```
anorth, line 380: Nit: I think "index" was clearer here.

f3/granite.go
```diff
@@ -391,15 +391,10 @@ func (i *instance) VerifyJustification(justification Justification) error {
 
 	payload := SignaturePayload(justification.Payload.Instance, justification.Payload.Round, justification.Payload.Step, justification.Payload.Value)
 	signers := make([]PubKey, 0)
-	if err := justification.QuorumSignature.Signers.ForEach(func(bit uint64) error {
-		if int(bit) >= len(i.powerTable.Entries) {
-			return fmt.Errorf("invalid signer index: %d", bit)
-		}
+	_ = justification.QuorumSignature.Signers.ForEach(func(bit uint64) error {
```
anorth, line 394: Ignoring an error raises a big flag while reading the code, so please add a comment explaining why.

"Note that indexing the power table entries with an invalid index will panic here, but the index was checked earlier. There's no other error returned".

This raises a bigger question though: why are we iterating Signers twice? Can you collapse this to a single iteration?

anorth (commented):
I don't agree with the error returns from the beginPhase methods. I will take discussion to the issue though

f3/granite.go
ranchalp: I can collapse it, but thought of separating conceptually different checks on different iterations (and also exit quickly without any extra work if the message does not contain a strong quorum). But actually you are right, not like spamming with certificates not containing a strong quorum in the list of signers will change much (if the sender wants to spam, that is). Pushed the conflated iteration.

f3/granite.go

anorth:
The commit https://github.com/filecoin-project/go-f3/commit/c62c9d687a3e721598545a0456aba6c30e277630 added error returns to the beginPhase methods which are rooted in internal state invariant violations. In this case in fact I think there is a good chance that we can refactor the state to make the error go away.

Move the signatures from the `map[ActorId][TipsetID]` in `received` to the `map[TipsetID][ActorID]` in `chainSupport`. The error arises from incompatibility between these collections, but we never read signatures from the formed except immediately after looping up the actorID keys in the latter. So this error is pointing at a design flaw.

ranchalp:
> Move the signatures from the map[ActorId][TipsetID] in received to the map[TipsetID][ActorID] in chainSupport. The error arises from incompatibility between these collections, but we never read signatures from the former except immediately after looping up the actorID keys in the latter. So this error is pointing at a design flaw.

We can argue one of the errors in the beginPhase functions arises from that (the one arising from getSignatures returning an error), but the other one is more nuance and ingrained in the protocol: the Converge phase can start either because of a strong quorum of PREPAREs (either a (i) strong quorum explicitly received one by one or (ii) through a justification in a COMMIT), or because of (iii) a strong quorum of COMMITs. 

Our code would be terribly wrong if we manage to start a Converge phase and we don't have one of either (i), (ii), or (iii). But the situation remains with no easy refactor to remove the edge-case: either we just start the phase assuming this is just impossible, or we panic, or we return an error. 

What should we do then?

ranchalp:
Ah, and the same would occur even if finding a workaround to the above, to some extent, even using `FindStrongQuorumAgreement` instead of `HasStrongQuorumAgreement`, because if at some point the function `quorumState.Receive(..)` is modified and bugs introduced, there is no guarantee that `chainSupport.StrongQuorum` correlates with whether there is an actual strong quorum of signers or signatures. 

IMO we just need to make a decision to whether we want to return panics and recover them later at some point of the code (in the SMR that runs repeated consensus for example), or try to work around them.

ranchalp:
I think this is probably ready to be merged.

f3/granite.go
```diff
@@ -853,41 +838,38 @@ func newQuorumState(powerTable PowerTable) *quorumState {
...
-		candidate.power.Add(candidate.power, found.power)
-		candidate.signers = found.signers
-	}
-	candidate.signers[sender] = struct{}{}
+
+	sigCopy := make([]byte, len(signature))
+	copy(sigCopy, signature)
+	candidate.signers[sender] = sigCopy
+
+	// Add sender's power to total the first time a value is received from them.
```
anorth, line 871: Out of place comment

f3/powertable.go
```diff
@@ -36,10 +36,7 @@ func NewPowerTable(entries []PowerEntry) *PowerTable {
 		lookup[entry.ID] = i
 		total.Add(total, entry.Power)
 	}
-
-	if len(entries) != len(lookup) {
-		panic("duplicate power entries")
-	}
```
anorth, line 42: This one should return an error instead. It will be called directly by the client user. We need to protect against duplicates, but they need to be able to catch the error.

anorth (approved):
Only nits remain. Approving on the assumption that you will address them, or if you disagree and want to merge anyway, will create a follow-up conversation.

f3/granite.go
```diff
@@ -709,11 +695,16 @@ func (i *instance) tryCommit(round uint64) error {
 func (i *instance) beginDecide(round uint64) {
 	i.phase = DECIDE_PHASE
 	roundState := i.roundState(round)
-	if !roundState.committed.HasStrongQuorumAgreement(i.value.Head().CID) {
+	var (
+		signers    bitfield.BitField
+		signatures [][]byte
+		ok         bool
+	)
```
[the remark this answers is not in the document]
ranchalp: I prefer not having to write (and delete) 3 times var, but I do not have a big preference. I guess I thought it is the idiomatic way in Go (since `var (...)` can only be used in cases like this one).

f3/granite.go
```diff
@@ -813,8 +804,8 @@ func (i *instance) log(format string, args ...interface{}) {
 // which values have reached a strong quorum of support.
 // Supports receiving multiple values from each sender, and hence multiple strong quorum values.
 type quorumState struct {
-	// CID of each chain received, by sender. Allows detecting and ignoring duplicates.
-	received map[ActorID]senderSent
+	//// CID of each chain received, by sender. Allows detecting and ignoring duplicates.
```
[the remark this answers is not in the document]
ranchalp: Changed.

f3/granite.go
ranchalp: Fixed.

f3/powertable.go
ranchalp: Bc of line 33 of the same function this is impossible though.

ranchalp:
merging.

Issue #27, opened 2023-12-19
Implement multiple instances in higher level F3 protocol object


anorth:
The Granite code executes only a single instance.

- Implement repeated instances in an F3 protocol object (currently the Participant). Maintain state of the last finalised tipset.
- Prepare to handle new chains pushed from outside, without introducing any concurrency issues
- Add testing/simulation infra to be able to drive multi-participant, multi-epoch simulations.

This should be with #13 in mind as a consumer, as well as Lotus integration.

anorth:
Some design considerations:
- Computing a power table representation for an F3 instance will take some computational work. We don't want to do it unless necessary. So Lotus can't push a new power table for every change in EC head that it sees (there could be many per epoch). We only want to do that when F3 is going to use it as a new base.
- Lotus's heaviest chain may diverge during the execution of F3, i.e. F3 may finalise something that's not a prefix of Lotus current heaviest (though they will share a base). F3 won't know what the next candidate head is, it needs to wait for Lotus to rebase on the just-finalised tipset and find its next best head.
  - F3 could pull the next head from Lotus (blocking), or
  - F3 could notify Lotus and then wait for a push when Lotus has prepared the next head, power table etc
- We should probably delay advertising new heads to F3 by an epoch to allow some stabilisation. This is a parameter that could be set higher.

The FIP is unclear about timing assumptions for when the next Granite instance is supposed to start, but at present we're going with "immediately". This might often result in proposing the base chain, and many fast Granite instances that make no progress (but stay in sync). We can probably do something less wasteful, like wait at least 30s (parameter) from the timestamp of the last finalised tipset.

anorth:
@Kubuxu and I discussed and agreed on the following basic flow:
1. F3 notifies the host of a new finality decision (push)
2. F3 waits any delay implied by that decision (e.g. until at least one more EC epoch might have executed)
3. F3 asks the host for the chain to use as input to the next instance (pull)

At least of (1) or (3) must be blocking in the host to ensure that when F3 asks for the next chain, it is assured a base of the previous finality, no race condition.

The immediate effect on the code is to change the current ReceiveCanonicalChain call from a push to F3 into a pull from the Host, after some bootstrapping initialisation.

PR #61, opened 2024-01-31
Multi instance consensus

ranchalp:
Draft PR for multi-instance consensus, see #27 .

### DIFF
--- a/adversary/absent.go
+++ b/adversary/absent.go
@@ -33,7 +33,7 @@ func (a *Absent) ReceiveMessage(_ *f3.GMessage) error {
 	return nil
 }
 
-func (a *Absent) ReceiveAlarm(_ string) error {
+func (a *Absent) ReceiveAlarm(_ *f3.AlarmMsg) error {
 	return nil
 }
 
--- a/adversary/absent.go
+++ b/adversary/absent.go
@@ -21,13 +21,15 @@ func (a *Absent) ID() f3.ActorID {
 	return a.id
 }
 
-func (a *Absent) ReceiveCanonicalChain(_ f3.ECChain, _ f3.PowerTable, _ []byte) error {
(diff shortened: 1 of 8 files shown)


ranchalp:
My opinion as reflected in the latest commit:

- F3 pushes newly finalized chains to Lotus, prompting perhaps a change of chain head on EC (by updated fork choice rule)
- Lotus pushes a call to receiveCanonicalChain every time EC changes its chain head (of the heaviest chain)⁠⁠
   - This means a new push from Lotus every time F3 pushes itself its finalised chain (this is important to trigger a new F3 instance).
- ⁠⁠Lotus also pushes all beacons and their associated epochs from drand to F3. F3 handles storing them and deleting them as needed by F3.

- ReceiveECChain is only called locally by f3 (Lotus only calls ReceiveCanonicalChain everytime its chain head changes)

ranchalp:
The changes from the sync discussion and slack are both implemented here in the latest commit.

ranchalp:
I think it would be easier to have `receiveBeacon` as a separated function (even if it makes sense to have the `beacons` in the `receiveCanonicalChain` method. This is because of #69 .

anorth:
A lot has changed since this draft. I've had a quick look but not sure how much makes sense to attempt to rebase, vs just start over.
- Beacons are no longer relevant for timing
- Alarms don't need to identify instance
- I think some assumptions about chains received aren't quite right

I will try to recover our discussion about design here and jot some notes on the issue.

ranchalp:
>     Beacons are no longer relevant for timing
But we still need beacons for the VRF.
>     Alarms don't need to identify instance
Not sure how the signaling works in the rebased code otherwise but for now I will trust you on that one.
>     I think some assumptions about chains received aren't quite right
Happy to discuss this once more detailed.

PR #65, opened 2024-02-03
Cleanup signature payload, popagate more errors, add cbor-gen

Kubuxu:


ranchalp (approved):
LGTM, but see the comments below.

f3/granite.go
```diff
@@ -106,19 +120,9 @@ func (m GMessage) String() string {
...
-	_ = binary.Write(&buf, binary.BigEndian, round)
-	_ = binary.Write(&buf, binary.BigEndian, []byte(step.String()))
-	for _, t := range value {
-		_ = binary.Write(&buf, binary.BigEndian, t.Epoch)
-		buf.Write(t.CID.Bytes())
-		_ = binary.Write(&buf, binary.BigEndian, t.Weight)
-	}
-	return buf.Bytes()
-}
+//func SignaturePayload(instance uint64, round uint64, step Phase, value ECChain) []byte {
```
ranchalp, line 123: Remove

f3/granite.go
```diff
@@ -92,12 +99,19 @@ type SignedMessage struct {
 	Value ECChain
 }
 
-// Aggregated list of GossiPBFT messages with the same instance, round and value. Used as evidence for justification of messages
-type QuorumSignature struct {
-	// Indexes in the base power table of the signers (bitset)
-	Signers bitfield.BitField
-	// BLS aggregate signature of signers
-	Signature []byte
+func (p Payload) MarshalForSigning(nn NetworkName) ([]byte, error) {
```
ranchalp, line 102: Could we not get rid of using `.Current` everywhere if we just make this a method of `GMessage` and create a new method of `GMessage` `createJustification` that takes the signers and signatures as parameters returns the Justification? I think this would be my preferred approach.

adversary/withhold.go
```diff
@@ -52,66 +52,71 @@ func (w *WithholdCommit) ReceiveAlarm(_ string) error {
 	return nil
 }
 
+func broadcastHelper(host sim.AdversaryHost, sender f3.ActorID) func(f3.Payload, *f3.Justification) {
```
anorth, line 55: Nit: please move helpers to bottom of the file so more semantically interesting code comes first.

Also the tiny variable names do my head in.

f3/granite.go
```diff
@@ -65,22 +67,27 @@ type GMessage struct {
...
+	Ticket Ticket
 	// Justification for this message (some messages must be justified by a strong quorum of messages from some previous step).
 	Justification Justification
 }
 
 type Justification struct {
-	Payload         SignedMessage
-	QuorumSignature QuorumSignature
+	// Payload is the payload that is signed by the signature
+	Payload Payload
```
anorth, line 82: This is the key line that I think is the justification for a separate SignedMessage/Payload type. It's sharing the definition of the four fields that must be signed in a GMessage, and thus are the payload of an aggregation of signatures.

Is it worth it? It's pretty borderline. This change overall is an improvement, but just inlining all the fields like #58 gives up very little reuse while removing significant noise.

Let's go forward with this (we can still inline later), since serialisation stuff adds more information about how things need to work here.

f3/granite.go
```diff
@@ -92,32 +99,24 @@ type SignedMessage struct {
...
-}
-
-func (m GMessage) String() string {
-	// FIXME This needs value receiver to work, for reasons I cannot figure out.
-	return fmt.Sprintf("%s{%d}(%d %s)", m.Current.Step, m.Current.Instance, m.Current.Round, &m.Current.Value)
-}
-
-// Computes the payload for a GMessage signature.
-func SignaturePayload(instance uint64, round uint64, step Phase, value ECChain) []byte {
+func (p Payload) MarshalForSigning(nn NetworkName) ([]byte, error) {
```
anorth, line 102: This never returns an error, so remove the error return.

f3/api.go
```diff
@@ -34,7 +34,7 @@ type Receiver interface {
 type Network interface {
 	// Sends a message to all other participants.
 	// The message's sender must be one that the network interface can sign on behalf of.
-	Broadcast(msg *GMessage)
+	Broadcast(msg *GMessage) error
```
anorth, line 37: This is a client failures ,and F3 can't do anything with it. It will just end up back at the client as a result from one of the few entry points where the client calls F3. Yes the F3 instance internal state is still consistent, but the protocol execution may have a problem.

Whatever the client would do in response to this error from F3, it should have done it before sending the error to F3 in the first place. E.g. retrying any transient network failure, etc. If it can't do anything, then it still can't do anything when this is thrown back at it. If it would, say, tear down the instance and wait for a finality certificate, just do that right away. If it would shrug and say "well lets just continue and hope the protocol is fine", then just continue and don't return an error in the first place.

I could well be missing something here, since nothing returns an error into this interface yet, but where do you have in mind that this error can be useful?

(Can you separate this change from the rest of the PR, which is otherwise an improvement)?

f3/granite.go
```diff
@@ -65,22 +67,27 @@ type GMessage struct {
 	// ID of the sender/signer of this message (a miner actor ID).
 	Sender ActorID
 
-	Current SignedMessage
-	// VRF ticket for CONVERGE messages (otherwise empty byte array).
-	Ticket Ticket
+	// Current is the payload that is signed by the signature
+	Current Payload
```
anorth, line 71: We gotta find a better name for this. Payload would be better. Value would be ok, though overloaded (since it also contains a value). Body? Proposal? Vote?

f3/api.go
Kubuxu: I can for sure separate.
The only way to tear down the instance would be to panic within Broadcast.
One of the reasons I've added it was the possibility of network failures and letting the instance know about them, but I think you are right; it doesn't need to know.

adversary/withhold.go
Kubuxu: Yeah, the variable names should be a bit longer, the function grew up larger than I was planning.

adversary/withhold.go
```diff
@@ -53,65 +53,48 @@ func (w *WithholdCommit) ReceiveAlarm(_ string) error {
...
+	preparePayload := f3.Payload{
+		Instance: 0,
+		Round:    0,
+		Step:     f3.PREPARE_PHASE,
+		Value:    w.victimValue,
 	}
-	payload := f3.SignaturePayload(0, 0, f3.PREPARE_PHASE, w.victimValue)
+	broadcast(preparePayload, nil)
+
+	payload := f3.Payload{
```
anorth, line 73:
```suggestion
	commitPayload := f3.Payload{
```

f3/granite.go
```diff
@@ -517,37 +515,31 @@ func (i *instance) beginConverge() {
 	if signers, signatures, ok := prevRoundState.committed.FindStrongQuorumAgreement(ZeroTipSetID()); ok {
 		value := ECChain{}
 		aggSignature := i.host.Aggregate(signatures, nil)
-		justificationPayload := SignedMessage{
+		justificationPayload := Payload{
```
anorth, line 518: Nit: I suggest inlining this into the `Justification{ ... }` constructor here and everywhere else that it isn't yet too.

go.mod
```diff
@@ -4,24 +4,25 @@ go 1.20
 
 require (
 	github.com/filecoin-project/go-bitfield v0.2.4
+	github.com/ipfs/go-cid v0.0.6
```
anorth, line 7: Rant: It's annoying that cbor-gen forces this dependency even though we have no CIDs

anorth (approved):
Thanks for adding cbor-gen. Could you please follow-up with a Makefile rule to regenerate this file?

PR #66, opened 2024-02-05
15% improvement in tests speed

Kubuxu:


anorth:
FYI this shouldn't have been a bitfield in the first place, just a sorted list of ints. I'll fix it.

Kubuxu:
@anorth I'm digging in this area as part of the BLS signatures (I need to get pubkes from there as well); I can change it to a list of ints. It will need to end up as list as bitfield at some point either way but I see what you are saying.

anorth:
I already changed it

PR #67, opened 2024-02-05
Add test demonstrating agreement on longest common prefix

### LINKED ISSUE #64: QUALITY does not select the heaviest prefix of the inputs
At the moment, the quality step performs the [following](https://github.com/filecoin-project/go-f3/blob/4c3715ca65c7eac1d1c7f48ee5ddae547a9e1729/f3/granite.go#L1036C1-L1046C2):

```
foundQuorum := i.quality.HasStrongQuorumAgreement(i.proposal.Head().CID)
	timeoutExpired := i.host.Time() >= i.phaseTimeout

	if foundQuorum {
		// Keep current proposal.
	} else if timeoutExpired {
		strongQuora := i.quality.ListStrongQuorumAgreedValues()
		i.proposal = findFirstPrefixOf(strongQuora, i.proposal)
	}
[...]
// Returns the first candidate value that is a prefix of the preferred value, or the base of preferred.
func findFirstPrefixOf(candidates []ECChain, preferred ECChain) ECChain {
	for _, v := range candidates {
		if preferred.HasPrefix(v) {
			return v
		}
	}

	// No candidates are a prefix of preferred.
	return preferred.BaseChain()
}
```

This is not what the pseudocode at [FIP-XXX](https://github.com/filecoin-project/FIPs/blob/8b90d016fbf6ce867ca18fc45e9acef76988452a/FIPS/fip-xxxx.md?plain=1#L357-L365) describes:

_"Let c be the heaviest common prefix of the inputs of all honest participants in instance . Then, if an honest participant finalizes , is a prefix of with probability > 0.5."_

```
10:       BEBroadcast <QUALITY, value>; trigger (timeout)
11:       collect a clean set M of valid QUALITY messages
          until StrongQuorum(proposal, M) OR timeout expires
12:       let C={prefix : IsPrefix(prefix,proposal) and StrongQuorum(prefix,M)}
13:       if (C = ∅)
14:         proposal ← baseChain \* no proposals of high-enough quality
15:       else
16:         proposal ← heaviest prefix ∈ C \* this becomes baseChain or sth heavier
17:       value ← proposal
```

That is, in the current code, if the power table contains 4 nodes all with the same power and the inputs are [a b c], [a b d], [a b e], [a b f], then findFirstPrefixOf will finalize baseChain (i.e. "[a]"), whereas we want it to finalize [a b]. Fix this.

anorth:
Adds a test demonstrating that #64's claim about incorrect behaviour isn't right.

Closes #64.

### DIFF
--- a/test/honest_test.go
+++ b/test/honest_test.go
@@ -140,6 +140,27 @@ func TestRequireStrongQuorumToProgress(t *testing.T) {
 	}
 }
 
+func TestLongestCommonPrefix(t *testing.T) {
+	// This test uses a synchronous configuration to ensure timely message delivery.
+	// If async, it is possible to decide the base chain if QUALITY messages are delayed.
+	sm := sim.NewSimulation(newSyncConfig(4), GraniteConfig(), sim.TraceAll)
+	ab := sm.Base.Extend(sm.CIDGen.Sample())
+	abc := ab.Extend(sm.CIDGen.Sample())
+	abd := ab.Extend(sm.CIDGen.Sample())
+	abe := ab.Extend(sm.CIDGen.Sample())
+	abf := ab.Extend(sm.CIDGen.Sample())
+	sm.ReceiveChains(
+		sim.ChainCount{Count: 1, Chain: abc},
+		sim.ChainCount{Count: 1, Chain: abd},
+		sim.ChainCount{Count: 1, Chain: abe},
+		sim.ChainCount{Count: 1, Chain: abf},
+	)
+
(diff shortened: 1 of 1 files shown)


ranchalp:
You are right, my bad, I had not realised that it is at `receiveOne` that all suffices are considered. Seems weird to me this piece of core protocol logic is replaced as a simulation of receiving multiple QUALITY messages, so I still propose a different approach (which can also prevent spamming with ridiculously long suffices that will never reach a strong quorum) in #60 . But will close the issue I created. Thanks for the heads up!

anorth:
Please also review/approve this so we can merge it!

ranchalp (approved):

PR #70, opened 2024-02-09
Introduce BLS signing, adapt APIs for it to work

Kubuxu:
Only implements BLS with keys on G2, we will need to get keys on G1 into kyber.


Kubuxu:
Tests look good using BLS by using the `F3_TEST_USE_BLS=1` but take forever.
Still running:
```
=== RUN   TestAbsent
--- PASS: TestAbsent (192.31s)
=== RUN   TestSingleton
--- PASS: TestSingleton (0.01s)
=== RUN   TestSyncPair
--- PASS: TestSyncPair (0.03s)
=== RUN   TestASyncPair
--- PASS: TestASyncPair (238.78s)
=== RUN   TestSyncPairDisagree
--- PASS: TestSyncPairDisagree (0.03s)
=== RUN   TestAsyncPairDisagree
--- PASS: TestAsyncPairDisagree (180.86s)
=== RUN   TestSyncAgreement
--- PASS: TestSyncAgreement (465.37s)
```

sim/fakesig.go
```diff
@@ -0,0 +1,75 @@
...
+
+		aggHash.Write(signer)
+		aggHash.Write(sig)
+	}
+	if !bytes.Equal(aggSig, aggHash.Sum(nil)) {
+		return xerrors.Errorf("aggregate signature miss-match")
+	}
+
+	return nil
+}
```
anorth, line 75: Nice!

sim/sim.go
```diff
@@ -33,14 +33,18 @@ func NewSimulation(simConfig Config, graniteConfig f3.GraniteConfig, traceLevel
 	// Create a network to deliver messages.
 	lat := NewLogNormal(simConfig.LatencySeed, simConfig.LatencyMean)
 	ntwk := NewNetwork(lat, traceLevel)
-	vrf := f3.NewVRF(ntwk, ntwk)
+	if os.Getenv("F3_TEST_USE_BLS") == "1" {
+		ntwk.Signer = blssig.SignerWithKeyOnG2()
+		ntwk.Verifier = blssig.VerifierWithKeyOnG2()
+	}
```
anorth, line 39: Since there's already a Config, I would prefer to use that than reach out to the OS from deeper within the code. The env var is a good idea, but perhaps fetch it from the test code rather than here (so that some tests can configure explicitly regardless of the environment).

test/signing_suite.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+}
+
+func (s *SigningSuite) TestSignAndVerify(t *testing.T) {
+	pubKey := s.signer.GenerateKey()
+	msg := []byte("test message")
+	sig, err := s.signer.Sign(pubKey, msg)
+	require.NoError(t, err)
+
+	err = s.verifier.Verify(pubKey, msg, sig)
```
anorth, line 28: Test also that Verify isn't just `return nil` by checking an invalid signature. Ditto for aggregation

blssig/verifier.go
```diff
@@ -0,0 +1,34 @@
...
+func VerifierWithKeyOnG2() Verifier {
+	suite := bls12381.NewBLS12381Suite()
+	return Verifier{
+		suite:    suite,
+		keyGroup: suite.G2(),
+	}
+}
+
+func (v Verifier) Verify(pubKey f3.PubKey, msg, sig []byte) error {
+	// rejecting zero point pubkeys?
```
anorth, line 26: File an issue for this

anorth (approved):
The API and associated refactors look great!

I'm unqualified to review the use of BLS libraries, but it looks straightforward.

f3/granite.go
```diff
@@ -507,14 +511,23 @@ func (i *instance) tryQuality() error {
...
 	prevRoundState := i.roundState(i.round - 1)
 
 	var justification Justification
-	if signers, signatures, ok := prevRoundState.committed.FindStrongQuorumAgreement(ZeroTipSetID()); ok {
+	if quorum, ok := prevRoundState.committed.FindStrongQuorumAgreement(ZeroTipSetID()); ok {
 		value := ECChain{}
-		aggSignature := i.host.Aggregate(signatures, nil)
+		aggSignature, err := quorum.Aggregate(i.host)
+		if err != nil {
+			panic(fmt.Sprintf("aggregating for convergage: %v", err))
```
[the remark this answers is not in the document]
Kubuxu: not really, line information created in this line by xerrors would be redundant, but we can use it as there would be no drawback either

go.mod
```diff
@@ -1,9 +1,11 @@
 module github.com/filecoin-project/go-f3
 
-go 1.21
+go 1.20
```
Stebalien, line 3: why?

sim/network.go
```diff
@@ -34,7 +32,13 @@ const (
 
 const _ = TraceAll // Suppress unused constant warning.
 
+type signingBackend interface {
+}
```
Stebalien, line 36:
```suggestion
type signingBackend any
```

blssig/signer.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+}
+
+func (s *Signer) GenerateKey() f3.PubKey {
+	priv, pub := s.scheme.NewKeyPair(s.suite.RandomStream())
+	pubKeyB, err := pub.MarshalBinary()
+	if err != nil {
+		panic(err)
+	}
+	s.keys[string(pubKeyB)] = priv
```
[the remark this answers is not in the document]
Kubuxu: Yes, the signer will be outsourced to lotus.

go.mod
Kubuxu: It got bumped by go-broadcast. The CI runs 1.20, lotus runs 1.20.

f3/granite.go
```diff
@@ -894,40 +927,72 @@ func (q *quorumState) HasStrongQuorumAgreement(cid TipSetID) bool {
...
-		justificationPower.Add(justificationPower, q.powerTable.Entries[idx].Power)
-		strongQuorumSigners.Set(uint64(idx))
-		signatures = append(signatures, chainSupport.signers[q.powerTable.Entries[idx].ID])
+		entry := q.powerTable.Entries[idx]
+		justificationPower.Add(justificationPower, entry.Power)
+		signatures = append(signatures, chainSupport.signatures[entry.ID])
+		pubkeys = append(pubkeys, entry.PubKey)
 		if hasStrongQuorum(justificationPower, q.powerTable.Total) {
-			return strongQuorumSigners, signatures, true
+			found = true
```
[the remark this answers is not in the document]
Kubuxu: you are right, I can't recall why I went down this path, I think I had extra code before return

blssig/verifier.go
Kubuxu: IIRC zero points were safe, but excluded by some implementations due to an abundance of caution (but don't quote me on that) due to similar issues in EdDSA.
Yeah, I will exclude them.

sim/network.go
Kubuxu: This needs to get deleted as it is a dev remnant.

go.mod
Stebalien: Ah, I see.

f3/granite.go
```diff
@@ -894,40 +927,72 @@ func (q *quorumState) HasStrongQuorumAgreement(cid TipSetID) bool {
...
+	Signatures [][]byte
+}
+
+func (q QuorumResult) Aggregate(v Verifier) ([]byte, error) {
+	return v.Aggregate(q.PubKeys, q.Signatures)
+}
+func (q QuorumResult) SignersBitfield() bitfield.BitField {
+	signers := make([]uint64, 0, len(q.Signers))
+	for _, s := range q.Signers {
+		signers = append(signers, uint64(s))
```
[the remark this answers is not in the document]
Kubuxu: it gets ugly in FindStrongQuorumAgreement (between indexing and sorting using sort.Slice)

blssig/aggregation.go
```diff
@@ -0,0 +1,77 @@
...
+	kPubkeys := make([]kyber.Point, 0, len(pubkeys))
+	for i, p := range pubkeys {
+		point := v.keyGroup.Point()
+		err := point.UnmarshalBinary(p)
+		if err != nil {
+			return nil, xerrors.Errorf("unarshalling pubkey at index %d: %w", i, err)
+		}
+		if point.Equal(v.keyGroup.Point().Null()) {
+			return nil, xerrors.Errorf("the public key at %d is a null point", i)
+		}
```
Stebalien, line 24: I'd consider making a helper function for this.

Stebalien (approved):

blssig/aggregation.go
Kubuxu: I considered it, decided that 2x reuse does not yet warrant it.

Kubuxu:
Still have to resolve the test config and more tests for signing.

sim/sim.go
Kubuxu: I left it still deep but allowed tests to override.

f3/granite.go
```diff
@@ -536,14 +547,19 @@ func (i *instance) beginConverge() {
...
 			Signature: aggSignature,
 		}
 	} else if justification, ok = prevRoundState.committed.justifiedMessages[i.proposal.Head().CID]; ok {
 		//justificationPayload already assigned in the if statement
 	} else {
 		panic("beginConverge called but no evidence found")
 	}
+	ticket, err := i.vrf.MakeTicket(i.beacon, i.instanceID, i.round)
+	if err != nil {
+		i.log("creating VRF ticket: %v", err)
```
anorth, line 560: Include the word "error"? This will be clearer once we introduce a proper logger.

f3/granite.go
```diff
@@ -751,7 +774,12 @@ func (i *instance) broadcast(round uint64, step Phase, value ECChain, ticket Tic
 	}
 	sp := p.MarshalForSigning(TODONetworkName)
 
-	sig := i.host.Sign(i.participantID, sp)
+	sig, err := i.sign(sp)
+	if err != nil {
+		i.log("failed to sign message: %v", err)
```
anorth, line 779: Ditto say "error"?

f3/granite.go
```diff
@@ -848,7 +881,7 @@ func (q *quorumState) Receive(sender ActorID, value ECChain, signature []byte, j
 
 	sigCopy := make([]byte, len(signature))
 	copy(sigCopy, signature)
```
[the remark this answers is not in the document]
anorth: This may have been a bug in the previous fake sig implementation. yes I think we should remove the copy

anorth (approved):

blssig/aggregation.go
Stebalien: I'm just concerned that someone will do this then _not_ check if it's null. But we're probably fine.

PR #79, opened 2024-02-13
Use full TipSet object as map key, avoid using CIDs

### LINKED ISSUE #10: Validate weight etc of received tipsets
A tipset referenced in a protocol message includes its weight, but the receiver doesn't verify it. It probably makes sense to verify these values against locally-known chains, or against messages from other participants that have referenced the same tipset. These value are all logically constant for any Tipset CID.

anorth:
The existing code has a subtle bug where some participants could disagree about which epoch a tipset is for, and the correspondence between CID and epoch never checked. We now use the full TipSet object as a map key and for all other uses, and never look at either the CID or Epoch individually. Two nodes are only proposing the same thing if the TipSet fields fully match.

Closes #10

### DIFF
--- a/f3/chain.go
+++ b/f3/chain.go
@@ -73,7 +73,7 @@ func (c ECChain) IsZero() bool {
 	return len(c) == 0
 }
 
-// Returns a pointer to the base tipset.
+// Returns the base tipset.
 func (c ECChain) Base() TipSet {
 	return c[0]
 }
--- a/f3/chain.go
+++ b/f3/chain.go
@@ -1,13 +1,18 @@
 package f3
 
 import (
+	"encoding/binary"
 	"fmt"
+	"io"
 	"reflect"
 	"strconv"
 	"strings"
 )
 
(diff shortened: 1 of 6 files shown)


f3/chain.go
```diff
@@ -59,8 +74,8 @@ func (c ECChain) IsZero() bool {
 }
 
 // Returns a pointer to the base tipset.
```
ranchalp, line 76:
```suggestion
// Returns the base tipset.
```

f3/chain.go
```diff
@@ -75,16 +90,16 @@ func (c ECChain) Suffix() []TipSet {
 // Returns a pointer to the last tipset in the chain.
```
ranchalp, line 90:
```suggestion
// Returns the last tipset in the chain.
```

ranchalp (approved):
LGTM

Kubuxu (approved):

Kubuxu:
Rebased, confirmed diffs side by side.

PR #82, opened 2024-02-14
Implement Synchronization of participants in the current instance #69

### LINKED ISSUE #69: Implement Synchronization of participants in the current instance
From [FIP](https://github.com/filecoin-project/go-f3/issues/68): 

"GossiPBFT ensures termination provided that (i) all participants start the instance at most $\Delta$ apart, and (ii) the estimate on Δ is large enough for $\Delta$-synchrony in some round (either because of the real network delay recovering from asynchrony or because of the estimate increasing).

[Given prior tests performed on GossipSub](https://research.protocol.ai/publications/gossipsub-v1.1-evaluation-report/vyzovitis2020.pdf) (see also [here](https://gist.github.com/jsoares/9ce4c0ba6ebcfd2afa8f8993890e2d98)), we expect that almost all participants will reach sent messages within $Δ=3s$, with a huge majority receiving them even after $Δ=2s$. However, if several participants start the instance $Δ + ε$ after some other participants, termination is not guaranteed for the selected timeouts of $2*Δ$. Thus, we do not rely on an explicit synchrony bound for correctness. Instead, we (i) use drand as a beacon to synchronize participants within an instance and (ii) increase the estimate of Δ locally within an instance as rounds progress without decision.

The synchronization of participants is performed in the call to updateTimeout(timeout, round) (line 47), and works as follows:

* Participants start an instance with $Δ=2s$.
* If the first and second rounds fail to reach termination, participants start the 3rd round with $Δ=3s$.
* If 5 rounds fail to reach termination (rounds from 0 to 4), participants wait to receive the next drand epoch value to start round 5. This ensures that a potential ε delay between when participants started the instance is not carried over beyond round 4.
* If round 5 fails to reach a decision, participants set exponential timeout increase $Δ= 1.3*Δ$ for every failed round. This ensures termination even if the real network delay increases arbitrarily."

ranchalp:
Closes #69


f3/granite.go
```diff
@@ -4,16 +4,22 @@ import (
...
+	"math"
+	"sort"
 )
 
 type GraniteConfig struct {
 	// Initial delay for partial synchrony.
 	Delta float64
 	// Change to delta in each round after the first.
 	DeltaRate float64
+	// Absolut extra value to add to delta after the first few unsuccessful rounds.
```
Kubuxu, line 17:
```suggestion
	// Absolute extra value to add to delta after the first few unsuccessful rounds.
```

f3/granite.go
```diff
@@ -769,7 +778,23 @@ func (i *instance) broadcast(round uint64, step Phase, value ECChain, ticket Tic
...
 func (i *instance) alarmAfterSynchrony(payload string) float64 {
-	timeout := i.host.Time() + i.config.Delta + (float64(i.round) * i.config.DeltaRate)
+	timeout := i.host.Time() + i.config.Delta
+
+	if i.round >= 2 {
+		timeout += i.config.DeltaExtra
+	}
+
+	if i.round >= 5 {
+		timeout *= i.deltaRate
```
Kubuxu, line 788: rate should be applied on top of the increment, not absolute time

f3/granite.go
```diff
@@ -4,18 +4,24 @@ import (
...
 	"golang.org/x/xerrors"
+	"math"
+	"sort"
 )
 
 type GraniteConfig struct {
 	// Initial delay for partial synchrony.
 	Delta float64
 	// Change to delta in each round after the first.
 	DeltaRate float64
```
anorth, line 18: As far as I can tell, this is no longer used in the same way in the new spec. I looks like it only applies after the 5th round.
- Move this down below the other fields, since it takes effect last
- Fix the comment
- Perhaps rename to something more serious, like DeltaBackoffExponent?

f3/granite.go
```diff
@@ -4,18 +4,24 @@ import (
...
+	// Absolute extra value to add to delta after the first few unsuccessful rounds.
+	DeltaExtra float64
+	// Time to next clock tick (according to local clock)
+	// Used in timeouts in order to synchronize participants
+	// (by not carrying significant delays from previous/starting rounds)
+	ClockTickDelta float64
```
anorth, line 24: "Clock tick" is a very confusing name, it took me forever to work out. I think you mean something more like epoch duration (when it referred to drand). We're not using drand though, so it's more just an external clock re-sync point. I suggest a name like `ExternalClockResyncPeriod`. Update comments accordingly.

f3/granite.go
```diff
@@ -782,7 +791,24 @@ func (i *instance) broadcast(round uint64, step Phase, value ECChain, ticket Tic
...
-	timeout := i.host.Time() + i.config.Delta + (float64(i.round) * i.config.DeltaRate)
+	delta := i.config.Delta
+
+	if i.round >= 2 {
+		delta += i.config.DeltaExtra
+	}
+
+	if i.round >= 5 {
+		delta *= i.deltaRate
+		i.deltaRate *= i.config.DeltaRate
```
anorth, line 802: Please remove the mutable state variable, you can just count the rounds. 

`delta = delta * deltaRate ^ (i.round - 4)` or similar.

f3/granite.go
```diff
@@ -782,7 +791,24 @@ func (i *instance) broadcast(round uint64, step Phase, value ECChain, ticket Tic
...
+	}
+
+	timeout := i.host.Time() + delta
+	if i.round > 0 && i.round%5 == 0 && i.phase == CONVERGE_PHASE {
```
anorth, line 806: This reference to CONVERGE_PHASE etc is a complete surprise here. It's not mentioned in the spec at all so don't know what the intent is. I think probably there should be a parameter to this method that is passed in from the appropriate call site, and the name and documentatio of that parameter might help describe it.

The spec doesn't say anything about "mod 5". Instead it has an exponential backoff. Which you've implemented above, and seems incompatible with this fixed period.

f3sim.go
```diff
@@ -17,9 +17,10 @@ func main() {
 	maxRounds := flag.Uint64("max-rounds", 10, "max rounds to allow before failing")
 	traceLevel := flag.Int("trace", sim.TraceNone, "trace verbosity level")
 
-	graniteDelta := flag.Float64("granite-delta", 6.000, "granite delta parameter")
-	graniteDeltaRate := flag.Float64("granite-delta-rate", 2.000, "change in delta for each round")
-
+	graniteDelta := flag.Float64("granite-delta", 2.000, "granite delta parameter")
+	graniteDeltaRate := flag.Float64("granite-delta-rate", 1.300, "change in delta for each round")
```
anorth, line 21: Please update the descriptoin

f3sim.go
```diff
@@ -17,9 +17,10 @@ func main() {
...
+	graniteDeltaExtra := flag.Float64("granite-delta-extra", 1.000, "extra delta for each round")
+	clockTickDelta := flag.Float64("clock-tick-delta", 30.000, "clock tick delta in seconds")
```
anorth, line 23: Please rename here other than "clock tick" too

anorth (changes requested):

f3/granite.go
ranchalp: In #69 the pseudocode describes waiting for the next clock tick after the first round. This only occurs once at the beginning of the round. That is why here I control that by checking the the phase is CONVERGE.

What I am implementing here is the drand clock tick every 5 rounds, as the spec says. Would it make more sense in your opinion that we signal with a parameter in the call from CONVERGE that the round is starting?

ranchalp: Something that is not explicitly mentioned in the spec is that it makes more sense to first send the converge message and then wait for the clock tick than otherwise, which is another reason why the above code might seem confusing.

f3/granite.go
ranchalp: Done.

f3/granite.go
ranchalp: Done.

f3/granite.go
ranchalp: Done.

f3sim.go
ranchalp: Done.

f3sim.go
ranchalp: Done.

ranchalp:
See comments being addressed in #87 , and also comment [above](https://github.com/filecoin-project/go-f3/pull/82#discussion_r1491650624).

PR #83, opened 2024-02-14
Use time.Time() instead of float64

### LINKED ISSUE #76: Refactor the time in API to use time.Time instead of float64
We can use `time.Time#Sub` and `time.Time#Add` using duration to implement the same functionalities.

ranchalp:
Closes #76

### DIFF
--- a/f3/chain.go
+++ b/f3/chain.go
@@ -38,11 +38,6 @@ func (t TipSet) IsZero() bool {
 	return t.Epoch == 0 && t.CID.IsZero()
 }
 
-// Compares tipsets for equality.
-func (t *TipSet) Eq(other *TipSet) bool {
-	return t.Epoch == other.Epoch && t.CID == other.CID
-}
-
 func (t TipSet) String() string {
 	var b strings.Builder
 	b.Write(t.CID.Bytes())
--- a/f3/chain.go
+++ b/f3/chain.go
@@ -38,6 +38,11 @@ func (t TipSet) IsZero() bool {
 	return t.Epoch == 0 && t.CID.IsZero()
 }
 
+// Compares tipsets for equality.
+func (t *TipSet) Eq(other *TipSet) bool {
+	return t.Epoch == other.Epoch && t.CID == other.CID
+}
+
 func (t TipSet) String() string {
 	var b strings.Builder
 	b.Write(t.CID.Bytes())
--- a/f3/granite.go
+++ b/f3/granite.go
@@ -13,8 +13,9 @@ import (
 )
 
 type GraniteConfig struct {
+	// Initial delay for partial synchrony.
 	Delta time.Duration
-	// Change to delta in each round after the first.
+	// Delta back-off exponent for the round.
 	DeltaBackOffExponent float64
 }
 
@@ -545,7 +546,6 @@ func (i *instance) tryConverge() error {
 		return fmt.Errorf("unexpected phase %s, expected %s", i.phase, CONVERGE_PHASE)
 	}
 	timeoutExpired := atOrAfter(i.host.Time(), i.phaseTimeout)
-
 	if !timeoutExpired {
 		return nil
 	}
@@ -744,7 +744,8 @@ func (i *instance) broadcast(round uint64, step Phase, value ECChain, ticket Tic
 // The delay duration increases with each round.
 // Returns the absolute time at which the alarm will fire.
 func (i *instance) alarmAfterSynchrony() time.Time {
-	delta := i.config.Delta * time.Duration(math.Pow(i.config.DeltaBackOffExponent, float64(i.round)))
+	delta := time.Duration(float64(i.config.Delta) *
+		math.Pow(i.config.DeltaBackOffExponent, float64(i.round)))
 	timeout := i.host.Time().Add(delta)
 	i.host.SetAlarm(i.participantID, timeout)
 	return timeout
@@ -1070,6 +1071,7 @@ func hasWeakQuorum(part, total *StoragePower) bool {
 	return part.Cmp(weakThreshold) > 0
 }
 
-func atOrAfter(hostTime time.Time, t time.Time) bool {
-	return hostTime.After(t) || hostTime.Equal(t)
+// Tests whether lhs is equal to or greater than rhs.
+func atOrAfter(lhs time.Time, rhs time.Time) bool {
+	return lhs.After(rhs) || lhs.Equal(rhs)
 }
--- a/f3/granite.go
+++ b/f3/granite.go
@@ -491,9 +491,8 @@ func (i *instance) tryQuality() error {
 	}
 	// Wait either for a strong quorum that agree on our proposal,
 	// or for the timeout to expire.
-
 	foundQuorum := i.quality.HasStrongQuorumFor(i.proposal.Head())
-	timeoutExpired := i.host.Time().Equal(i.phaseTimeout) || i.host.Time().After(i.phaseTimeout)
+	timeoutExpired := atOrAfter(i.host.Time(), i.phaseTimeout)
 
 	if foundQuorum {
 		// Keep current proposal.
@@ -545,7 +544,7 @@ func (i *instance) tryConverge() error {
 	if i.phase != CONVERGE_PHASE {
 		return fmt.Errorf("unexpected phase %s, expected %s", i.phase, CONVERGE_PHASE)
 	}
-	timeoutExpired := i.host.Time().After(i.phaseTimeout)
+	timeoutExpired := atOrAfter(i.host.Time(), i.phaseTimeout)
 
 	if !timeoutExpired {
 		return nil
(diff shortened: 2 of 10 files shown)


test/constants.go
```diff
@@ -1,15 +1,18 @@
...
+	"time"
+)
 
 // Configuration constants used across most tests.
 // These values are not intended to reflect real-world conditions.
 // The latency and delta values are similar in order to stress "slow" message paths and interleaving.
 // The values are not appropriate for benchmarks.
-const DELTA = 0.400
-const DELTA_EXTRA = 0.100
+const DELTA = time.Duration(0.400 * float64(time.Second))
```
Kubuxu, line 12: `400 * time.Milisecond` would be better

f3/granite.go
```diff
@@ -563,7 +566,8 @@ func (i *instance) tryConverge() error {
 	if i.phase != CONVERGE_PHASE {
 		return fmt.Errorf("unexpected phase %s, expected %s", i.phase, CONVERGE_PHASE)
 	}
-	timeoutExpired := i.host.Time() >= i.phaseTimeout
+	timeoutExpired := i.host.Time().After(i.phaseTimeout) || i.host.Time().Equal(i.phaseTimeout)
```
Kubuxu, line 569: I don't think we need to check the `Equal` case. The local time uses nanosecond precision so the probability of it happening is insignificant.

ranchalp:
Done, but took a minor readjustment of the `Time()` function of the simulation. Thanks for the comments!

sim/network.go
```diff
@@ -86,19 +87,19 @@ func (n *Network) Broadcast(msg *f3.GMessage) {
...
 		}
 	}
 }
 
 ///// Clock interface
 
-func (n *Network) Time() float64 {
-	return n.clock
+func (n *Network) Time() time.Time {
+	return n.clock.Add(time.Duration(1 * time.Millisecond)) // To ensure not exactly equal time in simulation
```
Kubuxu, line 99: Was this required for things to work?

ranchalp: Yes, either this or placing again the `.Equal` comparisons on times

Kubuxu: We can have the time always moving forward by having a counter (incremented on every call to `Time()`) which adds a corresponding number of nanoseconds to the Time. This should resolve it, and it would resolve similar issues.

ranchalp: Done.

f3/granite.go
anorth: I think it's relatively important not to rely on time passing between whatever host system triggered the alarm (potentially at exactly the right time) and reaching this code. I am going to guess that in simulation the clock has not moved forward by the time this executes.

f3sim.go
```diff
@@ -34,10 +34,10 @@ func main() {
 			LatencyMean: *latencyMean,
 		}
 		graniteConfig := f3.GraniteConfig{
-			Delta:          *graniteDelta,
+			Delta:          time.Duration(*graniteDelta),
```
anorth, line 37: These all need to be multiplied by `time.Second`. I'm actually a bit confused why this type-checked ok at all

sim/latency.go
```diff
@@ -20,8 +21,8 @@ func NewLogNormal(seed int64, mean float64) *LogNormalLatency {
 	return &LogNormalLatency{rng: rng, mean: mean}
 }
 
-func (l *LogNormalLatency) Sample() float64 {
+func (l *LogNormalLatency) Sample() time.Duration {
 	norm := l.rng.NormFloat64()
 	lognorm := math.Exp(norm)
-	return lognorm * l.mean
+	return time.Duration(lognorm * l.mean)
```
anorth, line 27: I think this has amounted to changing the units (dividing by nano). The `mean` argument to NewLogNormal should be a duration too, then whatever conversion you make of that to a float needs to be reversed here.

sim/network.go
```diff
@@ -57,7 +60,7 @@ func NewNetwork(latency LatencyModel, traceLevel int) *Network {
 		participants:               map[f3.ActorID]f3.Receiver{},
 		participantIDs:             []f3.ActorID{},
 		queue:                      messageQueue{},
-		clock:                      0,
+		clock:                      time.Now(),
```
anorth, line 63: Please continue to use zero, or some zero-like value, rather than something that depends on the external world.

sim/network.go
```diff
@@ -86,19 +89,21 @@ func (n *Network) Broadcast(msg *f3.GMessage) {
...
 }
 
 ///// Clock interface
 
-func (n *Network) Time() float64 {
-	return n.clock
+func (n *Network) Time() time.Time {
+	counterDuration := time.Duration(n.counter * int(time.Nanosecond))
+	n.counter++
+	return n.clock.Add(counterDuration) // To ensure not exactly equal time in simulation
```
anorth, line 103: Please remove this hack. I think the problem is in the changes to comparisons made in the protocol code.

sim/network.go
```diff
@@ -260,7 +265,7 @@ func (n *Network) Tick(adv AdversaryReceiver) (bool, error) {
 
 func (n *Network) log(level int, format string, args ...interface{}) {
 	if level <= n.traceLevel {
-		fmt.Printf("net [%.3f]: ", n.clock)
+		fmt.Printf("net [%s]: ", n.clock.Format("2006-01-02 15:04:05.000"))
```
anorth, line 268: Please revert this to just being in seconds since the simulation started (ideally because time starts at zero)

test/constants.go
```diff
@@ -1,15 +1,18 @@
...
+	"time"
+)
 
 // Configuration constants used across most tests.
 // These values are not intended to reflect real-world conditions.
 // The latency and delta values are similar in order to stress "slow" message paths and interleaving.
 // The values are not appropriate for benchmarks.
-const DELTA = 0.400
-const DELTA_EXTRA = 0.100
+const DELTA = time.Duration(400 * float64(time.Millisecond))
```
anorth, line 12: I don't think the `float64` are needed here, everything is integral.

sim/network.go
anorth: No I think that's terrible

f3/granite.go
```diff
@@ -496,7 +499,7 @@ func (i *instance) tryQuality() error {
 	// Wait either for a strong quorum that agree on our proposal,
 	// or for the timeout to expire.
 	foundQuorum := i.quality.HasStrongQuorumAgreement(i.proposal.Head().CID)
-	timeoutExpired := i.host.Time() >= i.phaseTimeout
+	timeoutExpired := i.host.Time().After(i.phaseTimeout)
```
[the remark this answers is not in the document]
ranchalp: Done but forgot to signal here.

f3/granite.go
```diff
@@ -777,22 +781,26 @@ func (i *instance) broadcast(round uint64, step Phase, value ECChain, ticket Tic
 // Sets an alarm to be delivered after a synchrony delay.
 // The delay duration increases with each round.
 // Returns the absolute time at which the alarm will fire.
-func (i *instance) alarmAfterSynchrony(payload string) float64 {
-	timeout := i.host.Time() + i.config.Delta
-
+func (i *instance) alarmAfterSynchrony(payload string) time.Time {
+	// Assuming i.config.Delta and i.config.DeltaExtra are time.Duration
```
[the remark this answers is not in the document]

ranchalp:
@anorth Also ready for review and probably merge (but change base branch first)

f3/chain.go
```diff
@@ -38,6 +38,11 @@ func (t TipSet) IsZero() bool {
 	return t.Epoch == 0 && t.CID.IsZero()
 }
 
+// Compares tipsets for equality.
+func (t *TipSet) Eq(other *TipSet) bool {
+	return t.Epoch == other.Epoch && t.CID == other.CID
+}
+
```
anorth, line 45: This might be stray from a rebase. I don't think we need it, I removed it in #79. We can just use `==`.

f3/granite.go
```diff
@@ -751,25 +743,9 @@ func (i *instance) broadcast(round uint64, step Phase, value ECChain, ticket Tic
...
-
-	timeout := i.host.Time() + delta
-	if i.round > 0 && i.round%5 == 0 && i.phase == CONVERGE_PHASE {
-		// Wait for clock tick assuming synchronized clocks
-		timeToNextClockTick := i.config.ClockTickDelta - math.Mod(timeout, i.config.ClockTickDelta)
-		timeout += timeToNextClockTick
-	}
-
+func (i *instance) alarmAfterSynchrony() time.Time {
+	delta := i.config.Delta * time.Duration(math.Pow(i.config.DeltaBackOffExponent, float64(i.round)))
```
anorth, line 747: This early conversion to duration will truncate `1.3` down to `1`, etc.

sim/network.go
```diff
@@ -55,12 +57,14 @@ type Network struct {
 }
 
 func NewNetwork(latency LatencyModel, traceLevel int, sb SigningBacked, nn f3.NetworkName) *Network {
+	currentTime := time.Now()
```
anorth, line 60: Replace with the zero `time.Time{}` for complete determinism, and no need for startTime.

sim/latency.go
anorth: Fixed

anorth (approved):
I pushed a commit to address the comments.

PR #86, opened 2024-02-15
Reorganize packages

Kubuxu:


Kubuxu:
I'm not in love with the `module` package name as it results in the imported package being called only module and the struct `module.Module` which can be confusing (no reference to f3),

Stebalien:
Sub-packages are nice, but why `gpbft` and why not move some of that logic to the top level?

ranchalp:
I have been thinking for a while that gpbft should be a separated package from the one that uses gpbft and carries the multi-instance consensus logic (participant). The "top" logic (although I woul rather them be separate and not just a sub-package of each other) should be the latter.

Kubuxu:
I don't remember why exactly `gpbft`but IIRC, it is an acronym that explains the consensus.

ranchalp:
> I don't remember why exactly gpbft but IIRC, it is an acronym that explains the consensus.

Technically we should at some point change the name everywhere from Granite to GossiPBFT, and that's where gpbft comes from.

Kubuxu:
> Why not move some of that logic to the top level?
Should we move the module package to the top level?

ranchalp (approved):

anorth (approved):

Kubuxu:
I don't love the fact that f3.Module exposes gpbft interfaces but we can clean this up in followup if we want to.
Ready for review.

PR #87, opened 2024-02-15
Simplify synchronization of participants to simple exponential back-off

ranchalp:
Rolls forward accidental merge #82

### DIFF
--- a/f3/granite.go
+++ b/f3/granite.go
@@ -14,14 +14,14 @@ import (
 type GraniteConfig struct {
 	// Initial delay for partial synchrony.
 	Delta float64
-	// Change to delta in each round after the first.
-	DeltaRate float64
 	// Absolute extra value to add to delta after the first few unsuccessful rounds.
 	DeltaExtra float64
-	// Time to next clock tick (according to local clock)
+	// Time in between clock ticks
 	// Used in timeouts in order to synchronize participants
-	// (by not carrying significant delays from previous/starting rounds)
-	ClockTickDelta float64
+	// (by attempting to not carry significant delays from previous/starting rounds)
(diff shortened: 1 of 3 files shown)


f3/granite.go
```diff
@@ -798,14 +795,14 @@ func (i *instance) alarmAfterSynchrony() float64 {
...
 
 	if i.round >= 5 {
-		delta *= i.deltaRate
-		i.deltaRate *= i.config.DeltaRate
+		delta = delta * math.Pow(i.config.DeltaBackOffExponent, float64(i.round-4))
 	}
 
 	timeout := i.host.Time() + delta
+
 	if i.round > 0 && i.round%5 == 0 && i.phase == CONVERGE_PHASE {
```
anorth, line 803: Carrying conversation from https://github.com/filecoin-project/go-f3/pull/82#discussion_r1491628191

> Would it make more sense in your opinion that we signal with a parameter in the call from CONVERGE that the round is starting?

Yes, I think a parameter like `waitForNextClockBoundary` or similar would make this clearer here, and at the call sites.

> I do realize now that the spec does. not specify every 5th round but specifically in the 5th round. I think it makes more sense every 5th round, so should update the spec instead in this case.

This needs more thought at the spec level. I don't think every 5th round makes sense, though I also don't think once is particularly good either. I'll start a discussion in Slack, let's resolve something there, then come here to fix it.

ranchalp: waitForNextClockBoundary done, and for the slack discussion I answered there.

ranchalp:
@anorth ready for review

anorth (approved):

PR #91, opened 2024-02-19
Express message justification rules as data rather than procedural code

anorth:
This expression of the justification rules is easier to mentally check, and ensures that every field of the justification is always checked.

### DIFF
--- a/gpbft/gpbft.go
+++ b/gpbft/gpbft.go
@@ -386,69 +386,54 @@ func (i *instance) isJustified(msg *GMessage) error {
 		}
 		return nil
 	}
-
+	// Require justification for all other messages.
 	if msg.Justification == nil {
 		return fmt.Errorf("message for phase %v round %v has no justification", msg.Vote.Step, msg.Vote.Round)
 	}
+	// Always check that the justification is for the same instance.
 	if msg.Vote.Instance != msg.Justification.Vote.Instance {
 		return fmt.Errorf("message with instanceID %v has evidence from instanceID: %v", msg.Vote.Instance, msg.Justification.Vote.Instance)
 	}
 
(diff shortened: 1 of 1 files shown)


f3/granite.go
```diff
@@ -394,69 +394,54 @@ func (i *instance) isJustified(msg *GMessage) error {
...
+	// Anything else is disallowed.
+	expectations := map[Phase]map[Phase]struct {
+		Round uint64
+		Value ECChain
+	}{
+		// CONVERGE is justified by a strong quorum of COMMIT for bottom,
+		// or a strong quorum of PREPARE for the same value, from the previous round.
+		CONVERGE_PHASE: {
+			COMMIT_PHASE:  {msg.Vote.Round - 1, ECChain{}},
+			PREPARE_PHASE: {msg.Vote.Round - 1, msg.Vote.Value},
```
ranchalp, line 418: The previous code made sure that the `PREPARE` was not for the zero value, but actually that is ok to have, so this is good.

f3/granite.go
```diff
@@ -677,6 +662,11 @@ func (i *instance) beginDecide(round uint64) {
 		panic("beginDecide with no strong quorum for value")
 	}
 
+	// DECIDE messages always specify round = 0.
```
ranchalp, line 665: If they specify round=0, why do we have the special case for `math.MaxUint64`?

ranchalp (approved):

f3/granite.go
anorth: It's because the DECIDE round does not match the justification COMMIT round, nor can be calculated from it. So in the expectations it's a sentinel to say "don't check the justifying COMMIT round".

PR #97, opened 2024-02-20
Make message validate an exported API of instance and participant, to be called before receiving a message.

### LINKED ISSUE #40: Move message validation and signature checking to be synchronous with Receive

anorth:
@Kubuxu please let me know if this will likely work for your integration.

A complication is that message can only be validated for the currently executing instance. Messages for prior instances can be ignored by this one (but should probably still propagate up to some bound). Messages for future instances can't be validated but must be held and re-delivered when the appropriate Granite instance begins. The Participant used to have a message pool in anticipation of this, but I think that responsibility should be pushed out to the integration instead.

Closes #40.

### DIFF
--- a/gpbft/gpbft.go
+++ b/gpbft/gpbft.go
@@ -117,7 +117,6 @@ func (p Payload) MarshalForSigning(nn NetworkName) []byte {
 }
 
 func (m GMessage) String() string {
-	// FIXME This needs value receiver to work, for reasons I cannot figure out.
 	return fmt.Sprintf("%s{%d}(%d %s)", m.Vote.Step, m.Vote.Instance, m.Vote.Round, &m.Vote.Value)
 }
 
@@ -216,11 +215,25 @@ func (i *instance) Start() error {
 	return i.drainInbox()
 }
 
-// Receives a new acceptable chain and updates its current acceptable chain.
-func (i *instance) receiveAcceptable(chain ECChain) {
-	i.acceptable = chain
+// Receives a new chain, and updates its current chain if the received one is acceptable
+// (i.e. if it extends the current acceptable).
+func (i *instance) ReceiveAcceptable(chain ECChain) {
+	if chain.HasPrefix(i.acceptable) {
+		i.acceptable = chain
+	}
 }
 
+// Checks whether a message is valid.
+// An invalid message can never become valid, so may be dropped.
+// This method is read-only and inspects only immutable state, so should be safe to invoke
+// concurrently.
+func (i *instance) Validate(msg *GMessage) error {
+	return i.validateMessage(msg)
+}
+
+// Receives a validated message.
+// This method will not attempt to validate the message, the caller must ensure the message
+// is valid before calling this method.
 func (i *instance) Receive(msg *GMessage) error {
 	if i.terminated() {
 		return fmt.Errorf("senders message after decision")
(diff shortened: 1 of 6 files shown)


ranchalp:
IMO (could be wrong, but conceptually) responsability of buffering and delivering messages to a future instance belongs to the multi-instance consensus.

gpbft/gpbft.go
```diff
@@ -341,120 +343,130 @@ func (i *instance) tryCompletePhase() error {
...
-	return i.verifyJustification(msg.Justification)
-}
 
-func (i *instance) verifyJustification(justification *Justification) error {
-	power := NewStoragePower(0)
-	signers := make([]PubKey, 0)
-	if err := justification.Signers.ForEach(func(bit uint64) error {
-		if int(bit) >= len(i.powerTable.Entries) {
-			return fmt.Errorf("invalid signer index: %d", bit)
+		// Check justification signature.
```
ranchalp, line 448: What's the rationale for not calling `verifyJustification` anymore?

gpbft/gpbft.go
```diff
@@ -216,11 +215,25 @@ func (i *instance) Start() error {
 	return i.drainInbox()
 }
 
-// Receives a new acceptable chain and updates its current acceptable chain.
-func (i *instance) receiveAcceptable(chain ECChain) {
-	i.acceptable = chain
+// Receives a new chain that is acceptable to the host, and updates its current acceptable chain
```
ranchalp, line 218:
```suggestion
// Receives a new chain, and updates its current chain if the received one is acceptable (i.e. if it extends the current acceptable)
```

gpbft/participant.go
```diff
@@ -59,26 +61,45 @@ func (p *Participant) ReceiveCanonicalChain(chain ECChain, power PowerTable, bea
 // Receives a new EC chain, and notifies the current instance if it extends its current acceptable chain.
```
ranchalp, line 61:
```suggestion
// Receives a new EC chain, and notifies the current instance.
```

gpbft/participant.go
```diff
@@ -59,26 +61,45 @@ func (p *Participant) ReceiveCanonicalChain(chain ECChain, power PowerTable, bea
...
 // This modifies the set of valid values for the current instance.
```
ranchalp, line 62:
```suggestion
// This may modify the set of valid values for the current instance.
```

ranchalp (approved):

gpbft/gpbft.go
anorth: It's still doing the same work, so you mean rationale for not breaking out verifyJustification into a separate function?

It's a close thing, and low impact, so this is all just a matter of design sense in the moment. This change makes it clear that justification is just another part of message validation, rather than different like it used to be with implicit justification. The same question about breaking out a function might be asked of each of checking the vote semantics, the vote signature, the justification semantics, and justification signature. I did consider breaking them all out, but:
- this is just straight line code with very little branching
- the functions are all part of a logical single operation of validation, and don't mean much outside that; they're not re-usable except in another context that also calls all of them
- the functions would each have only one call site
- the functions should be placed spatially next to each other to reflect their one-ness
- breaking out functions would just make the code take more vertical space and add the noise of parameters and return values.

If someone else had done this change and broken them all up, I wouldn't object strongly. But I decided to go with the simple straight line code since I couldn't see any advantage to factoring out.

Kubuxu (commented):
SGWM

PR #100, opened 2024-02-26
Add API for participant to notify host of decisions, with signed justification.

anorth:
This is a part of extending to multiple instances #27. Adds a callback API for the GPBFT participant to notify the host of a finality decision. Internally, the instance still terminates by setting appropriate state, and the outer participant checks for termination after each message/alarm. This prevents unintended re-entrancy into the instance.

Following this change through to the simulation framework was surprisingly involved. It now better represents what the real host integration might look like, but still only handles a single instance.

### DIFF
--- a/gpbft/api.go
+++ b/gpbft/api.go
@@ -72,3 +72,10 @@ type Host interface {
 	// Logs a message at the "logic" level
 	Log(format string, args ...interface{})
 }
+
+type DecisionReceiver interface {
+	// Receives a finality decision from the instance, with signatures from a strong quorum
+	// of participants justifying it.
+	// The decision payload always has round = 0 and step = DECIDE.
+	ReceiveDecision(participant ActorID, decision Justification)
+}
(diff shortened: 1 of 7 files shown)


anorth:
As part of this I finally let go of being able to assert which round terminated in tests. It's not generally part of the decision output (because it's not always well defined).

ranchalp (approved):
LGTM

gpbft/api.go
```diff
@@ -72,3 +72,10 @@ type Host interface {
 	// Logs a message at the "logic" level
 	Log(format string, args ...interface{})
 }
+
+type DecisionReceiver interface {
+	// Receives a finality decision from the instance, with signatures from a strong quorum
+	// of participants justifying it.
+	// The decision payload always has round = 0 and step = DECIDE.
```
Kubuxu, line 79: Is it always Round=0? If it is, then I need to update my mental model.

ranchalp: Different nodes can terminate at different rounds but they all need to sign the same payload. Therefore, the payload at least always needs to be the same round (and the default round value seems most appropriate)

PR #101, opened 2024-02-26
Fix message validation to check the justification signers have strong quorum

anorth:
This is highlighting our embarassing lack of unit testing, stemming from origins as just a simulation.


ranchalp (approved):
We are only humans 🙃

PR #102, opened 2024-02-26
Remove abstraction noise around VRF construction and verification

anorth:
The VRF interfaces were intially created before we had any signing (even fake). The ticket payload creation need not be abstracted, and the signing can be directly invoked on the abstract host Signer.


ranchalp (approved):
Good remark. LGTM

PR #104, opened 2024-03-04
Nodes decide on strong quorum of DECIDE regardless of phase

### LINKED ISSUE #74: Decouple collection of DECIDE messages from primary phase transitions
Collection of DECIDE messages should be independent of other phases. A node can terminate given a strong quorum of DECIDE independent of anything else it has received. At the moment, `tryDecide` is only invoked when the node reaches the DECIDE phase, i.e. observed its own strong quorum of COMMITs.

Exercise this in simulation to demonstrate firstly the problem, then the solution. This might be by making an "adversary" that emits a DECIDE immediately.

ranchalp:
Closes #74

### DIFF
--- a/gpbft/gpbft.go
+++ b/gpbft/gpbft.go
@@ -312,7 +312,10 @@ func (i *instance) receiveOne(msg *GMessage) error {
 		}
 	case DECIDE_PHASE:
 		i.decision.Receive(msg.Sender, msg.Vote.Value, msg.Signature)
-		i.tryDecide()
+		if err := i.tryDecide(); err != nil {
+			return fmt.Errorf("failed to decide: %w", err)
+		}
+		
 	default:
 		i.log("unexpected message %v", msg)
 	}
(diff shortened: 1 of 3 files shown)


gpbft/gpbft.go
```diff
@@ -312,6 +312,10 @@ func (i *instance) receiveOne(msg *GMessage) error {
 		}
 	case DECIDE_PHASE:
 		i.decision.Receive(msg.Sender, msg.Vote.Value, msg.Signature)
+		if err := i.tryDecide(); err != nil {
+			return fmt.Errorf("failed to decide: %w", err)
+		}
```
anorth, line 317: That was easy!

test/decide_test.go
```diff
@@ -0,0 +1,40 @@
...
+		HonestCount: 1,
+		LatencyMean: 10 * time.Millisecond, // Near-synchrony
+	}, GraniteConfig(), sim.TraceNone)
+
+	// Create adversarial node
+	value := sm.Base.Extend(sm.CIDGen.Sample())
+	adv := adversary.NewImmediateDecide(99, sm.Network, sm.PowerTable, value)
+
+	// Add the adversarial nodes to the simulation
+	sm.SetAdversary(adv, 3) // Adversary has 30% of 10 total power.
```
anorth, line 24: I think this is 3/4 of the total power, so the comment isn't right.

anorth (commented):
Looks good. I'll tweak comments and then merge.

anorth (approved):

PR #105, opened 2024-03-04
Recover panics at participant entry points

### LINKED ISSUE #11: Replace panics with errors, catch remaining panics at library boundary

ranchalp:
Closes #11 

However, I am not sure what we want to do with the caught panics yet, so just logging it.

### DIFF
--- a/gpbft/participant.go
+++ b/gpbft/participant.go
@@ -7,9 +7,9 @@ import (
 
 // An F3 participant runs repeated instances of Granite to finalise longer chains.
 type Participant struct {
-	id     ActorID
-	config GraniteConfig
-	host   Host
+	id        ActorID
+	config    GraniteConfig
+	host      Host
 	decisions DecisionReceiver
 
 	// Chain to use as input for the next Granite instance.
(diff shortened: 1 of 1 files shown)


anorth (commented):
Without static analysis of where panics are possible, I think we need to guard all the nontrivial methods, including receiving chains and validating messages.

As for what we should do, I suggest creating a new error type so that the caller can use `errors.Is` to detect this panic-type of error, and then return it. We really want the caller to tear down the whole F3 and start over.

ranchalp:
@jsoares thanks for the comments, done.

Kubuxu (approved):

Kubuxu:
Pull in https://github.com/Kubuxu/kyber/commit/c5d6a111d0149c40c0c845a7033c313666dd0a5c for additional tests

anorth (approved):

gpbft/participant.go
```diff
@@ -35,8 +47,14 @@ func (p *Participant) ID() ActorID {
 
 // Fetches the preferred EC chain for the instance and begins the GPBFT protocol.
 func (p *Participant) Start() error {
-	chain, power, beacon := p.host.GetCanonicalChain()
 	var err error
+	defer func() {
+		if r := recover(); r != nil {
+			err = &PanicError{Err: err}
```
anorth, line 53: This doesn't do anything without naming the return value. This just sets a local which is then ignored as the function continues to return the default zero value.

ranchalp: Agreed, I realised of that after a commit during the merge and pushed [another commit](https://github.com/filecoin-project/go-f3/blob/ranchalp/catch-panics/gpbft/participant.go), but I guess somebody (perhaps myself) clicked merge before that commit. Will fix that in a new PR now.

Kubuxu: Sorry for missing this in CR.

PR #112, opened 2024-03-07
Ranchalp/catch panics (Fix wrong return in Start() function)

ranchalp:


ranchalp:
Closing as #111  already addresses this

PR #118, opened 2024-03-12
Scale power in the range 0-0xffff

### LINKED ISSUE #33: Settle representation of voting power, truncation of native power
Confirm if voting power can use uint or needs big.Int. Similar to the concern raised in #22.

Update: expanded scope to include design question about truncation

Stebalien:
When we construct a power table, we now:

1. Scale individual power into the range 0-0xffff.
2. Treat the sum of these _scaled_ powers as the maximum power from the perspective of "quorum". We can't use 0xffff as the "total power" due to compounded rounding errors.

Then, we use this scaled power when determining if we have quorum. However, we _don't_ use it for either:

1. Tickets weight.
2. Determining if an SP is eligible to be included in the power table itself.

fixes #33

### DIFF
--- a/certs/certs.go
+++ b/certs/certs.go
@@ -154,7 +154,13 @@ func verifyFinalityCertificateSignature(verifier gpbft.Verifier, powerTable gpbf
 				"finality certificate for instance %d specifies a signer %d but we only have %d entries in the power table",
 				cert.GPBFTInstance, i, len(powerTable))
 		}
-		signerPowers += scaled[i]
+		power := scaled[i]
+		if power == 0 {
+			return xerrors.Errorf(
+				"finality certificate for instance %d specifies a signer %d but they have no effective power after scaling",
+				cert.GPBFTInstance, powerTable[i].ID)
+		}
+		signerPowers += power
 		signers = append(signers, powerTable[i].PubKey)
 		return nil
 	}); err != nil {
--- a/certs/certs.go
+++ b/certs/certs.go
@@ -141,29 +141,28 @@ func ValidateFinalityCertificates(verifier gpbft.Verifier, network gpbft.Network
 // any other parts of the certificate, just that the _value_ has been signed by a majority of the
 // power.
 func verifyFinalityCertificateSignature(verifier gpbft.Verifier, powerTable gpbft.PowerEntries, nn gpbft.NetworkName, cert FinalityCertificate) error {
-	totalPower := new(gpbft.StoragePower)
-	for i := range powerTable {
-		totalPower = totalPower.Add(totalPower, powerTable[i].Power)
+	scaled, totalScaled, err := powerTable.Scaled()
+	if err != nil {
+		return xerrors.Errorf("failed to scale power table: %w", err)
 	}
 
 	signers := make([]gpbft.PubKey, 0, len(powerTable))
-	signerPowers := new(gpbft.StoragePower)
+	var signerPowers uint16
 	if err := cert.Signers.ForEach(func(i uint64) error {
 		if i >= uint64(len(powerTable)) {
 			return xerrors.Errorf(
 				"finality certificate for instance %d specifies a signer %d but we only have %d entries in the power table",
 				cert.GPBFTInstance, i, len(powerTable))
 		}
-		signer := &powerTable[i]
-		signerPowers = signerPowers.Add(signerPowers, signer.Power)
-		signers = append(signers, signer.PubKey)
+		signerPowers += scaled[i]
+		signers = append(signers, powerTable[i].PubKey)
 		return nil
 	}); err != nil {
 		return err
 	}
 
-	if !gpbft.IsStrongQuorum(signerPowers, totalPower) {
-		return xerrors.Errorf("finality certificate for instance %d has insufficient power: %s < 2/3 %s", cert.GPBFTInstance, signerPowers, totalPower)
+	if !gpbft.IsStrongQuorum(signerPowers, totalScaled) {
+		return xerrors.Errorf("finality certificate for instance %d has insufficient power: %d < 2/3 %d", cert.GPBFTInstance, signerPowers, totalScaled)
 	}
 
 	payload := &gpbft.Payload{
--- a/f3.go
+++ b/f3.go
@@ -3,6 +3,7 @@ package f3
 import (
 	"bytes"
 	"context"
+	"errors"
 
 	"github.com/filecoin-project/go-f3/certstore"
 	"github.com/filecoin-project/go-f3/gpbft"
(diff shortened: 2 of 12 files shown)


Stebalien:
Feedback from call: gpbft should use scaled powers everywhere except the commitment. That means u16 everywhere.

jennijuju:
@Stebalien - are we unblocked from this?

Stebalien:
It's not blocked, just put into the background because it's less critical than finality exchange.

certs/certs.go
```diff
@@ -141,29 +141,28 @@ func ValidateFinalityCertificates(verifier gpbft.Verifier, network gpbft.Network
 // any other parts of the certificate, just that the _value_ has been signed by a majority of the
 // power.
 func verifyFinalityCertificateSignature(verifier gpbft.Verifier, powerTable gpbft.PowerEntries, nn gpbft.NetworkName, cert FinalityCertificate) error {
-	totalPower := new(gpbft.StoragePower)
-	for i := range powerTable {
-		totalPower = totalPower.Add(totalPower, powerTable[i].Power)
+	scaled, totalScaled, err := powerTable.Scaled()
+	if err != nil {
+		return xerrors.Errorf("failed to scale power table: %w", err)
```
Stebalien, line 146: We scale the power on-demand when validating finality certificates because finality certificates take `PowerEntries` instead of a full `PowerTable`.

certs/certs.go
```diff
@@ -141,29 +141,28 @@ func ValidateFinalityCertificates(verifier gpbft.Verifier, network gpbft.Network
...
 	if err := cert.Signers.ForEach(func(i uint64) error {
 		if i >= uint64(len(powerTable)) {
 			return xerrors.Errorf(
 				"finality certificate for instance %d specifies a signer %d but we only have %d entries in the power table",
 				cert.GPBFTInstance, i, len(powerTable))
 		}
-		signer := &powerTable[i]
-		signerPowers = signerPowers.Add(signerPowers, signer.Power)
-		signers = append(signers, signer.PubKey)
+		signerPowers += scaled[i]
```
Stebalien, line 157: Now that we're rounding, SPs can have zero power (rounded down). Reject them here I assume?

gpbft/gpbft.go
```diff
@@ -1308,22 +1307,27 @@ func findFirstPrefixOf(preferred ECChain, candidates []ECChain) ECChain {
...
 
 // Check whether a portion of storage power is a weak quorum of the total
-func hasWeakQuorum(part, total *StoragePower) bool {
-	three := NewStoragePower(3)
-
-	weakThreshold := new(StoragePower).Div(total, three)
-	return part.Cmp(weakThreshold) > 0
+func hasWeakQuorum(part, whole uint16) bool {
+	// XXX ceiling!
+	return part > whole/3
```
Stebalien, line 1330: TODO: these calculations aren't quite correct.

gpbft/powertable.go
```diff
@@ -82,24 +107,36 @@ func (p *PowerTable) Add(entries ...PowerEntry) error {
 		default:
 			p.Total.Add(p.Total, entry.Power)
 			p.Entries = append(p.Entries, entry)
+			p.ScaledPower = append(p.ScaledPower, 0)
 			p.Lookup[entry.ID] = len(p.Entries) - 1
 		}
 	}
 	sort.Sort(p)
+	return p.rescale()
```
Stebalien, line 115: This is expensive, but `Add` is only used for testing.

Stebalien: Well, for now we only have testing...

gpbft/gpbft.go
```diff
@@ -1256,7 +1255,7 @@ func (c *convergeState) FindMaxTicketProposal(table PowerTable) ConvergeValue {
 
 	for key, value := range c.values {
 		for _, ticket := range c.tickets[key] {
-			senderPower, _ := table.Get(ticket.Sender)
+			_, senderPower, _ := table.Get(ticket.Sender)
```
Stebalien, line 1258: I use the unscaled power for tickets.

Stebalien:
@Kubuxu this is ready for an initial review. If the general structure looks good, I can fix up the tests.

certs/certs.go
Kubuxu: This touches https://github.com/filecoin-project/go-f3/issues/319 as well. SP's voting eligibility status can change.

Kubuxu:
The idea and first code seem sane.

certs/certs.go
```diff
@@ -141,29 +141,34 @@ func ValidateFinalityCertificates(verifier gpbft.Verifier, network gpbft.Network
...
 			return xerrors.Errorf(
 				"finality certificate for instance %d specifies a signer %d but we only have %d entries in the power table",
 				cert.GPBFTInstance, i, len(powerTable))
 		}
-		signer := &powerTable[i]
-		signerPowers = signerPowers.Add(signerPowers, signer.Power)
-		signers = append(signers, signer.PubKey)
+		power := scaled[i]
+		if power == 0 {
+			return xerrors.Errorf(
```
Stebalien, line 159: needs test.

gpbft/gpbft.go
```diff
@@ -555,20 +555,24 @@ func ValidateMessage(powerTable *PowerTable, beacon []byte, host Host, msg *GMes
...
+		var justificationPower uint16
 		signers := make([]PubKey, 0)
 		if err := msg.Justification.Signers.ForEach(func(bit uint64) error {
 			if int(bit) >= len(powerTable.Entries) {
 				return fmt.Errorf("invalid signer index: %d", bit)
 			}
-			justificationPower.Add(justificationPower, powerTable.Entries[bit].Power)
+			power := powerTable.ScaledPower[bit]
+			if power == 0 {
+				return xerrors.Errorf("signer %d has no power", powerTable.Entries[bit].ID)
```
Stebalien, line 566: needs test

f3.go
```diff
@@ -47,6 +48,9 @@ type clientImpl struct {
 func (mc clientImpl) BroadcastMessage(ctx context.Context, mb *gpbft.MessageBuilder) error {
 	msg, err := mb.Build(mc.nn, mc.SignerWithMarshaler, mc.id)
 	if err != nil {
+		if errors.Is(err, gpbft.ErrNoPower) {
+			return nil
+		}
```
Stebalien, line 53: @Kubuxu make sure the "no power" case makes sense.

gpbft/message_builder.go
```diff
@@ -36,20 +40,20 @@ func (mb *MessageBuilder) SetPayload(p Payload) {
 	mb.payload = p
 }
 
-func (mb MessageBuilder) Payload() Payload {
+func (mb *MessageBuilder) Payload() Payload {
```
Stebalien, line 43: drive-by fix to take receivers by-pointer.

gpbft/gpbft.go
Stebalien: Punted to https://github.com/filecoin-project/go-f3/pull/332.

f3.go
Kubuxu: Yeah, it does.

Kubuxu (approved):
SGWM
I think we might want to commit to this truncated  power table but let's get this PR moving.

Stebalien:
>  I think we might want to commit to this truncated power table but let's get this PR moving.

For security, I don't think we have to do that. For efficient verification inside snarks, we do. But... the idea was to commit to it inside the commitments tree.

PR #119, opened 2024-03-12
Delay next instance until EC chain progresses

### LINKED ISSUE #113: Implement timing/delay for starting subsequent protocol instances
As of #111, the participant will immediately start a new protocol instance after terminating the previous one. In simulation, it can always immediately observe a new EC chain to propose. This will be a poor approach in production.

When the protocol makes good progress, deciding on a recent tipset, there can be no new useful decision until EC produces another tipset and any stabilisation delay has passed. A GPBFT instance would only propose and decide base chain, repeatedly. Instead, the participant should be informed of chain timestamps and delay starting a new instance until a fixed period (30s? a bit more?) has passed since the timestamp of the last finalised tipset. This will also serve to synchronise nodes starting new instances in this happy case.

When the protocol does decide the base chain, due to disagreement among nodes, it can probably start again immediately. We might.

See also #27.

anorth:
Adds the timestamp for next instance as a return value from host's ReceiveDecision callback. This design delegates all knowledge of EC epoch timing and policy about how long to wait for stabilization to the host. Nodes will need to adopt the same policy here to stay in sync but this knowledge doesn't belong inside GBPFT.

Closes #113.

### DIFF
--- a/cmd/f3sim/f3sim.go
+++ b/cmd/f3sim/f3sim.go
@@ -27,9 +27,11 @@ func main() {
 		fmt.Printf("Iteration %d: seed=%d, mean=%f\n", i, seed, *latencyMean)
 
 		simConfig := sim.Config{
-			HonestCount: *participantCount,
-			LatencySeed: *latencySeed,
-			LatencyMean: time.Duration(*latencyMean * float64(time.Second)),
+			HonestCount:          *participantCount,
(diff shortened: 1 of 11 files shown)


ranchalp (approved):
LGTM

Kubuxu (approved):

PR #120, opened 2024-03-12
Fix PanicError construction and sim clock progress with historical messages.

anorth:
Another bug in the panic recovery - this "idiom" is easy to get wrong!

Also other minor fixes discovered during #119.


gpbft/participant.go
```diff
@@ -25,17 +25,13 @@ type Participant struct {
 }
 
 type PanicError struct {
-	Err error
+	Err any
```
Kubuxu, line 28: Why the change to `any`?

Kubuxu (approved):

Issue #122, opened 2024-03-12
Handle messages for future instances that can't be validated


anorth:
Message validation depends on information from the base tipset for an instance, such as the power table and signing keys. This means that when participants drift a little out of sync, some participant could receive a message for a future instance which it cannot yet validate. How should we handle this?

Message validation is currently exposed as a synchronous method so that the network layer can validate most messages as they arrive, and refuse to propagate invalid ones. If a message can't be validated (yet), should the network layer also refuse to propagate it? This could hamper propagation for fast running nodes.

The message cannot be rejected locally as it may be valid and therefore necessary for progress. area for progress. It must be queued somewhere for revalidation when the appropriate instance is started.

At the moment, the API punts that local queuing out to integration, expecting it to re-deliver them. This might not be the best (see [discussion](https://github.com/filecoin-project/go-f3/pull/97#issuecomment-1960320676)). This also causes complexity in simulation.

What if the participant locally queues messages for future instances, locally revalidates them when those instances start? The network layer can't get information about those message validity to inform propagation, but is it too late for that anyway? It also opens up some denial of service risk. This is something we have to handle somewhere and there is an existing issue about it #12.

ranchalp:
> What if the participant locally queues messages for future instances, locally revalidates them when those instances start?

I would do this to start with (that is how I proposed in my now deprecated PR on multi-instance consensus). DOS attacks are always a possibility even if we do not propagate, but yes we can mitigate then if we modify the network-layer propagation to only propagate upon validation. Either way, locally buffering at the f3 level seems sensible as a first step, to be extended by network propagation control.

jsoares:
Outcome of sync discussion: Let's merge #124 and create a tracking issue for later replacement. @ranchalp will look at how Aptos/Sui/Cosmos are handling this and draft a more formal proposal, but we're leaning towards having a queue on the sender instead and a request-response/broadcast protocol to suppress the DoS vector.

anorth:
Closing with #124 and in favour of #130 (allowing nodes to drop messages that can't be validated).

Issue #12, opened 2023-12-07
Implement basic denial-of-service resistence


anorth:
Implement some basic checks on the resources invoked when receiving messages.

```[tasklist]
### Tasks
- [x] Minimise work done validating or storing equivocating messages from one sender
- [x] Detect and ignore semantically-duplicate messages
- [x] https://github.com/filecoin-project/go-f3/issues/169
- [x] https://github.com/filecoin-project/go-f3/issues/153
- [x] https://github.com/filecoin-project/go-f3/issues/240
- [x] https://github.com/filecoin-project/go-f3/issues/151
- [x] drop messages from old rounds and phases that have already passed
- [x] #208
- [ ] #303
```

```[tasklist]
### Tasks
```

Stebalien:
Also, should we drop messages from old rounds and phases that have already passed?

anorth:
> Also, should we drop messages from old rounds and phases that have already passed?

We could, but I don't think it's particularly worthwhile. Old _instances_ yes (already the case).

anorth:
> Also, should we drop messages from old rounds and phases that have already passed?

Done for QUALITY, CONVERGE, and PREPARE. The COMMIT phase stays open, and a message can still cause a decision even if the node moved on.

anorth:
>  Minimise work done validating or storing equivocating messages from one sender
> Detect and ignore semantically-duplicate messages

The quorum state for each step maintains a set of senders, and ignores anything after the first message for that (sender, step).

anorth:
I think this is complete after #208.

- With only one message per (sender, round, phase) retained or causing any state allocation, work is limited to what can be received
- Since almost all messages require justification from 2/3 of power, that state or work is limited to relatively small amounts, and reflecting actual protocol progress
- Messages that do not require justification are QUALITY (only valid with round zero) and COMMIT for bottom. These latter are limited to a configurable number of rounds ahead of present.
- As of #300, I believe we don't allocate any state for dropped messages

I briefly looked through all other slices and maps in participant/instance state, and believe them all to be limited either concretely or by assumptions (e.g. size of power table is practically limited well below DOS values, number of rounds is limited by time taken for 2/3 of power to progress, number of instance committees is a configured value)

jennijuju:
as #208 is merged can i close this issue?

PR #124, opened 2024-03-14
Add local queue for messages for future instances

anorth:
This restores a queue of messages for future instances in the participant. Ensuring the delivery of these messages fixed the divergence in multi-instance tests which can now run for significantly longer (no divergence detected yet).

I'm not sure what the network layer should do with messages that can't be validated yet: see #122. Maybe we can close that with this PR, but I'm leaving it open for discussion about the integration problem.

These queues also introduce denial of service risk. See #12.

### DIFF
--- a/gpbft/participant.go
+++ b/gpbft/participant.go
@@ -2,7 +2,6 @@ package gpbft
 
 import (
 	"fmt"
-	"golang.org/x/xerrors"
 )
 
 // An F3 participant runs repeated instances of Granite to finalise longer chains.
@@ -15,6 +14,10 @@ type Participant struct {
 	nextInstance uint64
 	// Current Granite instance.
 	granite *instance
+	// Messages queued for future instances.
+	// Unbounded queues are a denial-of-service risk.
+	// See https://github.com/filecoin-project/go-f3/issues/12
+	mqueue map[uint64][]*GMessage
 	// The output from the last terminated Granite instance.
 	finalised *Justification
 	// The round number during which the last instance was terminated.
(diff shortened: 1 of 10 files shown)


gpbft/participant.go
```diff
@@ -91,31 +94,38 @@ func (p *Participant) ValidateMessage(msg *GMessage) (checked bool, err error) {
...
-// The message is delivered to the Granite instance if it is for the current instance,
-// else it is dropped.
-// This method *does not check message validity*.
-// The message must have been previously validated with ValidateMessage indicating success.
-// Since messages for future instances cannot be validated, a valid message
-// can only be for the current or some previous instance (hence dropping if not current).
-// Returns whether the message was accepted for the instance, and an error if it could not be
-// processed.
-func (p *Participant) ReceiveMessage(msg *GMessage) (accepted bool, err error) {
+// If the message is for the current grant instance, it is delivered immediately.
```
ranchalp, line 97: current grant? (GossiPBFT?)

gpbft/participant.go
```diff
@@ -91,31 +94,38 @@ func (p *Participant) ValidateMessage(msg *GMessage) (checked bool, err error) {
...
+// If it is for a future instance, it is locally queued, to be delivered when that instance begins.
+// If it is for a previous instance, it is dropped.
+//
+// This method checks message validity only if the validated parameter is false.
+// Validity can only be checked when the messages instance begins.
+// Returns whether the message was for the current instance, and an error if it was invalid or
+// could not be processed.
+func (p *Participant) ReceiveMessage(msg *GMessage, validated bool) (accepted bool, err error) {
```
ranchalp, line 105: Should we include sanity checks that we can include for future instances directly here? (Mitigate DOS attacks). I.e.: a strong quorum of decide must contain valid signatures (even if we cannot really verify that the signatures constitute a strong quorum yet)

ranchalp (commented):
LGTM but couple minor comments.

gpbft/participant.go
anorth: We can't even verify the signatures yet, because we don't know the keys to use. We could probably use a heuristic but I would prefer to defer thinking hard about that so we can get the basic thing working sooner

Kubuxu:
I don't love this solution; my preference would be to avoid queuing as it is an inherently unreliable mechanism.
I'm saying this because the queue needs to be capped. If that is the case and the queue is crucial for the protocol's functioning, then it becomes an easy DoS vector. Messages placed in the queue are not verified, so they can be easily spammed.

jsoares (approved):

PR #126, opened 2024-03-18
Order chains from QUALITY by height instead of length

### LINKED ISSUE #117: Change quality phase chain ordering to use epoch height instead of chain length
The quorum state `ListStrongQuorumValues` prefers chains that are longer (have more non-empty tipsets), but should be comparing chains by EC epoch ("height") instead. The assertion that multiple chains can't have the same length isn't valid (but would be for same epoch).

ranchalp:
Closes #117

### DIFF
--- a/gpbft/chain.go
+++ b/gpbft/chain.go
@@ -164,22 +164,6 @@ func (c ECChain) HasPrefix(other ECChain) bool {
 	return true
 }
 
-// Checks whether a chain has some prefix (including the head).
-// Always false for a zero value.
-func (c ECChain) HasPrefixAssumingCorrectChains(other ECChain) bool {
-	if c.IsZero() || other.IsZero() {
-		return false
-	}
-
-	for i := len(c) - 1; i >= 0; i-- {
-		if c[i] == other.Head() {
-			return true
-		}
-	}
-
-	return false
-}
-
 // Checks whether a chain has some tipset (including as its base).
 func (c ECChain) HasTipset(t TipSet) bool {
 	for _, t2 := range c {
--- a/gpbft/chain.go
+++ b/gpbft/chain.go
@@ -164,6 +164,22 @@ func (c ECChain) HasPrefix(other ECChain) bool {
 	return true
 }
 
+// Checks whether a chain has some prefix (including the head).
+// Always false for a zero value.
+func (c ECChain) HasPrefixAssumingCorrectChains(other ECChain) bool {
+	if c.IsZero() || other.IsZero() {
+		return false
+	}
+
+	for i := len(c) - 1; i >= 0; i-- {
+		if c[i] == other.Head() {
+			return true
+		}
+	}
+
+	return false
+}
+
 // Checks whether a chain has some tipset (including as its base).
 func (c ECChain) HasTipset(t TipSet) bool {
 	for _, t2 := range c {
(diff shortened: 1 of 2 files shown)


gpbft/chain.go
```diff
@@ -164,6 +164,22 @@ func (c ECChain) HasPrefix(other ECChain) bool {
 	return true
 }
 
+// Checks whether a chain has some prefix (including the head).
+// Always false for a zero value.
+func (c ECChain) HasPrefixAssumingCorrectChains(other ECChain) bool {
```
anorth, line 169: I can't figure out what this is doing. Is it supposed to have the same semantics as HasPrefix but be an optimisation? What is "correct chains"?

gpbft/gpbft.go
```diff
@@ -981,18 +981,21 @@ func (q *quorumState) ListStrongQuorumValues() []ECChain {
...
 	})
-	prevLength := 0
-	for _, v := range withQuorum {
-		if len(v) == prevLength {
-			panic(fmt.Sprintf("multiple chains of length %d with strong quorum", prevLength))
+	if len(withQuorum) <= 1 {
+		return withQuorum
+	}
+
+	// Check that the chains are prefixes of each other.
```
anorth, line 994: This is a different check than the existing one. A direct translation of the existing one would just replace `len(v)` with `v.Head().Epoch`, and I think we should have that check.

I can't work out what the new check is doing, or if we need it. Can you explain it more, and why just checking the Epoch isn't enough?

anorth (commented):
This is more complicated than I was expecting, let's explore why.

For "filter empty tipsets" – I don't think we should ever have zero-value tipsets in an ECChain. We're probably missing some data validation on received messages for this. Did this actually come up in testing.

There is a question to be answered about how null epochs are represented in the ECChain. Two possible options would be:
1. There is a Tipset for every epoch (no gaps), a null epoch has a TipsetID indicating no blocks, or
2. There is a Tipset only for non-null epochs, so there may be gaps in the chain.

I think we should take (2), and an existing comment says "epochs may be missing due to null rounds". So there should be no zero-value tipsets in any ECChain.  A vote for bottom is an empty list, not a list containing a zero value.

gpbft/chain.go
ranchalp: The other function that is written checks every tipset, but that is an overkill if chains are correct (i.e. valid): as soon as the head of one is contained in the chain of the other we can already conclude that they are prefixes, no need to check for all the parents following up.

We can probably just have this function because it seems weird to me that an ECChain that is not correct, but thought that it can be made in a different PR not to make too many things in this one.

ranchalp:
Agreeing on (2). In that case I just need to remove the function filterEmptyTipsets. Pushing in a minute.

gpbft/gpbft.go
ranchalp: Since chains are sorted now by decreasing head epoch number, we can just start with the first and make sure one is a prefix of each other. The previous test was wrong as you noted due to the possibility of null epochs, and just checking the previous head epoch with the subsequent head epoch would not work (which is what I understand by you saying "just replace `len(v)` with `v.Head().Epoch`). Or put differently, what I do here is what I understand as the most natural way of just replacing `len(v)` with `v.Head().Epoch`.

gpbft/chain.go
ranchalp: I just removed this function following your other comment.

anorth:
Thanks. I'm temporarily unsure whether this is a fix after all. See #117 and #135.

anorth:
Closing as #117 was misguided.

PR #129, opened 2024-03-19
Use `go generate` to generate sources

masih:
Use `go:generate` directive to generate sources such that executing `go generate ./...` in root of the project would work.

Set up CI to assert that checked in code is consistent with what `go generate` generates.

Fix typo in CI to trigger builds on the `main` branch.


masih:
@Stebalien if you are a repo admin, we probably also want `Generate` to be a required passing build for PRs

Kubuxu (approved):

Issue #128, opened 2024-03-19
Validate chains received


anorth:
There are some basic validation rules for ECChain values that we should enforce at all entry points (from messages and from the host).
- Epochs must be in order, monotonic (but may have gaps)
- Tipsets must not be zero value (this is also implied by epochs in order, if epoch >0)

Check other fields for validation rules too.

masih:
Apologies in advance for pedantic questions:
* By monotonic, we strictly mean monotonically increasing with optional gaps where repeated epochs are also not allowed. Correct?
* Should `TipSetID`s also be included in validation?

masih:
I have implemented validation based on the preexisting check that asserts there are no repeated epochs. Please let me know if that is unintended.

PR #133, opened 2024-03-26
Implement basic chain and types unit tests

masih:
Add unit tests that assert expected behaviour from the APIs exported by `gpbft` core types.

Relates to #9

### DIFF
--- a/gpbft/chain_test.go
+++ b/gpbft/chain_test.go
@@ -1,0 +1,128 @@
+package gpbft_test
+
+import (
+	"bytes"
+	"encoding/binary"
+	"testing"
+
+	"github.com/filecoin-project/go-f3/gpbft"
+	"github.com/stretchr/testify/require"
+)
+
+func TestTipSet(t *testing.T) {
+	t.Parallel()
+	tests := []struct {
+		name       string
+		subject    gpbft.TipSet
+		wantZero   bool
+		wantString string
+	}{
+		{
+			name:       "zero-value struct is zero",
+			wantZero:   true,
+			wantString: "@0",
+		},
+		{
+			name:       "ZeroTipSet is zero",
+			subject:    gpbft.ZeroTipSet(),
+			wantZero:   true,
+			wantString: "@0",
+		},
+		{
+			name:       "NewTipSet with zero values is zero",
+			subject:    gpbft.NewTipSet(0, gpbft.NewTipSetID(nil)),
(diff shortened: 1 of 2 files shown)


gpbft/chain_test.go
```diff
@@ -0,0 +1,138 @@
...
+		require.False(t, subject.SameBase(subject))
+		require.True(t, subject.Eq(subject))
+		require.True(t, subject.Eq(*new(gpbft.ECChain)))
+		require.Nil(t, subject.Suffix())
+		require.Panics(t, func() { subject.Prefix(0) })
+		require.Panics(t, func() { subject.Base() })
+		require.Panics(t, func() { subject.Head() })
+		require.Equal(t, zeroTipSet, subject.HeadOrZero())
+	})
+	t.Run("NewChain with zero tipset is not zero", func(t *testing.T) {
```
anorth, line 88: I think it should be invariant that there is no zero tipset in a chain. Could you please address https://github.com/filecoin-project/go-f3/issues/128 and then change this into testing those invariants?

If you'd like to defer that to a follow-up PR, I suggest removing this test and replacing it with a link to #128.

gpbft/chain_test.go
```diff
@@ -0,0 +1,138 @@
...
+		require.True(t, subject.HasTipset(zeroTipSet))
+		require.True(t, subject.SameBase(subject))
+		require.True(t, subject.Eq(subject))
+	})
+	t.Run("NewChain with unordered tipSets is error", func(t *testing.T) {
+		subject, err := gpbft.NewChain(gpbft.NewTipSet(2, gpbft.ZeroTipSetID()), zeroTipSet)
+		require.Error(t, err)
+		require.Nil(t, subject)
+	})
+	t.Run("NewChain with ordered duplicate epoch is not error", func(t *testing.T) {
```
anorth, line 105:
```suggestion
	t.Run("NewChain with ordered duplicate epoch is error", func(t *testing.T) {
```

anorth (approved):
LGTM after tweak

gpbft/chain_test.go
masih: Thank you for pointing that out.
I left a TODO in this PR, and will address #128 along with tests in a follow-up PR shortly.

masih: Resolved by https://github.com/filecoin-project/go-f3/pull/138

PR #138, opened 2024-03-27
Validate EC chain invariants on instantiation

### LINKED ISSUE #128: Validate chains received

masih:
EC chain cannot contain a zero-valued tip set. Enforce this in the exported API and return an error when such tip set is specified in the chain.

Add tests to assert the EC chain invariants are met:
* Epochs in chain are weakly and monotonically increasing, where repeated epochs are not allowed but gaps may exist in the epochs.
* Base cannot be a zero-valued tip set, as defined by `TipSet.IsZero`.

Implement `ECChain.Validate` to encapsulate the chain validation logic, and validate all received chains by a `Participant`. Further, strictly refuse to receive zero-valued chains, since a zero-valued chain is valid but not acceptable by by a `Participant`. The `ECChain` constructor is also refactored to reuse the newly encapsulated validation logic.

Additionally, optimise check for base, prefix, and equality based on the invariants listed above. Specifically, avoid reflection when checking for equality for better performance.

Fixes #128

### DIFF
--- a/gpbft/chain.go
+++ b/gpbft/chain.go
@@ -2,9 +2,8 @@ package gpbft
 
 import (
 	"encoding/binary"
-	"fmt"
+	"errors"
 	"io"
-	"reflect"
 	"strconv"
 	"strings"
 )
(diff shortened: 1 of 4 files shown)


anorth (commented):
The validation rules look good, but are not yet enforced widely enough.
- Validate all chains received in message payloads (in Instance::Validate)
- Validate chains received from Host::GetCanonicalChain

anorth (approved):

PR #140, opened 2024-03-28
Refactor `Host` interface to optionally trace logic logs

masih:
Tracing logic logs is not a fundamental function with gpbft algorithm. Instead, it is a crosscutting concern that currently aids one to reason about the logical changes of a participant.

Refactor the core `Host` interface to relax the requirement for implementing the logic log tracing functionality. Introduce a separate `Tracer` interface that when specified logs logic change logs.

This change separates the tracing requirements introduced primarily by the `sim` package from the core gpbft algorithm interfaces. Additionally, it makes mocking the `gpbft` behaviour for testing purposes simpler.

Note that the tracing concept in simulation defers from the "system logging", in that: it is a configurable textual output of a simulation. Therefore, the work here keeps them separate instead of replacing it with a typical logging library.

### DIFF
--- a/gpbft/gpbft.go
+++ b/gpbft/gpbft.go
@@ -164,6 +164,8 @@ type instance struct {
 	acceptable ECChain
 	// Decision state. Collects DECIDE messages until a decision can be made, independently of protocol phases/rounds.
 	decision *quorumState
+	// tracer traces logic logs for debugging and simulation purposes.
+	tracer Tracer
 }
 
 func newInstance(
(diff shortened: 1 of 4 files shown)


gpbft/gpbft.go
```diff
@@ -172,7 +175,7 @@ func newInstance(
 	instanceID uint64,
 	input ECChain,
 	powerTable PowerTable,
-	beacon []byte) (*instance, error) {
+	beacon []byte, tracer Tracer) (*instance, error) {
```
anorth, line 178: Chop down for consistency

anorth (approved):

PR #143, opened 2024-03-29
Upgrade to go `1.21`

masih:
Upgrade to go `1.21`, consistent with the go version currently in-use by Lotus.


masih:
This upgrade would also allow us to run go utilities without prior installation, like [this](https://github.com/filecoin-project/go-f3/pull/142/files#diff-eb0aa5f0960505c8eafaf64dbbe9f49148d1f32635cd3d2e29ad16febe9b78b6R3).

Kubuxu (approved):

Issue #144, opened 2024-03-31
Make ticket comparison in CONVERGE step proportional to weight 


ranchalp:
There should be a way to make the ticket generation used in CONVERGE proportional to the weights of the participant. Not a security issue but reduces the impact of Sybil.  

Credits to Dominic Tarr for finding the bug.

PR #145, opened 2024-04-01
Use weights to CONVERGE VRF ticket comparison

ranchalp:
See #144


gpbft/powertable.go
```diff
@@ -79,6 +79,17 @@ func (p *PowerTable) Get(id ActorID) (*StoragePower, PubKey) {
 	return nil, nil
 }
 
+// GetRelative returns the relative power of a specific participant.
+func (p *PowerTable) GetRelative(id ActorID) *big.Float {
```
anorth, line 83: Ditto comment about no floats. Just return the numerator and denominator both so the caller can use integer arithmetic.

gpbft/gpbft.go
```diff
@@ -1026,35 +1027,47 @@ type convergeState struct {
...
+
 	for cid, value := range c.values {
-		for _, ticket := range c.tickets[cid] {
-			if minTicket == nil || ticket.Compare(minTicket) < 0 {
-				minTicket = ticket
+		for sender, ticket := range c.tickets[cid] {
+			senderPower := table.GetRelative(sender)
+			if senderPower.Sign() <= 0 {
+				continue
+			}
```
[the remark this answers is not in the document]
ranchalp: Addressed.

gpbft/gpbft.go
```diff
@@ -1026,35 +1027,47 @@ type convergeState struct {
...
+
+	c.tickets[key][sender] = ticket
 
 	return nil
 }
 
-func (c *convergeState) findMinTicketProposal() ECChain {
-	var minTicket Ticket
+func (c *convergeState) findMinTicketProposal(table PowerTable) ECChain {
+	var minTicket *big.Float
```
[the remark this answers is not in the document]
ranchalp: Let us just use the max function then instead of min? This way we just multiply. This must be changed in the FIP though. Using min or max has no impact anywhere else.

gpbft/gpbft.go
```diff
@@ -1026,40 +1027,49 @@ type convergeState struct {
 	// Chains indexed by head CID
 	values map[TipSet]ECChain
 	// Tickets provided by proposers of each chain.
-	tickets map[TipSet][]Ticket
+	tickets map[TipSet]map[ActorID]Ticket
```
anorth, line 1030: The inner map is never looked up, only iterated. This suggests it should be a list of tuples instead.

Making it a map also changed the semantics so that an equivocating value from the same sender will be overwritten, rather than both maintained. In this particular case it might be ok to drop a CONVERGE message, but in general throughout, equivocations are both maintained since we don't know which value/s other nodes might have seen. We could change it to preserve the max ticket, but since an upcoming change will use the max ticket _with a valid value_, that's not 100% safe either.

(There's still the general #12 DOS issue to address, but let's do that systematically elsewhere. If cryptography means a sender can't produce two different tickets then great, we still don't need a map here).

gpbft/gpbft.go
```diff
@@ -1026,40 +1027,49 @@ type convergeState struct {
...
 }
 
-func (c *convergeState) findMinTicketProposal() ECChain {
-	var minTicket Ticket
-	var minValue ECChain
+func (c *convergeState) findMaxTicketProposal(table PowerTable) ECChain {
+	var maxTicket *big.Int
+	var maxValue ECChain
+
 	for cid, value := range c.values {
```
anorth, line 1061: I have realised that if the same ticket is used for two different values (equivocating), then the one selected here will be nondeterministic, governed by Go map iteration order. We should break ties deterministically instead. We can use the ECChain value itself to do that if we make them comparable. (You can file an issue for this instead of fixing it here if you want).

gpbft/powertable.go
anorth: This can be removed now.

ranchalp: Done.

gpbft/gpbft.go
ranchalp: > but in general throughout, equivocations are both maintained since we don't know which value/s other nodes might have seen

Change made, but let me clarify on using maps instead of storing tuples:

We do not need to know what other value/s other nodes might have seen (other nodes justify their messages). Whether we store one vote or multiple equivocating votes from the same sender in the quorum state has no effect on the correctness of the protocol. Keeping equivocating messages makes sense though, probably in a different data structure for a later point in time in order to punish misbehavior, but I would not conflate it with the same data structure we use to reason about strong quorums.

By the proofs, having a map and only considering one message per node does not affect correctness. Termination and progress occurs thanks to reaching a round in which there is synchrony and the max ticket is held by a correct. Also, having differing proposal values at the end of CONVERGE cannot lead to inconsistencies other than not deciding in a round, so it is safe for different correct nodes to disagree on what is the max ticket.

gpbft/gpbft.go
ranchalp: I think it is ok to have non-determinism here. If the same ticket is used for two different values then either we get a decision on one of them only or we go to a new round. Eventually there is a round where the max ticket is held by a correct, which will not double vote.

anorth: Ok, thanks. As we discussed in person, this is ok but just needs to be clearly documented in the spec (and ideally here too) so readers understand that.

gpbft/gpbft.go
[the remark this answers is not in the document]
anorth: As discussed, max seems fine. You can go back to min if implemented with truncating integer division, too.

gpbft/gpbft.go
anorth: Thanks for clarifying, and elsewhere where you clarified that the protocol spec and implementation should be changed to explicitly dropping equivocations (tracked as #153). In that case, a map was the right structure after all, but please implement so that a subsequent value from the same actor is dropped rather than overwriting the first one (as this is a consistent policy we can apply everywhere).

ranchalp:
@anorth can we get this approved?

gpbft/gpbft.go
```diff
@@ -1045,40 +1046,53 @@ type convergeState struct {
...
-func (c *convergeState) Receive(value ECChain, ticket Ticket) error {
+func (c *convergeState) Receive(sender ActorID, value ECChain, ticket Ticket) error {
 	if value.IsZero() {
 		return fmt.Errorf("bottom cannot be justified for CONVERGE")
 	}
 	key := value.Key()
 	c.values[key] = value
-	c.tickets[key] = append(c.tickets[key], ticket)
+
+	c.tickets[key] = append(c.tickets[key], ActorTicket{Actor: sender, Ticket: ticket})
```
anorth, line 1072: Please leave a comment referencing https://github.com/filecoin-project/go-f3/issues/12 that we need to drop duplicates rather than accumulate an unbounded list of them.

anorth (approved):

gpbft/gpbft.go
ranchalp: Done, thanks.

Issue #148, opened 2024-04-02
Align tipset identifier with the FIP


anorth:
The FIP at present specifies tipset identifiers as `[]Cid` in message. The code instead uses an opaque string (byte array). CIDs are opqaue to the protocol so there's not a good reason to depend on a CID package or import any semantics. But since CID has a specific IPLD tag, there's a mismatch to be bridged somewhere.

Kubuxu:
Closing given https://github.com/filecoin-project/go-f3/issues/166

PR #149, opened 2024-04-02
Make ECChain opaque [][]byte in GBPFT

### LINKED ISSUE #147: Use the whole chain when keying for proposals
Currently, we only use the head, which leads to confusion between ABC and ABxC chains.
https://filecoinproject.slack.com/archives/C06NTU2DPN1/p1712054917063199

anorth:
Removes all internal structure from ECTipset and ECChain, making them entirely opaque. Adds a key type which just stringifies the bytes, ensuring that the entire value is used as a map key. Note that this doesn't yet exactly match the draft FIP (#148).

The internal structure of chain values now needs to be specified elsewhere, but GPBFT will never introspect it.

This also deleted some useful tests for the consistency of chains. When a new type is introduces at the integration layer, those tests should be restored there instead.

Previous PR description below (just introduces the key type) :
<details>
This is a prototype of using the full ECChain payload as a map key, rather than just head tipset ID. It avoids hashing, but just concatenates the bytes and wraps as a string.

This also changes the TipsetID type to be `[][]byte`, removing an additional serialization step and, by design, making it unusable as a map key. Note that this doesn't yet exactly match the draft FIP (#148).

Review: please critique this design and suggest alternatives. There is an outstanding problem of generating a unique key for `[][]byte` (for both keying and signing) that will incorporate the slice boundaries. Ideas welcome.
</details>

Closes #147.

### DIFF
--- a/gpbft/chain.go
+++ b/gpbft/chain.go
@@ -2,73 +2,25 @@ package gpbft
 
 import (
 	"bytes"
-	"encoding/binary"
+	"encoding/hex"
 	"errors"
-	"io"
-	"strconv"
 	"strings"
 )
 
-// Information about a tipset that is relevant to the F3 protocol.
-// This is a lightweight value type comprising 3 machine words.
-// Fields are exported for CBOR generation, but are opqaue and should not be accessed
-// within the protocol implementation.
-type TipSet struct {
-	// The epoch of the blocks in the tipset.
-	Epoch int64
-	// The identifier of the tipset.
-	ID TipSetID
-}
-
-// Creates a new tipset.
-func NewTipSet(epoch int64, cid TipSetID) TipSet {
-	return TipSet{
-		Epoch: epoch,
-		ID:    cid,
-	}
-}
-
-// Returns a zero value tipset.
-// The zero value is not a meaningful tipset and may be used to represent bottom.
-func ZeroTipSet() TipSet {
-	return TipSet{}
-}
-
-func (t TipSet) IsZero() bool {
-	return t.Epoch == 0 && t.ID.IsZero()
-}
-
-func (t TipSet) Eq(other TipSet) bool {
-	return t.Epoch == other.Epoch && t.ID.Eq(other.ID)
-}
-
-// An identifier for this tipset suitable for use as a map key.
-// This must completely and verifiably determine the tipset data; it is not sufficient to use
-// the block CIDs and rely on external verification of the attached metadata.
-func (t TipSet) Key() string {
-	buf := bytes.Buffer{}
-	buf.Write(t.ID.Bytes())
-	_ = binary.Write(&buf, binary.BigEndian, t.Epoch)
-	return buf.String()
-}
-
-func (t TipSet) String() string {
-	var b strings.Builder
-	b.Write(t.ID.Bytes())
-	b.WriteString("@")
-	b.WriteString(strconv.FormatInt(t.Epoch, 10))
-	return b.String()
-}
-
-func (t TipSet) MarshalForSigning(w io.Writer) {
-	_ = binary.Write(w, binary.BigEndian, t.Epoch)
-	_, _ = w.Write(t.ID.Bytes())
-}
+// Opaque type representing a tipset.
+// This is expected to be:
+// - a canonical sequence of CIDs of block headers identifying a tipset,
+// - a commitment to the resulting power table,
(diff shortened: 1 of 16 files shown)


Stebalien:
What if we just agreed on opaque byte slices and let the client deal with all the serialization and deserialization? This package doesn't actually care about the structure of the data, correct?

I.e.:

```go
type ECChain [][]byte
```

masih:
Is there a specific reason for not removing the redundant data from phases other than COMMIT? That seems like a more optimal solution IMHO.

Stebalien:
> It might care about power table in there

IMO, it shouldn't. `GetCanonicalChain` returns the expected power table.

masih:
> We can try that after this, if we think it's worthwhile.

Thank you, yes I think it is worth it.

anorth:
> type ECChain [][]byte

@Stebalien @Kubuxu @masih I've implemented this now. Please see updated PR description and re-review.

gpbft/chain.go
```diff
@@ -1,61 +1,31 @@
...
-	_ = binary.Write(w, binary.BigEndian, t.Epoch)
-	_, _ = w.Write(t.CID.Bytes())
-}
+// Opaque type representing a tipset.
+// This is expected to be:
+// - a canonical sequence of CIDs of block headers identifying a tipset,
+// - a commitment to the resulting power table,
+// - a commitment to additional derived values.
+// However, GossipPBFT doesn't need to know anything about that structure.
+type TipSet = []byte
```
masih, line 16: I am curious: why type alias instead of type definition?

gpbft/chain.go
```diff
@@ -192,25 +149,29 @@ func (c ECChain) Validate() error {
...
 	}
 	return nil
 }
 
+// Returns an identifier for the chain suitable for use as a map key.
+// This must completely determine the sequence of tipsets in the chain.
+func (c ECChain) Key() ChainKey {
+	buf := bytes.Buffer{}
+	for _, t := range c {
+		buf.Write(t)
```
masih, line 165: Minor comment: since the length of writes is known, grow buffer beforehand.
```suggestion
	var buf bytes.Buffer
	for _, t := range c {
		buf.Grow(len(c))
```

gpbft/types.go
```diff
@@ -31,59 +26,3 @@ func (nn NetworkName) PubSubTopic() string {
...
-// Orders two tipset IDs.
-// Returns -1 if a < b, 0 if a == b, 1 if a > b.
-// This is a lexicographic ordering of the bytes of the IDs.
-func (t TipSetID) Compare(o TipSetID) int {
-	return strings.Compare(t.value, o.value)
-}
-
-func (t TipSetID) Bytes() []byte {
-	return []byte(t.value)
-}
```
masih, line 89: nice 👍

masih (approved):

gpbft/chain.go
```diff
@@ -192,25 +149,29 @@ func (c ECChain) Validate() error {
...
 		}
 	}
 	return nil
 }
 
+// Returns an identifier for the chain suitable for use as a map key.
+// This must completely determine the sequence of tipsets in the chain.
+func (c ECChain) Key() ChainKey {
+	buf := bytes.Buffer{}
+	for _, t := range c {
```
Stebalien, line 164: There needs to be some form of length prefix here as well. Concatenating isn't guaranteed to be unambiguous.

gpbft/chain.go
anorth: cbor-gen was unable to generate correct code for [][]byte. A type definition would have been slightly better, but this isn't a yak I'm prepared to shave.

gpbft/chain.go
anorth: Added lengths.

gpbft/gpbft.go
```diff
@@ -114,7 +114,7 @@ func (p Payload) MarshalForSigning(nn NetworkName) []byte {
 	_ = binary.Write(&buf, binary.BigEndian, p.Round)
 	_ = binary.Write(&buf, binary.BigEndian, p.Step)
 	for _, t := range p.Value {
-		t.MarshalForSigning(&buf)
+		buf.Write(t)
```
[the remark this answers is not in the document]
anorth: Added length prefixes.

Kubuxu (approved):
SGTM

PR #152, opened 2024-04-03
Implement sortable power table and test invariants

### LINKED ISSUE #146: Validate power tables received
Power tables received by the instance should satisfy some invariants.
- Entries are in order by power desc, ID asc
- All entries have positive power

Verify these for all power tables received from the host.

See also #128

masih:
Refine the implementation of power table to utilise the built-in `sort.Interface` for a more concise code.

Change the `Add` receiver to reduce the number of ways by which entry addition can be done. This is to reduce the scope of error across the power table APIs.

Add `Validate` function to power table that asserts its invariants and call it upon receiving a power table instance from the host within gpbft implementation.

Add tests that asserts the validity of a variety of cases, including other power table APIs.

Fixes #146

### DIFF
--- a/gpbft/powertable_test.go
+++ b/gpbft/powertable_test.go
@@ -1,0 +1,200 @@
+package gpbft_test
+
+import (
+	"sort"
+	"testing"
+
+	"github.com/filecoin-project/go-f3/gpbft"
+	"github.com/stretchr/testify/require"
+)
+
+func TestPowerTable(t *testing.T) {
+
+	var (
+		oneValidEntry        = gpbft.PowerEntry{ID: 1413, Power: gpbft.NewStoragePower(1414), PubKey: []byte("fish")}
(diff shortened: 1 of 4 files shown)


gpbft/powertable_test.go
```diff
@@ -0,0 +1,196 @@
...
+			subject := gpbft.NewPowerTable()
+			require.NoError(t, subject.Add(oneValidEntry, anotherValidEntry))
+			require.NoError(t, subject.Validate())
+			requireAddedToPowerTable(t, subject, oneValidEntry)
+			requireAddedToPowerTable(t, subject, anotherValidEntry)
+			require.True(t, sort.IsSorted(subject))
+
+			require.NoError(t, subject.Add(yetAnotherValidEntry))
+			requireAddedToPowerTable(t, subject, yetAnotherValidEntry)
+			require.True(t, sort.IsSorted(subject))
```
anorth, line 50: I suggest checking Validate at the end of each test here, too/instead

anorth (approved):
Great

PR #156, opened 2024-04-14
Test repeated messages do not interfere with consensus

masih:
Implement an adversary that echos all messages it receives with a configurable echo count distribution.

Test that  echoed messages do not interfere with consensus even at adversary power of larger than 1/3 of network size.

### DIFF
--- a/adversary/repeat.go
+++ b/adversary/repeat.go
@@ -1,0 +1,64 @@
+package adversary
+
+import (
+	"github.com/filecoin-project/go-f3/gpbft"
+)
+
+type (
+	Repeat struct {
+		id            gpbft.ActorID
+		host          gpbft.Host
+		echoCountDist CountSampler
+	}
+	CountSampler interface {
+		// TODO replace with math/rand/v2 rand.Source once upgraded to go 1.22.
+		Uint64() uint64
+	}
+)
+
+func NewRepeat(id gpbft.ActorID, host gpbft.Host, echoCountDist CountSampler) *Repeat {
+	return &Repeat{
+		id:            id,
+		host:          host,
+		echoCountDist: echoCountDist,
+	}
+}
+
+func (r *Repeat) ReceiveMessage(msg *gpbft.GMessage, _ bool) (bool, error) {
+	sigPayload := msg.Vote.MarshalForSigning(r.host.NetworkName())
+	_, power, beacon := r.host.GetCanonicalChain()
+	_, pubkey := power.Get(r.id)
+
+	sig, err := r.host.Sign(pubkey, sigPayload)
+	if err != nil {
+		panic(err)
+	}
+
+	var ticket gpbft.Ticket
+	if len(msg.Ticket) != 0 {
+		var err error
+		ticket, err = gpbft.MakeTicket(beacon, msg.Vote.Instance, msg.Vote.Round, pubkey, r.host)
+		if err != nil {
+			panic(err)
+		}
+	}
+	echo := &gpbft.GMessage{
+		Sender:        r.ID(),
+		Vote:          msg.Vote,
+		Signature:     sig,
+		Justification: msg.Justification,
+		Ticket:        ticket,
+	}
+	echoCount := int(r.echoCountDist.Uint64())
+	for i := 0; i < echoCount; i++ {
+		r.host.Broadcast(echo)
+	}
+	return true, nil
+}
+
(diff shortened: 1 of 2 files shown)


Kubuxu (approved):

adversary/repeat.go
```diff
@@ -0,0 +1,64 @@
...
+	Repeat struct {
+		id            gpbft.ActorID
+		host          gpbft.Host
+		echoCountDist CountSampler
+	}
+	CountSampler interface {
+		// TODO replace with math/rand/v2 rand.Source once upgraded to go 1.22.
+		Uint64() uint64
+	}
+)
```
Stebalien, line 17: Style nit: grouping types like this isn't usually done.

adversary/repeat.go
```diff
@@ -0,0 +1,64 @@
...
+
+func NewRepeat(id gpbft.ActorID, host gpbft.Host, echoCountDist CountSampler) *Repeat {
+	return &Repeat{
+		id:            id,
+		host:          host,
+		echoCountDist: echoCountDist,
+	}
+}
+
+func (r *Repeat) ReceiveMessage(msg *gpbft.GMessage, _ bool) (bool, error) {
```
Stebalien, line 27: Can we assert the interface we're conforming to?

masih: Addressed in https://github.com/filecoin-project/go-f3/pull/164

adversary/repeat.go
```diff
@@ -0,0 +1,64 @@
...
+		host          gpbft.Host
+		echoCountDist CountSampler
+	}
+	CountSampler interface {
+		// TODO replace with math/rand/v2 rand.Source once upgraded to go 1.22.
+		Uint64() uint64
+	}
+)
+
+func NewRepeat(id gpbft.ActorID, host gpbft.Host, echoCountDist CountSampler) *Repeat {
```
anorth, line 19: Please add a comment explaining the intent of this adversary. There is some nuance to capture, e.g. that the messages are re-signed by this participant (it's not implementing a network-level duplication). I would say the messages are copied or echo'd by this participant, rather than repeated, though the distinction is fine.

test/repeat_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+	"github.com/filecoin-project/go-f3/adversary"
+	"github.com/filecoin-project/go-f3/sim"
+	"github.com/stretchr/testify/require"
+)
+
+func TestRepeat(t *testing.T) {
+	t.Parallel()
+
+	// Repeated messages should not interfere with consensus. Hence, test with larger than 1/3 adversary power.
```
anorth, line 15: > Repeated messages should not interfere with consensus

This is a bit vague to me. The adversary copycats other participants (all of them, so it thus equivocates). When it has >1/3 of power, it is _necessary_ that it sends messages with votes matching some that it has received, otherwise consensus would not be possible. 

We could say that equivocations should not interfere with consensus up to 1/3 power, but with >1/3 power we should be able to engineer (or fluke) a divergent decision.

To demonstrate that we could detect a divergent decision if there was one, could you add a test where the adversary has enough power to convince two separate quarters of the network to decide different values?

adversary/repeat.go
```diff
@@ -0,0 +1,64 @@
...
+
+type (
+	Repeat struct {
+		id            gpbft.ActorID
+		host          gpbft.Host
+		echoCountDist CountSampler
+	}
+	CountSampler interface {
+		// TODO replace with math/rand/v2 rand.Source once upgraded to go 1.22.
+		Uint64() uint64
```
anorth, line 15: Can you document this? What does the return value mean? It's not clear how I can make one of these that:
- echoes every message once
- echoes some fraction of messages at most once
- echoes every messages a <distribution> number of times
- echoes some fraction of messages <distribution> number of times

... or if some of these are not possible.

test/repeat_test.go
```diff
@@ -0,0 +1,74 @@
...
+				}
+			},
+		},
+		{
+			name: "zipf",
+			echoCountDist: func(repetition int) adversary.CountSampler {
+				rng := rand.New(rand.NewSource(int64(repetition)))
+				return rand.NewZipf(rng, 1.2, 1.0, 100)
+			},
+		},
```
anorth, line 42: Can you add a simple one that echoes every message once?

PR #157, opened 2024-04-14
Get PowerTable from Host during adversary testing

masih:
`Decide` and `Withhold` adversaries require an instance of `Host` and access to its power table. Simplify the test setup by retrieving the power table directly from `Host` instead of taking it as an explicit parameter.


Stebalien (approved):
Code LGTM, but I don't know the history so someone else should make sure this makes sense.

Kubuxu (approved):

anorth (commented):
Thanks!

PR #158, opened 2024-04-15
Remove module placeholder

masih:
I believe the gossipsub integration along with dependency injection module for go-f3 will live outside of this repo. Regardless, the existing module is not fully implemented, and it's unclear if it remains useful in future Lotus integration.

Remove it from the codebase for now to simplify refactoring, code organisation and dependencies.


masih:
@Kubuxu I'd love your eyes on this PR please. My sense is that by the time we are done getting this repo ready for integration with Lotus the `module.go` as it is today would look very different. Considering its current state I'd like to remove it and revisit later.

masih:
To be discussed further when there are more concrete details on how F3 is going to be integrated in lotus. 

For now leave place holders be.

PR #159, opened 2024-04-15
Move simulation related packages to `sim`

masih:
Refactor the top-level packages to move all simulation related packages to `sim`.

The changes here make no functional alteration; they only move packages.


Stebalien (commented):
I agree with this change, but can't really approve it.

Kubuxu (approved):

anorth (commented):
👍

PR #161, opened 2024-04-15
Skip long tests when running in short mode

Kubuxu:


Stebalien (approved):

Stebalien:
Ideally we'd make them faster...

PR #168, opened 2024-04-22
Signal whether an EC chain was accepted by an instance

masih:
Update `ChainReceiver` interface to signal whether an EC chain was accepted by an instance or not. This allows tests to assert the current state of a participant and its current GPBFT instance specifically during equivocations testing.

Update `ChainReceiver` godoc to document the pre-existing and new behavior.

### DIFF
--- a/gpbft/api.go
+++ b/gpbft/api.go
@@ -2,10 +2,15 @@ package gpbft
 
 import "time"
 
-// Receives EC chain values.
+// ChainReceiver defines the interface for receiving updates to an Expected Consensus (EC) chain.
+// Implementers of this interface can act upon received EC chain data, potentially considering the
+// given chain as the expected consensus chain in the current GPBFT instance.
 type ChainReceiver interface {
-	// Receives a new EC chain, and notifies the current instance if it extends its current acceptable chain.
-	// This modifies the set of valid values for the current instance.
+	// ReceiveECChain processes an incoming EC chain update. It assesses whether the new chain
+	// extends or is compatible with the currently accepted chain maintained by the instance.
+	//
+	//Returns ErrECChainNotAcceptable error if the chain was not accepted. Otherwise, an error
+	// is returned if the chain is invalid.
 	ReceiveECChain(chain ECChain) error
 }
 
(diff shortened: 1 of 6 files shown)


gpbft/api.go
```diff
@@ -2,11 +2,16 @@ package gpbft
...
 type ChainReceiver interface {
-	// Receives a new EC chain, and notifies the current instance if it extends its current acceptable chain.
-	// This modifies the set of valid values for the current instance.
-	ReceiveECChain(chain ECChain) error
+	// ReceiveECChain processes an incoming EC chain update. It assesses whether the new chain
+	// extends or is compatible with the currently accepted chain maintained by the instance.
+	//
+	//Returns a boolean indicating whether the chain was accepted. An error is returned if the
+	// given chain is invalid.
+	ReceiveECChain(chain ECChain) (bool, error)
```
Kubuxu, line 14: We could use sentinel error instead. Is there a reason not to do that?

masih: Is there a reason not to use a boolean flag considering the pattern is already present in the gpbft API? 

Either works. My only desirable is consistency.

masih: In the interest of quicker turnaround I have updated the code to use sentinel error instead.

Please let me know if there is anything else that I might have missed.

Kubuxu (approved):

gpbft/api.go
```diff
@@ -2,10 +2,15 @@ package gpbft
...
+// ChainReceiver defines the interface for receiving updates to an Expected Consensus (EC) chain.
+// Implementers of this interface can act upon received EC chain data, potentially considering the
+// given chain as the expected consensus chain in the current GPBFT instance.
 type ChainReceiver interface {
-	// Receives a new EC chain, and notifies the current instance if it extends its current acceptable chain.
-	// This modifies the set of valid values for the current instance.
+	// ReceiveECChain processes an incoming EC chain update. It assesses whether the new chain
+	// extends or is compatible with the currently accepted chain maintained by the instance.
+	//
+	//Returns a boolean indicating whether the chain was accepted. An error is returned if the
```
Kubuxu, line 12: Although the doc now needs to be adjusted.

masih: My bad; thanks.

PR #179, opened 2024-04-26
Introduce additional simulation options and restructure constants.

masih:
Refactor the simulation initialization to accept a variable number of Option arguments. These options transform constants into adjustable parameters, such as the latency model, enabling a versatile combination of simulation scenarios.

Update the test suite to reflect these changes, eliminating redundant code. Convert dedicated tests into table tests, which now include basic tests covering BLS, non-BLS, agreement, and split convergence onto the base chain.

Refactor no-latency, also known as sync tests, to explicitly utilize a latency.None model, enhancing clarity.

Restructure adversary instantiation within a Generator. This change streamlines the adversary API, allowing it to leverage the existing gpbft API for initiating adversary logic, and obviates the need for specific Begin functions previously invoked sporadically across tests. Additionally, introducing generators enables the simulation API to present a fixed set of receivers and privatize most fields and receiver functions, minimizing potential errors when utilizing simulation APIs.

The current revisions include several TODO markers for future enhancements to maintain a manageable scope of changes in this pull request. The refactoring facilitates the future addition of probabilistic adversary populations and probabilistic ec chain mutations, with implementation details to follow in subsequent PRs.


sim/options.go
```diff
@@ -0,0 +1,184 @@
...
+		opts.honestCount = defaultHonestCount
+	}
+	if opts.latencyModel == nil {
+		opts.latencyModel = defaultLatencyModel
+	}
+	if opts.graniteConfig == nil {
+		opts.graniteConfig = &defaultGraniteConfig
+	}
+	if opts.signingBacked == nil {
+		opts.signingBacked = signing.NewFakeBackend()
```
Kubuxu, line 83: BLS envvar override is missing

Kubuxu (approved):
Apart from the one comment, SGWM.

sim/options.go
masih: Moved to f3sim command as it is the entrypoint to running a simulation. 

Does that make sense?

Kubuxu: When working on BLS, I used it to run all the tests with BLS as a sanity check.

Issue #170, opened 2024-04-22
Implement changes from 1st protocol audit


anorth:
See FIP PR https://github.com/filecoin-project/FIPs/pull/985

```[tasklist]
### Tasks
- [x] Replace ECCompatibleChains with `C` (give a better name) which is updated when node observes a value that could have been decided by another participant
- [x] Remove the ECUpdate (`ReceiveAcceptable`) method
- [x] Implement might-have-been-decided predicate (replaced with strong-quorum-of-bottom check)
- [x] Require evidence for PREPARE when round > 0
- [x] Carefully check the message validation and justification methods match spec
- [x] Carefully check for other changes to the protocol pseudocode such as timeouts, thresholds etc
- [x] Implement #136
```

PR #181, opened 2024-04-30
Replace externally-provided acceptable chains with candidate list including possibly-decided values

anorth:
Replace collection of acceptable chains received from outside with a collection of candidate chains that is updated to include any that might have been decided by another participant in a previous round.

- Prepare phase now waits to hear from >2/3 of power
- Evidence for CONVERGE is retained in order to justify swaying (and later to propagate to PREPARE)

The sequence of tests in COMMIT is slightly different than the spec, checking first for a strong quorum of COMMIT for bottom, and inferring that if that doesn't exist, any non-bottom COMMIT value could have been decided by another participant.

There is currently a shortcut in CONVERGE, where if the max ticket value cannot be accepted, the participant falls back to their own value instead of iterating through sequential tickets. I think this must be ok, since the node might not have received any others. But to be discussed whether this is good enough. Iterating the tickets will involve a larger change (I can follow up with it).

Part of #170.

### DIFF
--- a/gpbft/gpbft.go
+++ b/gpbft/gpbft.go
@@ -515,7 +515,7 @@ func (i *instance) tryQuality() error {
 	}
 
 	if foundQuorum || timeoutExpired {
-		// Add quorum prefixes to candidates (skipping base chain, which is already there).
+		// Add prefixes with quorum to candidates (skipping base chain, which is already there).
 		for l := range i.proposal {
 			if l > 0 {
 				i.candidates = append(i.candidates, i.proposal.Prefix(l))
@@ -584,9 +584,10 @@ func (i *instance) tryConverge() error {
 	if i.isCandidate(winner.Chain) {
 		i.proposal = winner.Chain
 		i.log("adopting proposal %s after converge", &winner.Chain)
-	}
-	// Else preserve own proposal
-	// XXX spec says to loop to next lowest ticket, rather than fall back to own proposal.
+	} // Else preserve own proposal
+	// NOTE: FIP-0086 says to loop to next lowest ticket, rather than fall back to own proposal.
+	// But using own proposal is valid (the spec can't assume any others have been received),
+	// considering others is an optimisation.
 
 	i.value = i.proposal
 	i.beginPrepare()
--- a/gpbft/gpbft.go
+++ b/gpbft/gpbft.go
@@ -284,7 +284,7 @@ func (i *instance) receiveOne(msg *GMessage) error {
 			i.quality.Receive(msg.Sender, prefix, msg.Signature)
 		}
 	case CONVERGE_PHASE:
-		if err := round.converged.Receive(msg.Sender, msg.Vote.Value, msg.Ticket); err != nil {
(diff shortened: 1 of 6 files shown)


gpbft/gpbft.go
```diff
@@ -579,27 +579,32 @@ func (i *instance) tryConverge() error {
...
+		i.log("⚠️ swaying from %s to %s by CONVERGE", &i.proposal, &winner.Chain)
+		i.candidates = append(i.candidates, winner.Chain)
 	}
-	i.beginPrepare()
+	if i.isCandidate(winner.Chain) {
+		i.proposal = winner.Chain
+		i.log("adopting proposal %s after converge", &winner.Chain)
+	}
+	// Else preserve own proposal
+	// XXX spec says to loop to next lowest ticket, rather than fall back to own proposal.
```
anorth, line 604: I'll delete this comment before merging, either because we decide it's ok or because I've added a task in #170 to do the iteration

test/constants_test.go
```diff
@@ -15,7 +15,7 @@ const (
 	tipSetGeneratorSeed = 0x264803e715714f95
 
 	latencyAsync         = 100 * time.Millisecond
-	maxRounds            = 10
+	maxRounds            = 20
```
anorth, line 18: Some multi-instance tests now require >15 rounds to complete in their worst case. The slow ones are with fewer participants, where it's more likely to get repeated rounds of no strong quorum observed in PREPARE, and so many commits for bottom.

test/honest_test.go
```diff
@@ -156,6 +156,35 @@ func TestAsyncAgreement(t *testing.T) {
 	}
 }
 
+func TestAsyncAgreementXXX(t *testing.T) {
```
anorth, line 159: I'll delete this before merging. It's a clone of the above, specialised to specific failing instances and with no parallelism so the logs are readable. FYI @masih

gpbft/participant.go
```diff
@@ -67,24 +67,6 @@ func (p *Participant) CurrentRound() uint64 {
...
-	defer func() {
-		if r := recover(); r != nil {
-			err = &PanicError{Err: r}
-		}
-	}()
-	if chain.IsZero() {
-		err = errors.New("cannot receive zero chain")
-	} else if err = chain.Validate(); err == nil && p.granite != nil {
-		if p.granite.ReceiveAcceptable(chain) {
-			err = ErrECChainNotAcceptable
```
masih, line 82: Please feel free to also remove `ErrECChainNotAcceptable`, which should be unused.

test/honest_test.go
masih: Alternatively, use `F3_TEST_REPETITION_PARALLELISM=1` to serialise repetitions?

ranchalp (approved):
LGTM

ranchalp:
The changes are justified and are correct. Later changes may cause us to reevaluate the implementation decisions here, but for the moment they make sense.

gpbft/gpbft.go
anorth: Thanks. Consider simplifying the FIP psuedocode too, and just noting the loop as a potential optimisation.

PR #188, opened 2024-05-03
Message templating, allowing multiple identities in one instance

Kubuxu:
Removes ParitcipantID from gpbft; the signing is outside of gpbft, leading to internal messages being delivered "from the outside".


masih:
@Kubuxu can you resolve conflicts before I review please?

Kubuxu:
@masih rebased

gpbft/messagetemplate.go
```diff
@@ -0,0 +1,97 @@
...
+
+type SigningTemplate struct {
+	ParticipantID ActorID
+	NetworkName   NetworkName
+	Payload       Payload
+	Justification *Justification
+
+	PubKey        PubKey
+	PayloadToSign []byte
+	VRFToSign     []byte
```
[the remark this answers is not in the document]
Kubuxu: They are exported for marshalling.

gpbft/api.go
```diff
@@ -32,9 +31,8 @@ type Chain interface {
 type Network interface {
 	// Returns the network's name (for signature separation)
 	NetworkName() NetworkName
-	// Sends a message to all other participants.
-	// The message's sender must be one that the network interface can sign on behalf of.
-	Broadcast(msg *GMessage)
+	// Requests that the message is signed and broadcasted, it should also be delivered locally
+	RequestBroadcast(mt *MessageTemplate)
```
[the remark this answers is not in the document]
Kubuxu: Broadcast failures are not exposed to gpbft, they will be logged in F3 layer

gpbft/messagetemplate_test.go
```diff
@@ -0,0 +1,101 @@
...
+			PubKey: PubKey{0},
+			Power:  big.NewInt(1),
+		},
+		{
+			ID:     1,
+			PubKey: PubKey{1},
+			Power:  big.NewInt(1),
+		},
+	}...)
+	assert.NoError(t, err)
```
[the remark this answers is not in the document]
Kubuxu: Yeah, I had require and assert swapped around.

sim/network.go
```diff
@@ -44,31 +49,77 @@ func newNetwork(opts *options) *Network {
...
+				deliverAt: nf.clock.Add(latencySample),
+			})
+	}
+}
+
 func (n *Network) NetworkName() gpbft.NetworkName {
 	return n.networkName
 }
 
 func (n *Network) Broadcast(msg *gpbft.GMessage) {
```
anorth, line 108: Please move this now next to Broadcast as part of the adversary interface only.

sim/network.go
```diff
@@ -44,31 +49,77 @@ func newNetwork(opts *options) *Network {
...
-			latencySample := n.latency.Sample(n.Time(), msg.Sender, dest)
-			n.queue.Insert(
-				messageInFlight{
-					source:    msg.Sender,
-					dest:      dest,
-					payload:   *msg,
-					deliverAt: n.clock.Add(latencySample),
-				})
+		if dest == msg.Sender {
+			continue
```
anorth, line 112: As far as I can tell this is functionally equivalent. If that's right, I think the previous approach without unstructured control flow is marginally better. Non-blocking.

sim/host.go
```diff
@@ -23,13 +23,20 @@ type simHost struct {
...
-		Network:  sim.network,
-		Verifier: sim.signingBacked,
-		Signer:   sim.signingBacked,
-		sim:      sim,
-		id:       id,
+		SimNetwork: sim.network.NetworkFor(sim.signingBacked, id),
+		Verifier:   sim.signingBacked,
+		Signer:     sim.signingBacked,
+		sim:        sim,
+		id:         id,
```
anorth, line 39: This `id` field might be redundant now, you can get it from SimNetwork.

sim/adversary/adversary.go
```diff
@@ -14,6 +15,8 @@ type Host interface {
 	// Note that the adversary can subsequently delay delivery to some participants,
 	// before messages are actually received.
 	BroadcastSynchronous(sender gpbft.ActorID, msg gpbft.GMessage)
+	// Sends a message to all other participants.
+	Broadcast(*gpbft.GMessage)
```
anorth, line 19: Nit: suggest listing the "normal" broadcast before the special one. These two could possibly be collapsed into one method with another parameter, too.

gpbft/participant_test.go
```diff
@@ -94,22 +94,19 @@ func (pt *participantTestSubject) expectBeginInstance() {
 	// Expect alarm is set to 2X of configured delta.
 	pt.host.EXPECT().SetAlarm(pt.time.Add(2 * pt.delta))
 
-	wantSignature := generateRandomBytes(pt.rng)
-
-	// Expect message is signed with expected payload.
-	pt.host.On("Sign", pt.pubKey, pt.matchMessageSigningPayload()).Return(wantSignature, nil)
-
 	// Expect a broadcast occurs with quality phase message, and the expected chain, signature.
```
anorth, line 97:
```suggestion
	// Expect a broadcast occurs with quality phase message, and the expected chain.
```

gpbft/messagetemplate.go
```diff
@@ -0,0 +1,97 @@
+package gpbft
+
+import (
+	"math/big"
+
+	xerrors "golang.org/x/xerrors"
+)
+
+type MessageTemplate struct {
+	powerTable powerTableAccess
```
anorth, line 10: This feels a bit clumsy to me, though I see how it was reached from the current code. The power table came from the host in the first place, and we're just passing it back here. This field could instead be just an instance ID (if the host doesn't already know this), and then the host could pass the right power table to `Build`.

This is a symptom of another awkwardness, that the signing key is looked up by public key rather than by participant ID. Previously I thought this was not worth changing, but it may be contributing here.

gpbft/messagetemplate.go
```diff
@@ -0,0 +1,97 @@
...
+
+	NetworkName NetworkName
```
anorth, line 12: Ditto the host should know this, no need to plumb it through GPBFT instance.

gpbft/messagetemplate.go
```diff
@@ -0,0 +1,97 @@
...
+	Payload     Payload
+
+	BeaconForTicket []byte
```
anorth, line 15: This came from the host too. Maybe we can avoid plumbing this through too. Should be able to look it up by instance. (The sim host probably doesn't have this mapping yet).

gpbft/gpbft.go
```diff
@@ -842,16 +835,11 @@ func (i *instance) buildJustification(quorum QuorumResult, round uint64, phase P
 func (i *instance) log(format string, args ...interface{}) {
 	if i.tracer != nil {
 		msg := fmt.Sprintf(format, args...)
-		i.tracer.Log("P%d{%d}: %s (round %d, step %s, proposal %s, value %s)", i.participant.id, i.instanceID, msg,
+		i.tracer.Log("{%d}: %s (round %d, step %s, proposal %s, value %s)", i.instanceID, msg,
```
anorth, line 838: Blocking: participant IDs in the logs are critical to development and debugging of the protocol itself. Please put prefix a participant ID inside the `Log` call.

Please enable tracing on one of the current unit tests, check out the logs, and make sure we can replicate something similar after this change.

anorth (changes requested):
The big items:
- [BLOCKING] Logs must have participant IDs
- [BLOCKING] Now-redundant local message inbox should be deleted
- There's lots of plumbing of values from outside through gpbft and then back to outside as part of the message template. The host already has these things, can we avoid plumbing them and just look up by instance ID?

The change to async delivery to self is a notable enough change that I would suggest doing it in an independent PR, before removing the participant ID. There's a bunch of other pre-factor work in the sim/network which could probably be pulled out ahead of the actual message templates, too.

gpbft/messagetemplate.go
Kubuxu: The point is that the MessageTemplate is exposed to whoever integrates it (Lotus), which can create the signing payload that can cross RPC boundaries to be signed and given back.
I will document it more, does that change your CR?

gpbft/gpbft.go
Kubuxu: It was done in the sim:
https://github.com/filecoin-project/go-f3/pull/188/files#diff-4fa4e1c1a7b22e16444c976a3fecfaf2d47f9c48ca7d1ebcb955034e9daab511R76-R79
I will check if it is sufficient.

sim/network.go
```diff
@@ -44,31 +49,77 @@ func newNetwork(opts *options) *Network {
...
+
+func (n *Network) NetworkFor(signer gpbft.Signer, id gpbft.ActorID) *NetworkFor {
+	return &NetworkFor{
+		ParticipantID: id,
+		Signer:        signer,
+		Network:       n,
+	}
+}
+
+type NetworkFor struct {
```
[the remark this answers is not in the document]
Kubuxu: Is there any reason not to? It is a wrapper onto Network which is also exported.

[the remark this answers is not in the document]
masih: My vote and the common convention is to not export unless needed. This is not a blocker. Please don't let me slow you down. We can refactor it later.

sim/sim.go
```diff
@@ -13,11 +13,20 @@ type Simulation struct {
...
 	network      *Network
 	ec           *EC
-	participants []*gpbft.Participant
+	participants []Participant
 	adversary    *adversary.Adversary
 	decisions    *DecisionLog
 }
 
+type Participant struct {
+	ParticipantID gpbft.ActorID
```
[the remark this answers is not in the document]
Kubuxu: It would conflict with method name

gpbft/messagetemplate.go
anorth: I don't have a good enough view into how this will be used in the integration, so I won't hold you up on it. I do feel like we can do better, but don't know enough to say exactly how.

PR #190, opened 2024-05-03
Investigate sensitivity to latency in multi instance tests

### LINKED ISSUE #189: `TestMultiAsyncAgreement` is latency sensitive
`TestMultiAsyncAgreement` fails if the seed is incremented by 1.

See: https://github.com/filecoin-project/go-f3/commit/b342125d0a25d59fa4028480b3c98b79aa485f5c

masih:
Assert that when latency seed is changed the tests passes eventually. In this case, it looks like the change in latency results in few additional rounds for the test to pass; 21 to be precise.

Increase the max rounds to 30 in order to avoid intermittent errors during dev work that might change latency seeds.

Closes #189

### DIFF
--- a/test/constants_test.go
+++ b/test/constants_test.go
@@ -15,7 +15,7 @@ const (
 	tipSetGeneratorSeed = 0x264803e715714f95
 
 	latencyAsync         = 100 * time.Millisecond
-	maxRounds            = 18
+	maxRounds            = 30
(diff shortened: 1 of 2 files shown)


Kubuxu (approved):

Kubuxu:
I'm wondering if such behaviour also exists in other places but now it won't be a mystery (at least for me).

PR #191, opened 2024-05-03
Set default Delta value to 3 seconds

masih:
Update the default delta value to 3 seconds and document the rationale behind it.

The defaults may change in the future once concrete measurements on a live network are performed.

### DIFF
--- a/gpbft/options.go
+++ b/gpbft/options.go
@@ -6,7 +6,7 @@ import (
 )
 
 var (
-	defaultDelta                = 2 * time.Second
+	defaultDelta                = 3 * time.Second
 	defaultDeltaBackOffExponent = 1.3
 )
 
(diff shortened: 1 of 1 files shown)


masih:
> except for the comment

Comment adjusted; thank you

gpbft/options.go
```diff
@@ -35,7 +35,13 @@ func newOptions(o ...Option) (*options, error) {
 }
 
 // WithDelta sets the expected bound on message propagation latency. Defaults to
-// 2 seconds if unspecified. Delta must be larger than zero.
+// 3 seconds if unspecified. Delta must be larger than zero.
+//
+// The default of 3 seconds for the value of Dela is based on the current EC
+// propagation time (i.e. round-trip time of 6 seconds) as the initial
+// propagation timeout. The actual propagation time may be significantly lower.
```
jsoares, line 42: EC doesn't assume a 6s _round-trip time_, as that's not even well-defined for a pubsub network. What it assumes is an upper bound on the gossipsub network-wide propagation time, which is actually 10s now, following https://github.com/filecoin-project/lotus/pull/9290. 

The increase from 6s to 10s was due to having time to spare in the epoch envelope due to faster proof computation; it does not imply the need to go higher than 6s. The value remains user-configurable, 6s worked well, and we have reason to believe that we can go lower.

I think we can even skip talking about EC and just say that we have reason to believe, based on previous data, that 3s is enough.

jsoares (approved):
See comment on comment, looks good otherwise.

gpbft/options.go
masih: Noted thank you; merge queue got the PR merged. I have take note to update the comments.

jsoares: Oops, forgot we had it enabled in the repo 🤦

masih: It's my fault. I added this pr to queue before your excellent review.

PR #194, opened 2024-05-07
Implement skipping directly to DECIDE phase on receiving a valid (justified) DECIDE message.

### LINKED ISSUE #136: Implement skipping to DECIDE phase after receiving single DECIDE message
A DECIDE message carries justification of a strong quorum of COMMITs, and so is enough to convince a node to decide the same value immediately, and send its own DECIDE message. Implement this shortcut.

This is a specific case of a more general ability to use the justification in place of receiving the messages that it is evidence of (e.g. a COMMIT with a strong quorum of PREPAREs).

anorth:
Closes #136

See also #170

### DIFF
--- a/gpbft/gpbft.go
+++ b/gpbft/gpbft.go
@@ -183,12 +183,12 @@ func newInstance(
 		phase:       INITIAL_PHASE,
 		proposal:    input,
 		value:       ECChain{},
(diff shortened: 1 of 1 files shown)


ranchalp (approved):
LGTM!

Kubuxu (approved):

PR #198, opened 2024-05-07
Limit max chain length to 100 tipsets

### LINKED ISSUE #121: Maximum ECChain Length
If the network goes down for a while, the chain we all need to agree on could get pretty long. If that happens, we could (in theory) start sending messages that approach or exceed the maximum pubsub message size of 1MiB.

We should set some reasonable upper limit on the length of the ECChain. This will increase the time it takes to catch up to the main chain, but the limit can easily be hundreds. We should just have _some_ limit.

anorth:
Closes #121

### DIFF
--- a/gpbft/chain.go
+++ b/gpbft/chain.go
@@ -27,6 +27,9 @@ type ECChain []TipSet
 // A map key for a chain. The zero value means "bottom".
 type ChainKey string
 
+// Maximum length of a chain value.
+const CHAIN_MAX_LEN = 100
+
 // Creates a new chain.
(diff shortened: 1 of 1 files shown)


Stebalien (approved):
LGTM. Although, given that we're committing to these values via a merkle-tree, I'd consider specifying this as 128.

Stebalien:
> LGTM. Although, given that we're committing to these values via a merkle-tree, I'd consider specifying this as 128.

Eh, that's a spec change. I'll add that to my spec change.

PR #200, opened 2024-05-07
Ignore equivocating and duplicate messages from a sender

### LINKED ISSUE #153: Drop equivocations
Prior versions of the protocol spec required a node to retain all equivocating messages (same instance, round, step and peer but different value). The code reflects this. The explicit justification removes that need, so the code should instead ignore equivocations. Change it to ignore all but the first message from a peer for each round and step (this will also help one angle of #12).

With luck, this can be done just by changing the quorum state data structures to be only able to represent one value per participant, and checking if such value has been received before updating.

anorth:
Closes #153

### DIFF
--- a/gpbft/gpbft.go
+++ b/gpbft/gpbft.go
@@ -279,10 +279,7 @@ func (i *instance) receiveOne(msg *GMessage) error {
 	switch msg.Vote.Step {
 	case QUALITY_PHASE:
 		// Receive each prefix of the proposal independently.
-		for j := range msg.Vote.Value.Suffix() {
-			prefix := msg.Vote.Value.Prefix(j + 1)
-			i.quality.Receive(msg.Sender, prefix, msg.Signature)
-		}
+		i.quality.ReceiveEachPrefix(msg.Sender, msg.Vote.Value)
 	case CONVERGE_PHASE:
 		if err := round.converged.Receive(msg.Sender, msg.Vote.Value, msg.Ticket, msg.Justification); err != nil {
 			return fmt.Errorf("failed processing CONVERGE message: %w", err)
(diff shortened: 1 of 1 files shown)


gpbft/gpbft.go
```diff
@@ -890,39 +891,63 @@ func newQuorumState(powerTable PowerTable) *quorumState {
...
+	}
+	q.receiveInner(sender, value, senderPower, signature)
+}
+
+// Receives each prefix of a chain as a distinct value from a sender.
+// Note that this method does not store signatures, so it is not possible later to
+// create an aggregate for these prefixes.
+// This is intended for use in the QUALITY phase.
+// Ignores any subsequent values from a sender from which a value has already been received.
+func (q *quorumState) ReceiveEachPrefix(sender ActorID, values ECChain) {
```
Stebalien, line 909: Did you mean to use this? Or is this for a future patch?

gpbft/gpbft.go
```diff
@@ -1090,22 +1122,29 @@ type ConvergeTicket struct {
...
 
-	// !!! TODO: Drop duplicates (see https://github.com/filecoin-project/go-f3/issues/12)
-	c.values[key] = ConvergeValue{Chain: value, Justification: justification}
-	c.tickets[key] = append(c.tickets[key], ConvergeTicket{Sender: sender, Ticket: ticket})
-
+	// Keep only the first justification and ticket received for a value.
+	if _, found := c.values[key]; !found {
+		c.values[key] = ConvergeValue{Chain: value, Justification: justification}
+		c.tickets[key] = append(c.tickets[key], ConvergeTicket{Sender: sender, Ticket: ticket})
+	}
```
Stebalien, line 1147: I think I missed this part of the conversation. The code looks correct, but I don't have a strong enough understanding of the protocol to give it an OK from that perspective.

Stebalien (approved):
LGTM with a caveat.

ranchalp (approved):
LGTM!

gpbft/gpbft.go
ranchalp: This is ok!

gpbft/gpbft.go
anorth: Thanks, good catch. A call to this replaces the equivalent code in the QUALITY_PHASE case in `receieveOne`.

PR #201, opened 2024-05-07
Panic when receiving message signed by self

### LINKED ISSUE #57: Audit/test code for safety when multiple instances run with the same identity
In real life, operational errors may cause two nodes to run with the same identity (same secret key). This could invalidate implicit assumptions in the code and control flow, e.g. that a message from yourself can't be received before it is sent.

Desk-check the code for assumptions like this, and exercise it in simulation with multiple participants sharing an identity.

anorth:
Closes #57

A test for this added as an item in #9

### DIFF
--- a/gpbft/gpbft.go
+++ b/gpbft/gpbft.go
@@ -401,6 +401,14 @@ func (i *instance) validateMessage(msg *GMessage) error {
 		return xerrors.Errorf("invalid signature on %v, %v", msg, err)
 	}
 
+	// Checks that message is not from self.
+	// This panics rather than returning a simple error, since it represents
+	// a serious misconfiguration and may violate assumptions made by this algorithm.
+	// This check must come after the signature is verified, to avoid a simple denial attack.
+	if msg.Sender == i.participant.id {
+		panic("received message from self")
+	}
+
 	// Check justification
 	needsJustification := !(msg.Vote.Step == QUALITY_PHASE ||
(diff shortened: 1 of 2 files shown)


Stebalien:
I thought we counted our own votes by receiving our own messages. Is this not the case?

Anyways, anyone on the network can send us our own message back. If we're going to panic here, we should reject messages from ourself in `ReceiveMessage`.

Stebalien:
Also, IIRC, @Kubuxu was working to remove the participant ID from the instance (as signing will be handled by the client).

Kubuxu:
I agree with Stebalien. We still should have some way to detect this case, but I'm moving towards removing the ParticipantID from the instance in #188. 
The current code in there passes our own messages the same way as any other, but we can differentiate them.

Given the replay tests, where the adversary replays messages, I also wonder why this PR works right now.

anorth:
> I thought we counted our own votes by receiving our own messages. Is this not the case?

This happens internally to the GPBFT instance, below this entry point. 

Given that, yes, just rejecting the message here would be enough. Or if the participant ID is removed from this level, then this check needs to be performed higher up where the participant ID is known.

I'm fairly confident that dropping of equivocations #153 will mean things are probably ok here now. But this check for misconfiguration would still be valuable. I'll close this PR, but leave the issue open and assign to @Kubuxu to do somewhere closer to the network.

Issue #184, opened 2024-05-01
Use options with reasonable defaults to configure gpbft


masih:
The current implementation of the gPBF algorithm takes a `GraniteConfig` struct as input, which contains delta and backoff exponent parameters. Additionally, the `Participant` struct constructor accepts several other configuration parameters.

To improve maintainability, consistency, and usability, I propose unifying the configuration mechanism to use a consistent terminology, a standardised way of setting config for a gPBFT instance, and sensible defaults when config parameters are unset. Specifically:

1. **Consistent Terminology:** Remove the `Granite` prefix from the config struct, as it seems to be used only once in the codebase.
2. **Config Mechanism:** Introduce an `options` pattern to set config for a gPBFT instance. This will provide a convenient and self-descriptive way to configure the algorithm.
3. **Sensible Defaults:** When config parameters are unset, provide sensible default values to simplify configuration and reduce errors.
4. **Strict Configuration Validation:** Implement strict validation checks on the configured options to ensure they meet specific requirements.

**Acceptance Criteria:**

1. The `GraniteConfig` struct is removed, and a new `options` pattern is introduced.
2. The `Participant` class constructor no longer accepts multiple configuration parameters.
3. A mechanism is provided to set config for a gPBFT instance using the `options` pattern.
4. Sensible default values are implemented when config parameters are unset.
5. Strict validation checks are performed on configured options.

PR #202, opened 2024-05-08
Implement storage power evolution in simulations for honest participants

### LINKED ISSUE #114: Consider supporting power table evolution in simulator
As of #111, the sim uses a fixed power table for all instances. 

There might not be much value in implementing evolution of the power table since the participant code simply takes what it is given each time by the host. Logic of constructing the next power table from EC state is outside this repo.

masih:
Implement the ability to evolve the storage power across honest participants over the course of a simulation. The changes introduce primitives to allow doing this and reflect the changes across the tests using a uniformed fixed storage power generator.

The adversary storage power evolution is not yet implemented and left to future work in favour of reducing the number of lines changed.

Add two tests that exercise increase and decrease of storage power over sync and async message passing:

1. Sudden increase of storage power results in the corresponding group of nodes to dominate consensus as expected.

2. Gradual decrease of storage power results in loss of dominance among the corresponding group.

Fixes #114

**Rebased on top of #187; review that first.**


anorth:
Could you change this PR base to #184 so it's easier to review the diff proposed here?

sim/host.go
```diff
@@ -18,24 +18,29 @@ type simHost struct {
 	gpbft.Verifier
 	gpbft.Clock
 
-	id  gpbft.ActorID
-	sim *Simulation
+	id     gpbft.ActorID
+	sim    *Simulation
+	pubkey gpbft.PubKey
```
anorth, line 23: @Kubuxu re my comments on #188 , putting the pubkey here seems reasonable and may avoid the power table plumbing.

sim/sim.go
```diff
@@ -152,12 +139,15 @@ func (s *Simulation) initParticipants() error {
 	// Limit adversaryCount to exactly one for now to reduce LOC up for review.
 	// Future PRs will expand this to support a group of adversaries.
```
anorth, line 140: Nit: I think you can just remove this, and the TODO above and `adversaryCount`. Just note there's at most one adversary (though it can have arbitrary power).

sim/sim.go
```diff
@@ -152,12 +139,15 @@ func (s *Simulation) initParticipants() error {
...
 	if s.adversaryGenerator != nil && s.adversaryCount == 1 {
+		host := newHost(nextID, s, NewFixedECChainGenerator(*s.baseChain), nil)
 		// Adversary implementations currently ignore the canonical chain.
 		// Set to a fixed ec chain generator and expand later for possibility
 		// of implementing adversaries that adapt based on ec chain.
 		// TODO: expand adversary archetypes.
-		host := newHost(nextID, s, NewFixedECChainGenerator(*s.baseChain))
 		s.adversary = s.adversaryGenerator(nextID, host)
+		// TODO: implement evolving adversary power; postponed to reduce PR size.
```
anorth, line 148: Replace TODO with a note that adversary power does not evolve.

test/power_evolution_test.go
```diff
@@ -0,0 +1,168 @@
...
+					require.NotEqual(t, chainBackedByMostPower.Suffix(), chainBackedByLeastPower.Suffix())
+
+					// Assert the consensus is reached on the chain with most power.
+					requireConsensusAtInstance(t, sm, i, chainBackedByMostPower...)
+					base = instance.BaseChain.Head()
+				}
+			})
+		})
+	}
+}
```
anorth, line 98: 👍

anorth (commented):
Basically looks good, but will wait for prereq to land

sim/host.go
Kubuxu: The power table plumbing is external to the sim, as F3 won't know for which participants it will be signing. Only the application itself knows that.

Kubuxu (approved):

PR #206, opened 2024-05-09
Avoid proposing chains that are longer than max.

### LINKED ISSUE #203: Avoid proposing `Value`s with more than `CHAIN_MAX_LEN`
Right now, I believe the _client_ needs to avoid telling go-f3 about more than 99 (`CHAIN_MAX_LEN-1`) tipsets. Otherwise, we'll attempt to propose _more_ than 100 tipsets in a single instance and fail the validity check.

We need to implement a rate-limit and, IMO, the right place to put this is in go-f3 itself. I.e., let the client tell go-f3 about as many tipsets as it wants but have go-f3 take chunks of at most `CHAIN_MAX_LEN` tipsets at a time.

anorth:
Closes #203

### DIFF
--- a/gpbft/participant.go
+++ b/gpbft/participant.go
@@ -145,9 +145,11 @@ func (p *Participant) ReceiveAlarm() (err error) {
 
 func (p *Participant) beginInstance() error {
 	chain, power, beacon := p.host.GetCanonicalChain()
+	// Limit length of the chain to be proposed.
 	if chain.IsZero() {
 		return errors.New("canonical chain cannot be zero-valued")
 	}
+	chain = chain.Prefix(CHAIN_MAX_LEN - 1)
 	if err := chain.Validate(); err != nil {
 		return fmt.Errorf("invalid canonical chain: %w", err)
 	}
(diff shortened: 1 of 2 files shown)


gpbft/participant.go
```diff
@@ -145,6 +145,8 @@ func (p *Participant) ReceiveAlarm() (err error) {
 
 func (p *Participant) beginInstance() error {
 	chain, power, beacon := p.host.GetCanonicalChain()
+	// Limit length of the chain to be proposed.
+	chain = chain.Prefix(CHAIN_MAX_LEN - 1)
```
Stebalien, line 149: The chain includes the base, right? This should be `CHAIN_MAX_LEN` in that case (no `-1`).

Stebalien: (it would be _nice_ to assert that the base matches the end of our previous proposal...

anorth: The `Prefix` method adds one: `Prefix(0)` is the base chain. Yes perhaps this isn't the ideal semantics, but it makes more sense at some other call sites. I wouldn't object to you overhauling that in another PR, but it's below my threshold at the moment.

anorth: Or did you mean that the max length should be CHAIN_MAX_LEN tips plus one for the base?

Stebalien: Ah, no. I misunderstood prefix. LGTM.

Stebalien (approved):

Kubuxu (approved):

PR #207, opened 2024-05-09
Drop messages from prior rounds that can't affect the protocol outcome.

### LINKED ISSUE #205: Drop messages from past rounds (except COMMIT)
Messages from prior rounds can't affect the protocol execution, except for late-arriving COMMITs which can cause a decision. I don't think it causes any problem if we process them, but dropping is an easy and worthwhile optimisation.

(They should still be rebroadcast to other participants, so don't mark them as invalid).

anorth:
Closes #205

### DIFF
--- a/gpbft/gpbft.go
+++ b/gpbft/gpbft.go
@@ -238,7 +238,7 @@ func (i *instance) Receive(msg *GMessage) error {
 }
 
 func (i *instance) ReceiveAlarm() error {
-	if err := i.tryCompletePhase(); err != nil {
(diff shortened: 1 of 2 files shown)


Stebalien (approved):
(modulo test fixes)

This seems like a good way to avoid storing cruft. We should also consider _dropping_ state from previous rounds.

Stebalien:
> We should also consider dropping state from previous rounds.

I see we've already discussed this and your point about it not being worth it is fair.

ranchalp (approved):

Kubuxu (approved):

Issue #196, opened 2024-05-07
Investigate the cause for the max round increase during testing


masih:
The expectation is that the tests should converge in smaller number of rounds than the current max. It was increased [from 10 to 18](https://github.com/filecoin-project/go-f3/pull/181/files#diff-b3137011cf7dcbb378c64ffff6f238da9434b459d5287e789a8e41971a5a99faR18), and again from [18 to 30 (though only needing 21) ](https://github.com/filecoin-project/go-f3/pull/190).

Ideally, it should take fewer rounds for the consensus to occur. Currently it is unclear if this is a side effect of the gPBFT algorithm changes or some unintended behaviour in simulations.

PR #209, opened 2024-05-10
Simplify simulation messaging with priority queue and parameterise GST

masih:
Use `heap.Interface` to simplify the implementation of message passing during simulation by implementing a priority queue. The messages in-flight are priorities in ascending order of their `deliverAt` time, i.e. messages with earlier delivery time have higher priority.

Additionally, explicitly introduce an option to set Global Stabilization Time (GST). GST duration is one of the fundamental assumptions made by gPBFT, beyond which message delivery is assumed to be guaranteed. Enhance Withold adversary tests with varying GST time.

Relates to #196

### DIFF
--- a/sim/message_queue.go
+++ b/sim/message_queue.go
@@ -1,49 +1,111 @@
 package sim
 
 import (
-	"sort"
+	"container/heap"
 	"time"
 
 	"github.com/filecoin-project/go-f3/gpbft"
 )
 
+var _ heap.Interface = (*messageQueue)(nil)
+
+// messageQueue is a priority queue that implements heap.Interface and holds
+// simulation messages in flight prioritised by their delivery time in ascending
+// order.
+type messageQueue struct {
+	mailbox []*messageInFlight
+}
+
 type messageInFlight struct {
 	source    gpbft.ActorID // ID of the sender
 	dest      gpbft.ActorID // ID of the receiver
-	payload   interface{}   // Message body
+	payload   any           // Message body
 	deliverAt time.Time     // Timestamp at which to deliver the message
+
+	index int // Index in the heap used internally by the heap implementation
 }
 
-// A queue of directed messages, maintained as an ordered list.
-type messageQueue []messageInFlight
-
-func (h *messageQueue) Insert(x messageInFlight) {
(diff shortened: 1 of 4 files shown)


Kubuxu (approved):

sim/message_queue.go
```diff
@@ -1,49 +1,111 @@
...
+}
+
+// Pop removes and returns the highest priority message from the queue.
+//
+// This function is part of heap.Interface and must not be called externally.
+// See: Remove.
+func (pq *messageQueue) Pop() any {
+	old := pq.mailbox
+	n := len(old)
+	item := old[n-1]
```
anorth, line 76: This looks like it's removing the lowest priority (highest deliverAt) item, if they are in ascending order. Which I don't think is what we want here.

This message queue is lacking tests (my fault).

sim/network.go
```diff
@@ -108,49 +120,49 @@ func (n *Network) BroadcastSynchronous(sender gpbft.ActorID, msg gpbft.GMessage)
...
+		}
+		n.log(TraceRecvd, "P%d %s", msg.source, payload)
 		if err := receiver.ReceiveAlarm(); err != nil {
-			return false, fmt.Errorf("failed receiving alarm: %w", err)
+			return false, fmt.Errorf("failed to deliver alarm from %d to %d: %w", msg.source, msg.dest, err)
+		}
+	case gpbft.GMessage:
+		// If GST has not elapsed, check if adversary allows the propagation of message.
+		if adv != nil && !n.globalStabilisationElapsed {
+			if n.hasGlobalStabilizationTimeElapsed() {
```
anorth, line 145: This is a bit confusing to read. We should just remove the boolean field if this can be calculated instead.

sim/message_queue.go
masih: The implementation of `Pop` is correct, i.e. adheres to what `heap.Interface` demands. It's my documentation that's misleading; apologies.

Unit tests are added [here](https://github.com/filecoin-project/go-f3/pull/221) along with documentation correction.

sim/network.go
anorth: Ok makes sense, thanks.

PR #215, opened 2024-05-14
Truncate ECChain capacity when taking a prefix

Stebalien:
We don't need to be thread-safe, but the `chain.Prefix(n).Extend(tip)` would previously have overwritten `chain`. I don't think this is an issue (we only use `Extend` in tests?) but we might as well fix it.


Stebalien:
Hm. Actually, this is insufficient. multiple calls to `Extend` on the same chain can also cause issues.

PR #216, opened 2024-05-14
EVM-friendly TipSet and ECChain formats

### LINKED ISSUE #150: Add power table commitment (CID) and accessory commitment to Tipset structure
Add a power table commitment and a placeholder for an accessory commitment into the Tipset structure. Both are opaque to GPBFT but must be included in considerations of equality and signing payloads (wait for #149 first).

See FIP PR https://github.com/filecoin-project/FIPs/pull/983

### LINKED ISSUE #166: ECTipset Format Redux
@masih, @Kubuxu, and I were discussing the eventual upgrade path to cross-chain bridging and ran into an issue. Even with a 100 tipset limit:

1. We can still easily reach 26KiB+ of data in a single finality certificate.
2. This data is CBOR, so it's hard to make it a private input to a SNARK (requires parsing CBOR in a snark). I think we're going to be _very_ unhappy if we have to start proving facts about CBOR-encoded data inside a snark.
3. Even without snarks, parsing CBOR is non-trivial and not well supported on most chains.

We initially claimed that this was fine because we don't yet support cross-chain bridging but... we still want an upgrade path towards cross-chain bridging (hence the commitment tree). I hate to re-open this can of worms, but I think we need to commit to a sequence of commitment hashes instead of CBOR values. That is:

```go
type Value = [32]byte // root hash of a merkle tree
type ECChain = []Value // concatenated hashes, we could also make a tree here, but I don't think that's necessary.
```

That way, any future SNARKs can:

1. Given a snark friendly commitment to the power-table from the last instance.
2. Validate the signature over the root commitment (32 bytes).
3. Validate a simple merkle-proof down to facts we're trying to prove.
4. Use those facts (e.g., in a new SNARK).

The one objection we envision is observability. But we can handle _that_ by treating this commitment scheme as a part of the signature scheme (suggested by @Kubuxu). Basically:

1. The GossiPBFT protocols still transmit full tipset information.
2. The signature also includes the merkle-proof proving that the signed commitment commits to the transmitted tipste.
3. The v1 finality certificate format can also include the tipsets/power table directly. It's just that the signature will _only_ cover the root hash.

Importantly, this would let us create small finality certificates that _don't_ require parsing CBOR to understand.

Stebalien:
This patch implements the changes described in
https://github.com/filecoin-project/FIPs/pull/1004.

Unfortunately, go-f3 now needs to be aware of the internal structure of `TipSet` objects as it sends them over the wire in one format, but signs another. This effectively reverts #149.

Specific changes:

1. On the wire (and in finality certificates), `TipSet` objects are now directly cbor-marshaled within the payload instead of first being marshaled to byte strings.
2. Instead of directly covering the entire tipset list with the payload signature:
    1. `TipSet` objects are "marshaled for signing" per the FIP change proposed above.
    2. These marshaled tipset objects are stuffed into a merkle tree.
    3. The merkle tree root is signed (along with the payload's other fields), again according the proposed changes.

fixes #166
fixes #150


gpbft/chain.go
```diff
@@ -5,16 +5,76 @@ import (
...
+// TipSetKey is a
+type TipSetKey = []byte
+
+type CID = []byte
+
+var cidPrefix = []byte{0x01, 0x71, 0xA0, 0xE4, 0x02, 0x20}
+
+// Hashes the given data and returns a CBOR + blake2b-256 CID.
+func MakeCid(data []byte) []byte {
+	// TODO: Consider just using go-cid? We implicitly depend on it through cbor-gen anyways.
```
Stebalien, line 24: Thoughts? The only reason to keep it _this_ way is performance, tbh.

gpbft/chain.go
```diff
@@ -45,8 +105,8 @@ func (c ECChain) IsZero() bool {
 }
 
 // Returns the base tipset.
-func (c ECChain) Base() TipSet {
-	return c[0]
+func (c ECChain) Base() *TipSet {
```
Stebalien, line 108: I'm returning by pointer here and elsewhere as this structure isn't small.

gpbft/chain_test.go
```diff
@@ -7,56 +7,17 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestTipSet(t *testing.T) {
```
Stebalien, line 10: This test really wasn't doing much.

Stebalien: I'm happy to add something back to test `IsZero`, `Equals` and the stringer if desired.

gpbft/gpbft.go
```diff
@@ -90,26 +91,30 @@ type Payload struct {
...
-	_ = binary.Write(&buf, binary.BigEndian, p.Round)
+
 	_ = binary.Write(&buf, binary.BigEndian, p.Step)
-	for _, t := range p.Value {
-		_ = binary.Write(&buf, binary.BigEndian, uint32(len(t)))
-		buf.Write(t)
-	}
+	_ = binary.Write(&buf, binary.BigEndian, p.Round)
+	_ = binary.Write(&buf, binary.BigEndian, p.Instance)
+	_, _ = buf.Write(root[:])
```
Stebalien, line 117: TODO: write a test asserting the correct layout and expected offsets.

gpbft/chain.go
```diff
@@ -5,16 +5,76 @@ import (
...
+	_ = cbg.WriteByteArray(&buf, ts.TipSet)
+	tsCid := MakeCid(buf.Bytes())
+	buf.Reset()
+	buf.Grow(len(tsCid) + len(ts.PowerTable) + 32 + 8)
+	// epoch || commitments || tipset || powertable
+	_ = binary.Write(&buf, binary.BigEndian, ts.Epoch)
+	_, _ = buf.Write(ts.Commitments[:])
+	_, _ = buf.Write(tsCid)
+	_, _ = buf.Write(ts.PowerTable)
+	return buf.Bytes()
```
Stebalien, line 68: TODO: write test asserting the expected layout and offsets.

gpbft/chain.go
Stebalien: I'm also happy to bikeshed on this (e.g., take by pointer in more places to reduce the dereferencing, etc.).

Stebalien:
A few TODOs (needs a few more tests) but should otherwise be ready for review.

Stebalien:
Hm. Apparently this is killing performance (probably all the hashing). We may need to make that pluggable (e.g., for testing).

gpbft/chain_test.go
masih: Please feel free to remove the test. 

(not now) At some point we should review the exported `gpbft` package APIs and avoid having multiple ways of doing the same thing. Which I suspect might be the root cause.

Stebalien:
If I look at the total time for `TestRepeat`:

- before - 6.3s
- keccak merkle-tree - 41s
- sha2 merkle-tree - 19s
- sha2 merkle-tree + no tipset hashing - 13.6s

gpbft/chain.go
```diff
@@ -5,16 +5,76 @@ import (
...
 	"encoding/hex"
 	"errors"
+	"fmt"
 	"strings"
+
+	cbg "github.com/whyrusleeping/cbor-gen"
+	"golang.org/x/crypto/blake2b"
 )
 
+// TipSetKey is a
```
anorth, line 15: Incomplete comment

gpbft/chain.go
```diff
@@ -5,16 +5,76 @@ import (
...
+type TipSetKey = []byte
+
+type CID = []byte
+
+var cidPrefix = []byte{0x01, 0x71, 0xA0, 0xE4, 0x02, 0x20}
```
anorth, line 20: Can you add a comment describing this derivation?

gpbft/chain.go
anorth: Yes the dep made me sad. I still prefer to avoid using if it's not too hard.

gpbft/chain.go
```diff
@@ -5,16 +5,76 @@ import (
...
+	// TODO: Consider just using go-cid? We implicitly depend on it through cbor-gen anyways.
+	digest := blake2b.Sum256(data)
+
+	out := make([]byte, 0, 38)
+	out = append(out, cidPrefix...)
+	out = append(out, digest[:]...)
+	return out
+}
+
 // Opaque type representing a tipset.
```
anorth, line 33: Not opaque any more! Move the lines of this comment next to the fields, now they are present.

gpbft/chain.go
```diff
@@ -5,16 +5,76 @@ import (
...
 // This is expected to be:
 // - a canonical sequence of CIDs of block headers identifying a tipset,
 // - a commitment to the resulting power table,
 // - a commitment to additional derived values.
 // However, GossipPBFT doesn't need to know anything about that structure.
-type TipSet = []byte
+type TipSet struct {
+	Epoch       int64
+	TipSet      TipSetKey
```
anorth, line 41:
```suggestion
	Key?      TipSetKey
```

This will probably often show up as `tipset.TipSet`

gpbft/chain.go
```diff
@@ -5,16 +5,76 @@ import (
...
+func (ts *TipSet) Equal(b *TipSet) bool {
+	return ts.Epoch == b.Epoch &&
+		bytes.Equal(ts.TipSet, b.TipSet) &&
+		bytes.Equal(ts.PowerTable, b.PowerTable) &&
+		ts.Commitments == b.Commitments
+}
+
+func (ts *TipSet) MarshalForSigning() []byte {
+	var buf bytes.Buffer
+	_ = cbg.WriteByteArray(&buf, ts.TipSet)
```
anorth, line 59: Why is this using `cbg`? Just `buf.Write`?

gpbft/chain.go
anorth: Yeah this interface was mostly pointers before I stripped out the type, if I recall. Should probably have more pointer receivers and parameters in that case.

anorth (commented):
Partial review, not done yet

gpbft/chain.go
Stebalien: Can do.

gpbft/chain.go
Stebalien: Because https://github.com/filecoin-project/FIPs/blob/b98ae5f0e9c0b033279f29fa5c507e457f826947/FIPS/fip-0086.md?plain=1#L307

I'm actually not sure where the wires got crossed in the FEVM FIP. IMO, it should have either been `CBOR([Cid, Cid, Cid])` or `Raw(Cid || Cid || Cid)`. Instead we got `CBOR(Cid || Cid || Cid)`.

Stebalien: Honestly, I'm tempted to change it to `CBOR([Cid, Cid, Cid])` because then we'd be able to dereference the CIDs. E.g.,`$TIPSET_CID/0` would refer to the CID of the first block. But if we did that, we'd probably (eventually) want a small FIP to make FEVM use the new format.

sim/signing/bls.go
```diff
@@ -46,3 +46,7 @@ func (b *BLSBackend) GenerateKey() (gpbft.PubKey, any) {
 	b.signersByPubKey[string(pubKeyB)] = blssig.SignerWithKeyOnG1(pubKeyB, priv)
 	return pubKeyB, priv
 }
+
+func (b *BLSBackend) MarshalPayloadForSigning(nn gpbft.NetworkName, p *gpbft.Payload) []byte {
+	return p.MarshalForSigning(nn)
```
Stebalien, line 51: I'm _actually_ marshaling here because, in this case, I assume we don't really care about performance.

sim/signing/fake.go
```diff
@@ -84,3 +85,37 @@ func (s *FakeBackend) VerifyAggregate(payload, aggSig []byte, signers []gpbft.Pu
 	}
 	return nil
 }
+
+func (v *FakeBackend) MarshalPayloadForSigning(nn gpbft.NetworkName, p *gpbft.Payload) []byte {
```
Stebalien, line 89: This skips _all_ hashing. We'll eventually hash with sha256 when signing, but only doing that once is much faster.

gpbft/api.go
```diff
@@ -92,4 +92,6 @@ type Host interface {
 	Signer
 	Verifier
 	DecisionReceiver
+
+	MarshalPayloadForSigning(*Payload) []byte
```
Stebalien, line 96: I added this as a new function on the base `Host` because it doesn't really fit on either the `Signer` or the `Verifier`. I can also create a new `SignatureOps` (or something) interface that combines `Signer`, `Verifier`, and `MarshalPayloadForSigning`.

gpbft/gpbft.go
```diff
@@ -400,7 +405,7 @@ func (i *instance) validateMessage(msg *GMessage) error {
 	}
 
 	// Check vote signature.
-	sigPayload := msg.Vote.MarshalForSigning(i.participant.host.NetworkName())
+	sigPayload := i.participant.host.MarshalPayloadForSigning(&msg.Vote)
```
Stebalien, line 408: This turned out to be "not terrible" because we already needed the host for the network name.

gpbft/participant_test.go
```diff
@@ -84,12 +84,13 @@ func (pt *participantTestSubject) expectBeginInstance() {
 	// Prepare the test host.
 	pt.host.On("GetCanonicalChain").Return(pt.canonicalChain, *pt.powerTable, pt.beacon)
 	pt.host.On("Time").Return(pt.time)
-	pt.host.On("NetworkName").Return(pt.networkName)
+	pt.host.On("NetworkName").Return(pt.networkName).Maybe()
+	pt.host.On("MarshalPayloadForSigning", mock.AnythingOfType("*gpbft.Payload")).
+		Return([]byte(gpbft.DOMAIN_SEPARATION_TAG + ":" + pt.networkName))
```
Stebalien, line 90: Is this the right way to do this? Basically:

1. We may _not_ call `NetworkName` now, because that's handled inside of `Host.MarshalPayloadForSigning`.
2. Whenever we call `Sign`, we call `Host.MarshalPayloadForSigning`.

gpbft/api.go
Stebalien: I merged it into a `Signatures` interface. Open to suggestions on better names.

masih: Bikeshedding on name; alternative suggestions: 
* `Signing`, 
* `SignerVerifier`
* (don't hate me) separate all the sign related stuff into their own package called `signing` and call the interface `Scheme` then in `gpbft` it would read `signing.Scheme

Stebalien: I'll leave 3 for a future patch. Between 1, 2, and the current name... I don't have any preference.

merkle/merkle_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+			}
+			root, paths := TreeWithProofs(test)
+			root2 := Tree(test)
+			require.Equal(t, root, root2)
+			require.Equal(t, len(test), len(paths))
+			assert.Equal(t, len(paths[0]), depth(test))
+
+			for i, path := range paths {
+				valid, more := VerifyProof(root, i, test[i], path)
```
anorth, line 26: Add some test for VerifyProof where it should return false.

anorth (approved):
Looks solid. I suggest waiting for one more peer review on this one, though.

merkle/merkle_test.go
Stebalien: Done.

masih (approved):

Kubuxu (commented):
Late but LGTM

PR #217, opened 2024-05-14
Truncate ECChain capacity when extending/truncating

Stebalien:
Truncate the capacity when truncating and extending the chain. That way:

1. Prefixing then extending won't mutate the original chain.
2. Extending the same base chain with one tipset then another won't clobber the first by re-using the capacity from the original slice.

This isn't great for performance, but:

1. `Extend` is only used for testing.
2. Truncating the capacity in `Prefix` has no performance cost unless we decide to append to the chain later.

### DIFF
--- a/gpbft/chain.go
+++ b/gpbft/chain.go
@@ -72,14 +72,14 @@ func (c ECChain) BaseChain() ECChain {
 }
 
 func (c ECChain) Extend(tip ...TipSet) ECChain {
-	return append(c, tip...)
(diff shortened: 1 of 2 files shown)


Stebalien:
I'm mostly concerned that we'll end up with tests that mysteriously break and/or work when they shouldn't.

Kubuxu (approved):

anorth:
Thanks!

PR #221, opened 2024-05-15
Add unit tests for simulation message queue

masih:
Assert that simulation priority queue used to order message delivery is behaving as expected:

* queued messages are sorted in ascending order of their `deliverAt`.
* removal of a message returns the one with earliest `deliverAt`.
* conditional upsert is respected.
* duplicate messages are allowed and ordered as expected.
* queue is fully drained upon removing all messages.

Update misleading godoc for `Pop` receiver.

Addresses https://github.com/filecoin-project/go-f3/pull/209#discussion_r1600731429

### DIFF
--- a/sim/message_queue_test.go
+++ b/sim/message_queue_test.go
@@ -1,0 +1,158 @@
+package sim
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestMessageQueue_IsInAscendingOrderOfDeliverAt(t *testing.T) {
+	subject := newMessagePriorityQueue()
+
+	want1st := &messageInFlight{deliverAt: time.Time{}.Add(time.Second)}
+	want2nd := &messageInFlight{deliverAt: want1st.deliverAt.Add(12 * time.Second)}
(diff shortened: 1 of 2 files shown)


sim/message_queue_test.go
```diff
@@ -0,0 +1,158 @@
...
+		{
+			name:            "matching more than one updates first",
+			givenInsertions: []*messageInFlight{m1, m2, m4},
+			givenMatcher:    func(msg *messageInFlight) bool { return msg.payload == m2.payload || msg.payload == m4.payload },
+			givenUpsert:     m3,
+			wantMessages:    []*messageInFlight{m1, m3, m4},
+		},
+	}
+	for _, test := range tests {
+		test := test
```
Kubuxu, line 123: Do we require Go 1.22? If so this is no longer required.

Kubuxu: Ok, no looks like we are on Go 1.21

Kubuxu (approved):

sim/message_queue_test.go
masih: yep; can't wait.

anorth (commented):
Great, thank you!

PR #222, opened 2024-05-15
Report code coverage via Continuous Integration

masih:
Change `make test` target to measure code coverage in atomic mode, and extend CI test workflow to upload the reports to Codecov.


masih:
Once this PR is merged to `main` we should start seeing [the reports](https://app.codecov.io/github/filecoin-project/go-f3).

Kubuxu:
Let's ensure that codecov build status is marked as not required until we tune the settings (it can be fiddly).

Kubuxu:
I would also move coverage collection to separate make target as collecting coverage has non-insignificant perf hit.

Kubuxu (approved):

PR #230, opened 2024-05-15
Collect cross-package coverage

Kubuxu:


masih (approved):

Kubuxu:
Usually, I would have separate unit test and integration test coverage, but we can start with this (focusing on integration testing coverage).

PR #254, opened 2024-05-20
Consensus over pubsub

Kubuxu:
To test it out:
```
make f3
./f3 manifest gen
```
then in two terminals, in quick succession:
```
[T1] ./f3 run --id=0
[T2] ./f3 run --id=1
```

Many things are still incomplete, the input chains are completely faked
as an example.

I would prefer to keep merging smaller improvements instead of having to
make a huge PR at the end.

Part of #94


sim/signing/fake.go
```diff
@@ -36,10 +32,8 @@ func (s *FakeBackend) Sign(signer gpbft.PubKey, msg []byte) ([]byte, error) {
 }
 
 func (s *FakeBackend) generateSignature(signer gpbft.PubKey, msg []byte) ([]byte, error) {
-	priv, known := s.keyPairs[string(signer)]
-	if !known {
-		return nil, errors.New("unknown signer")
```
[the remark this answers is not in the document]
Kubuxu: I can parse and compare with `i`, but that would require testing to produce a public key from nowhere. Are we concerned about that?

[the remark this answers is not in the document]
masih: There is a signing interface in sim/singning package that defines the GenerateKey reciever. 

I think the question worth thinking about is: what should happen here in production code if a given public key is unknown?

Whatever the answer to that question is should be reproducible by the fake signer for testing purposes. Otherwise, fake signer would miss a whole class of behaviour in tests.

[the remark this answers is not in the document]
Kubuxu: Then, we should separate the fake sigher by the simulated participant.

cmd/f3/run.go
```diff
@@ -0,0 +1,96 @@
...
+		if err != nil {
+			return xerrors.Errorf("creating gossipsub: %w", err)
+		}
+
+		err = setupDiscovery(h)
+		if err != nil {
+			return xerrors.Errorf("setting up discovery: %w", err)
+		}
+
+		tmpdir, err := os.MkdirTemp("", "f3-*")
```
masih, line 50: Where does `tmpdir` gets removed?

cmd/f3/run.go
```diff
@@ -0,0 +1,96 @@
...
+	fmt.Printf("discovered new peer %s\n", pi.ID)
+	err := n.h.Connect(context.Background(), pi)
+	if err != nil {
+		fmt.Printf("error connecting to peer %s: %s\n", pi.ID, err)
+	}
+}
+
+func setupDiscovery(h host.Host) error {
+	// setup mDNS discovery to find local peers
+	s := mdns.NewMdnsService(h, DiscoveryTag, &discoveryNotifee{h: h})
```
masih, line 94: Setup interrupt signal to close this off gracefully upon `SIGINT`?

gpbft/participant.go
```diff
@@ -160,6 +160,7 @@ func (p *Participant) beginInstance() error {
 	if p.granite, err = newInstance(p, p.nextInstance, chain, power, beacon); err != nil {
 		return fmt.Errorf("failed creating new granite instance: %w", err)
 	}
+	p.granite.tracer = p.tracer
```
masih, line 163: Move to `newInstance` as constructor arg since there is a constructor for `granite`. That would limit scope for error in case `newInstrance` manipulates `tracer` as a field that's owned by `instance` struct.

host.go
```diff
@@ -0,0 +1,188 @@
...
+		Verifier:         v,
+		MessageQueue:     make(chan *gpbft.GMessage, 20),
+		SelfMessageQueue: make(chan *gpbft.GMessage, 20),
+		broadcast:        broadcast,
+
+		log: log,
+	}
+
+	// create a stopped timer
+	h.timer = time.NewTimer(100 * time.Hour)
```
masih, line 43: what does this timer do?

host.go
```diff
@@ -0,0 +1,188 @@
...
+// These are:
+// - the EC chain to propose,
+// - the power table specifying the participants,
+// - the beacon value for generating tickets.
+// These will be used as input to a subsequent instance of the protocol.
+// The chain should be a suffix of the last chain notified to the host via
+// ReceiveDecision (or known to be final via some other channel).
+func (h *gpbfthost) GetCanonicalChain() (chain gpbft.ECChain, power gpbft.PowerTable, beacon []byte) {
+	// TODO: this is just a complete fake
+	ts := sim.NewTipSetGenerator(1)
```
masih, line 109: I understand that this is part of a larger body of work. But I am struggling to see why we are using simulation tipset generator here and am unsure where this is going. 

If the purpose of this is to showcase a network over gossipsub then would it make sense to instead: 
* build an API that runs gPBFT over libp2p.
* write a test that uses that library to run a network with gopssipsub transport.

I am struggling to see what the F3 binary is going to be once the larger body of work this is part of is complete.

host.go
```diff
@@ -0,0 +1,188 @@
...
+	return h.manifest.NetworkName
+}
+
+// Sends a message to all other participants.
+// The message's sender must be one that the network interface can sign on behalf of.
+func (h *gpbfthost) Broadcast(msg *gpbft.GMessage) {
+	h.log.Info("broadcast")
+
+	bw := &bytes.Buffer{}
+	err := msg.MarshalCBOR(bw)
```
masih, line 136:
```suggestion
	var bw bytes.Buffer
	err := msg.MarshalCBOR(&bw)
```

manifest.go
```diff
@@ -0,0 +1,8 @@
+package f3
+
+import "github.com/filecoin-project/go-f3/gpbft"
+
+type Manifest struct {
+	NetworkName       gpbft.NetworkName
+	InitialPowerTable []gpbft.PowerEntry
```
masih, line 7: I recommend adding JSON struct tags to hardcode field names if the future production API will include this.

module.go
```diff
@@ -91,14 +93,34 @@ func (m *Module) Run(ctx context.Context) error {
...
+
+	go func() {
+		err := h.Run(ctx)
+		m.log.Errorf("running host: %+v", err)
+	}()
 	for {
 		msg, err := sub.Next(ctx)
 		if err != nil {
 			m.log.Errorf("pubsub subscription.Next() returned an error: %+v", err)
 			break
```
masih, line 111: It seems slightly odd to swallow the original error here, break out of the loop and potentially return some other error that may be unrelated to the original error that triggered the whole thing.

Makefile
```diff
@@ -30,3 +30,7 @@ lint:
 generate:
 	go generate ./...
 .PHONY: generate
+
+f3:
```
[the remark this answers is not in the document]
Kubuxu: Usually, in makefiles, targets that produce artefacts are named as just the file. 
I can add `make build` target that builds it as well (and other binaries if we have them).

cmd/f3/run.go
Kubuxu: It doesn't help examination when it is used; in the future, it will be a long-lived directory.

host.go
Kubuxu: I need something in the place of the timer while it is not initialized.

module.go
Kubuxu: This error won't get captured either way, I can wrap it but it will end up just logged

manifest.go
Kubuxu: This isn't even close to the final shape. Once we know what we want there, we can hardcode field names.

host.go
Kubuxu: I'm using a simulation tipset generator here because this is the easiest thing to stub this out while we work out the rest of it.

1. We want a standalone binary at some point to facilitate observation, testing, development against a live network, and so on.
2. Library-level tests will be there once things are more complete. Currently, I expect the code to change so much that they would be a distraction.
3.  We need tooling to set up the bootstrap and test networks, which the f3 tool will facilitate.

Kubuxu: I also don't want to spend time rebasing a multiple thousand lines PR until things are finished. That is why I'm using stubs.

cmd/f3/manifest.go
```diff
@@ -0,0 +1,79 @@
...
+		&cli.IntFlag{
+			Name:  "N",
+			Usage: "number of participant",
+			Value: 2,
+		},
+	},
+	Action: func(c *cli.Context) error {
+		path := c.String("manifest")
+		var m f3.Manifest
+		m.NetworkName = "localnet"
```
Stebalien, line 34: I'd randomize this.

cmd/f3/run.go
```diff
@@ -0,0 +1,100 @@
...
+		return module.Run(ctx)
+	},
+}
+
+type discoveryNotifee struct {
+	h host.Host
+}
+
+func (n *discoveryNotifee) HandlePeerFound(pi peer.AddrInfo) {
+	fmt.Printf("discovered new peer %s\n", pi.ID)
```
Stebalien, line 85: debug log

cmd/f3/run.go
```diff
@@ -0,0 +1,100 @@
...
+	if err != nil {
+		fmt.Printf("error connecting to peer %s: %s\n", pi.ID, err)
+	}
+}
+
+func setupDiscovery(ctx context.Context, h host.Host) error {
+	// setup mDNS discovery to find local peers
+	s := mdns.NewMdnsService(h, DiscoveryTag, &discoveryNotifee{h: h})
+	go func() {
+		<-ctx.Done()
```
Stebalien, line 96: :unamused:

cmd/f3/run.go
```diff
@@ -0,0 +1,100 @@
...
+
+type discoveryNotifee struct {
+	h host.Host
+}
+
+func (n *discoveryNotifee) HandlePeerFound(pi peer.AddrInfo) {
+	fmt.Printf("discovered new peer %s\n", pi.ID)
+	err := n.h.Connect(context.Background(), pi)
+	if err != nil {
+		fmt.Printf("error connecting to peer %s: %s\n", pi.ID, err)
```
Stebalien, line 88: log

cmd/f3/run.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+}
+
+func setupDiscovery(ctx context.Context, h host.Host) error {
```
Stebalien, line 92: configurable via manifest?

host.go
```diff
@@ -0,0 +1,188 @@
...
+	MessageQueue     chan *gpbft.GMessage
+	SelfMessageQueue chan *gpbft.GMessage //for the future when self messages are async
+
+	log Logger
+
+	timer      *time.Timer
+	runningCtx context.Context
+}
+
+func newHost(id gpbft.ActorID, m Manifest, broadcast func(context.Context, []byte) error,
```
Stebalien, line 29: I'd make `broadcast` an interface so we don't have to change it later. Really, I'd define a single interface like:

```go
type Client interface {
	Signer
	Verifier
	Logger // maybe not?

	BroadcastMessage(context.Context []byte) error
	// receive message? what am I missing?
}
```

host.go
```diff
@@ -0,0 +1,188 @@
...
+
+type gpbfthost struct {
+	participant *gpbft.Participant
+	manifest    Manifest
+	gpbft.Signer
+	gpbft.Verifier
+	broadcast func(context.Context, []byte) error
+
+	MessageQueue     chan *gpbft.GMessage
+	SelfMessageQueue chan *gpbft.GMessage //for the future when self messages are async
```
Stebalien, line 21: Please avoid public channels.

host.go
```diff
@@ -0,0 +1,188 @@
...
+	"github.com/filecoin-project/go-f3/gpbft"
+	"github.com/filecoin-project/go-f3/sim"
+	"golang.org/x/xerrors"
+)
+
+type gpbfthost struct {
+	participant *gpbft.Participant
+	manifest    Manifest
+	gpbft.Signer
+	gpbft.Verifier
```
Stebalien, line 17: put these at the top?

host.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+func newHost(id gpbft.ActorID, m Manifest, broadcast func(context.Context, []byte) error,
+	s gpbft.Signer, v gpbft.Verifier, log Logger) (*gpbfthost, error) {
+	h := &gpbfthost{
+		manifest:         m,
+		Signer:           s,
+		Verifier:         v,
+		MessageQueue:     make(chan *gpbft.GMessage, 20),
+		SelfMessageQueue: make(chan *gpbft.GMessage, 20),
```
Stebalien, line 36: Why the queue size?

host.go
Stebalien: call it `alertTimer` and add a comment?

host.go
```diff
@@ -0,0 +1,188 @@
...
+		select {
+		case <-h.timer.C:
+			h.log.Infof("alarm fired")
+			err = h.participant.ReceiveAlarm()
+		case msg := <-h.SelfMessageQueue:
+			_, err = h.participant.ReceiveMessage(msg, true)
+		case msg := <-h.MessageQueue:
+			_, err = h.participant.ReceiveMessage(msg, false)
+		case <-ctx.Done():
+			err = ctx.Err()
```
Stebalien, line 85: return immediately and don't return an error.

host.go
```diff
@@ -0,0 +1,188 @@
...
+	if err != nil {
+		return nil, xerrors.Errorf("creating participant: %w", err)
+	}
+	h.participant = p
+	return h, nil
+}
+
+func (h *gpbfthost) Run(ctx context.Context) error {
+	h.runningCtx = ctx
+	time.Sleep(2 * time.Second)
```
Stebalien, line 59: comment

host.go
```diff
@@ -0,0 +1,188 @@
...
+	p, err := gpbft.NewParticipant(id, h, gpbft.WithTracer(tracer{log}))
+	if err != nil {
+		return nil, xerrors.Errorf("creating participant: %w", err)
+	}
+	h.participant = p
+	return h, nil
+}
+
+func (h *gpbfthost) Run(ctx context.Context) error {
+	h.runningCtx = ctx
```
Stebalien, line 58: wrap in a cancel, defer a cancel of the context in case this function fails for some _other_ reason

cmd/f3/run.go
Kubuxu: It is more of a temporary hack but if we decide to keep it in the future. Maybe.

cmd/f3/run.go
Kubuxu: I made the same face at MdsnService for not taking in context and instead of creating it internally

module.go
```diff
@@ -91,21 +93,47 @@ func (m *Module) Run(ctx context.Context) error {
...
+		var gmsg gpbft.GMessage
+		err = gmsg.UnmarshalCBOR(bytes.NewReader(msg.Data))
+		if err != nil {
+			m.log.Info("bad pubsub message: %+v", err)
+			continue
+		}
+		select {
+		case h.MessageQueue <- &gmsg:
+		case <-ctx.Done():
+			break
```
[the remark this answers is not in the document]
Stebalien: (found by linter)

cmd/f3/run.go
Stebalien: Services _should not_ take contexts for shutdown. It creates a context internally for canceling its own requests.

host.go
```diff
@@ -0,0 +1,188 @@
...
+// At most one alarm can be set at a time.
+// Setting an alarm replaces any previous alarm that has not yet fired.
+// The timestamp may be in the past, in which case the alarm will fire as soon as possible
+// (but not synchronously).
+func (h *gpbfthost) SetAlarm(at time.Time) {
+	h.log.Infof("set alarm for %v", at)
+	// we cannot reuse the timer because we don't know if it was read or not
+	h.timer.Stop()
+	h.timer = time.NewTimer(time.Until(at))
+}
```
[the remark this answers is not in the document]
Kubuxu: It is the 2nd case; it is only called by the Participant. We still really can't reuse the timer because we don't know if we are in participant because the timer fired and was read, or the timer fired while the participant was running.

host.go
Kubuxu: IDK seemed reasonable for the case when messages are coming in and we are running participants right now; I can make it a const.

cmd/f3/run.go
Kubuxu: It creates context internally to control its own lifetime.

host.go
Stebalien: Just make sure there's no deadlock.

host.go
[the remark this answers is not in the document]
Stebalien: I see.... I'd still consider sending a message on a control channel isntead.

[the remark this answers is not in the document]
Kubuxu: The Participant is inherently single-threaded, and we don't currently have a control channel. I can add them if you think it warrants the complexity.

host.go
Kubuxu: There is no loopback from host to module/pubsub apart from Broadcast which doesn't go through the same loop.

Stebalien (commented):
Code nits. I still need to dig into the design

host.go
```diff
@@ -0,0 +1,195 @@
...
+	gpbft.Verifier
+	gpbft.Tracer
+
+	BroadcastMessage(context.Context, []byte) error
+	IncommingMessages() <-chan *gpbft.GMessage
+	Logger() Logger
+}
+
+type gpbfthost struct {
+	client
```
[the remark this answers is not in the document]
Stebalien: (or, well, make it clear what we want to expose and what we don't)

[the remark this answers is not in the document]
Stebalien: Ok, these all look like private fields so I'd hide them.

host.go
```diff
@@ -0,0 +1,195 @@
...
+	h.log.Info("broadcast")
+
+	var bw bytes.Buffer
+	err := msg.MarshalCBOR(&bw)
+	if err != nil {
+		h.log.Errorf("marshalling GMessage: %+v", err)
+	}
+	err = h.client.BroadcastMessage(h.runningCtx, bw.Bytes())
+	if err != nil {
+		h.log.Errorf("broadcasting GMessage: %+v", err)
```
Stebalien, line 157: I'd return the error to GPBFT and let it decide how to handle this.

host.go
```diff
@@ -0,0 +1,195 @@
...
+
+// Sends a message to all other participants.
+// The message's sender must be one that the network interface can sign on behalf of.
+func (h *gpbfthost) Broadcast(msg *gpbft.GMessage) {
+	h.log.Info("broadcast")
+
+	var bw bytes.Buffer
+	err := msg.MarshalCBOR(&bw)
+	if err != nil {
+		h.log.Errorf("marshalling GMessage: %+v", err)
```
Stebalien, line 153: return the error, definitely don't continue.

host.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+
+	return chain, nil
+}
+
+func (h *gpbfthost) GetCommitteeForInstance(instance uint64) (*gpbft.PowerTable, []byte, error) {
+	table := gpbft.NewPowerTable()
+	err := table.Add(h.manifest.InitialPowerTable...)
+	if err != nil {
+		h.log.Errorf("creating powertable: %+v", err)
```
Stebalien, line 134: We're doing a lot of logging. Shouldn't this be up to GPBFT?

module.go
```diff
@@ -17,49 +19,84 @@ import (
...
+	logger         Logger
+	loggerWithSkip Logger
+
+	// Populated after Run is called
+	messageQueue <-chan *gpbft.GMessage
+	topic        *pubsub.Topic
+}
+
+func (mc moduleClient) BroadcastMessage(ctx context.Context, msg []byte) error {
+	return mc.topic.Publish(ctx, msg)
```
Stebalien, line 47: note: we may need to make this async.

module.go
```diff
@@ -86,65 +123,80 @@ func (m *Module) Run(ctx context.Context) error {
...
+	m.client.messageQueue = messageQueue
+
+	h, err := newHost(m.id, m.Manifest, m.client)
+	if err != nil {
+		return xerrors.Errorf("creating gpbft host: %w", err)
+	}
+
+	go func() {
+		err := h.Run(ctx)
+		m.log.Errorf("running host: %+v", err)
```
Stebalien, line 141: We should:

1. Cancel `ctx` from the outer module on exit (e.g., on error).
2. Send the error on some channel if there's a failure (or close the channel to indicate that we've exited.

module.go
```diff
@@ -86,65 +123,80 @@ func (m *Module) Run(ctx context.Context) error {
...
 
 // validator for the pubsub
 func (m *Module) pubsubTopicValidator(ctx context.Context, pID peer.ID, msg *pubsub.Message) pubsub.ValidationResult {
+	var gmsg gpbft.GMessage
+	err := gmsg.UnmarshalCBOR(bytes.NewReader(msg.Data))
+	if err != nil {
+		return pubsub.ValidationReject
+	}
+
+	// TODO more validation
```
Stebalien, line 186: call `m.host.Validate`? Any reason not to do that now?

Stebalien (approved):
I'd fix the error handling then merge (as long as we have clear TODOs in the code).

module.go
Kubuxu: Yeah, for sure, but it will change significantly with message builder.

Stebalien: Sg. Then I'd just leave a TODO and do it later.

cmd/f3/manifest.go
```diff
@@ -0,0 +1,83 @@
...
+	},
+	Action: func(c *cli.Context) error {
+		path := c.String("manifest")
+		rng := make([]byte, 4)
+		_, _ = rand.Read(rng)
+		var m f3.Manifest
+		m.NetworkName = gpbft.NetworkName(fmt.Sprintf("localnet-%X", rng))
+		fsig := signing.NewFakeBackend()
+		for i := 0; i < c.Int("N"); i++ {
+			pubkey, _ := fsig.GenerateKey()
```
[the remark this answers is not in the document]
Kubuxu: this is not an error (it is a private key)

host.go
```diff
@@ -0,0 +1,200 @@
...
+		case msg, ok := <-messageQueue:
+			if !ok {
+				err = xerrors.Errorf("incoming messsage queue closed")
+				break loop
+			}
+			_, err = h.participant.ReceiveMessage(msg, false)
+		case <-ctx.Done():
+			return nil
+		}
+		if err != nil {
```
anorth, line 111: You'll need much better error handling here. Right now this will just bail upon receiving an invalid message. Wait for #264 to land, but more work needed.

Issue #151, opened 2024-04-03
Use delayed power table to validate or drop messages from future instances


anorth:
We have agreed that voting weights for an instance should come from the power table resulting from not the immediately previous instance, but some distance (10 instances / 5 minutes? longer) further back. This allows us to do proper message validation and avoid retransmitting bad messages.

Since the power table is given from the node, it might be that theres little or no code change required in GPBFT, but (1) confirm this, and (2) ensure that multi-instance tests do this properly (#295), and (3) let's use #125 to help us gain confidence in the lookback distance.

```[tasklist]
### Tasks
- [x] #256
- [ ] #264
- [ ] #286
- [ ] #294
```

anorth:
The point of this is to be able to validate message from near-future instances (from a node's point of view). So there will be work here to implement that validation and queue of _validated_ messages. This may involve changes to the host API so the participant can reliably keep track of N power tables and have the right one on hand for any message.

anorth:
I'm closing #130 and expanding the scope here a little with a task list including maintaining the delayed power tables and  using them to validate messages.

Stebalien:
IMO, we should use `int64` for instances _regardless_ (I've seen too many issues with MaxUint64 overflowing int64).

That aside, I don't think having the lookback inside go-f3 will actually be all that difficult:

1. We can still initialize instances with the desired power table (no lookback).
2. A `GetPowerTableFromInstance` method can simply assert that the passed instance has the correct lookback for the instance being simulated, then return the power table.

Kubuxu:
We are storing finality certificates, which should contain the power tables that are being finalized.

Stebalien:
Well, the finality certificates only store power table diffs. But looking up the power table associated with an instance isn't difficult (`instance - lookback_distance` -> head ts -> power table).

Stebalien:
See https://github.com/filecoin-project/go-f3/pull/273/files#r1610685278 for an example of the power tables I'll need if we implement #257.

anorth:
Testing effectiveness of this broken out to #295

PR #256, opened 2024-05-21
Change host API to separate fetching chain from fetching power tables

anorth:
Step 1 of #151. This is in preparation for fetching power tables for validation of messages for future instances.

See discussion in #151 about the interface design. The powertable offset parameter is encapsulated by the host, which can respond to a request for the power table to use for some (possibly-future) instance. This capability isn't used yet.

This change also removes fixed beacons in the sim, which was previously using the same beacon for every instance. Now it's generated (by setting it equal to the base tipset key bytes).

### DIFF
--- a/gpbft/api.go
+++ b/gpbft/api.go
@@ -23,15 +23,17 @@ type Receiver interface {
 }
 
 type Chain interface {
-	// Returns inputs to the next GPBFT instance.
-	// These are:
-	// - the EC chain to propose,
-	// - the power table specifying the participants,
-	// - the beacon value for generating tickets.
-	// These will be used as input to a subsequent instance of the protocol.
-	// The chain should be a suffix of the last chain notified to the host via
-	// ReceiveDecision (or known to be final via some other channel).
-	GetCanonicalChain() (chain ECChain, power PowerTable, beacon []byte)
+	// Returns the chain to propose for a new GPBFT instance.
+	// This should be a suffix of the chain finalised by the immediately prior instance.
+	// Returns an error if the chain for the instance is not available.
+	GetChainForInstance(instance uint64) (chain ECChain, err error)
+
+	// Returns the power table and beacon value to be used for a GPBFT instance.
+	// These values should be derived from a chain previously received as final by the host,
+	// or known to be final via some other channel (e.g. when bootstrapping the protocol).
+	// The offset (how many instances to look back) is determined by the host.
+	// Returns an error if the committee for the instance is not available.
+	GetCommitteeForInstance(instance uint64) (power *PowerTable, beacon []byte, err error)
 }
 
 // Endpoint to which participants can send messages.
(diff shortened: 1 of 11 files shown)


gpbft/api.go
```diff
@@ -23,15 +23,17 @@ type Receiver interface {
...
+	// This should be a suffix of the chain finalised by the immediately prior instance.
+	// Returns an error if the chain for the instance is not available.
+	GetChainForInstance(instance uint64) (chain ECChain, err error)
+
+	// Returns the power table and beacon value to be used for a GPBFT instance.
+	// These values should be derived from a chain previously received as final by the host,
+	// or known to be final via some other channel (e.g. when bootstrapping the protocol).
+	// The offset (how many instances to look back) is determined by the host.
+	// Returns an error if the power table for the instance is not available.
+	GetCommitteeForInstance(instance uint64) (power PowerTable, beacon []byte, err error)
```
masih, line 36: Thinking out loud: The phrase `Committee` seems reasonable. But, unless it is already used elsewhere, I am curious if we can avoid introducing a new terminology.  Having said that, I am failing to think of an existing alternative.

masih (approved):

gpbft/api.go
Kubuxu, line 36: Pass the power table by pointer, or change the constructor to return value instead of pointer.

gpbft/api.go
```diff
@@ -23,15 +23,17 @@ type Receiver interface {
...
-	// These are:
-	// - the EC chain to propose,
-	// - the power table specifying the participants,
-	// - the beacon value for generating tickets.
-	// These will be used as input to a subsequent instance of the protocol.
-	// The chain should be a suffix of the last chain notified to the host via
-	// ReceiveDecision (or known to be final via some other channel).
-	GetCanonicalChain() (chain ECChain, power PowerTable, beacon []byte)
+	// Returns the chain to propose for a new GPBFT instance.
+	// This should be a suffix of the chain finalised by the immediately prior instance.
```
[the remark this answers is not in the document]
anorth: The comment is written from the narrow point of view of what does this code need. And in fact it will function correctly whether or not this is true (e.g. at genesis, or in case of a social agreement to re-start somewhere else, etc). This code doesn't _need_ it to be a suffix, we can't write any test that will fail if it isn't, etc. Whether or not the multi-party protocol will make progress if this doesn't hold is a higher level matter.

I'd rather delete the comment than change it to a must.

gpbft/api.go
anorth: Yes it is new in the code (though common in the literature). If we generalised to sub-sampling the power table in the future (for scale), this would be exactly the right term.

I went through a similar thought process as you but couldn't come up with anything better.

anorth: Done, and propagated the pointer to a few other places.

PR #258, opened 2024-05-21
Drop spammable messages beyond max allowed lookahead rounds

### LINKED ISSUE #240: GPBFT instance should drop spammable messages from future rounds
Most messages carry a justification which demonstrates that 2/3 of power is up to the same round (or previous) as the message itself. An attacker can't generate these for rounds arbitrarily far in the future, unless the network has actually progressed to that round, so these are not a DOS vector (assuming #153).

A COMMIT for bottom carries no such evidence. Pick some finite number of rounds ahead at which to accept these messages, and drop anything further ahead. The upcoming rebroadcast mechanism will be responsible for getting the network up to the same round if a network partition lasts longer than that.

Check also that there are no other such spammable messages. Perhaps just use `msg.justification= nil` as the test.

masih:
The gPBFT rebroadcast design revision introduced an improvement whereby "spammable" messages, i.e. `COMMIT` for bottom, for future rounds should be dropped by the participants. The specification leaves room for implementers to optionally retain some number of such messages to reduce reliance on the need for rebroadcast and ultimately help reach consensus in fewer rounds.

The work here:
1. introduces `WithMaxLookaheadRounds` configuration option to `gpbft` package, defaulting to zero if unset.
2. drops all spammable messages belonging to rounds that are beyond the configured max lookahead rounds.

A `Spam` adversary is introduced to replicate the conditions at which max lookahead rounds would start dropping messages. A set of tests then use the `Spam` adversary to assert that honest participants reach expected consensus despite the presence of spam messages.

Note that there are no APIs to assert that the upper lookahead bound is respected by the implementation. A TODO is left to revisit this once telemetry is introduced whereby tests can observe future message queue size metrics to assert a configured lookahead is indeed respected.

Resolves #240

### DIFF
--- a/gpbft/gpbft.go
+++ b/gpbft/gpbft.go
@@ -288,6 +288,28 @@ func (i *instance) receiveOne(msg *GMessage) error {
 		return nil
 	}
 
+	// Drop message that:
+	//  * belong to future rounds, beyond the configured max lookahead threshold, and
+	//  * carry no justification, i.e. are spammable.
+	//
+	// The only messages that are spammable are COMMIT for bottom. QUALITY and
+	// PREPARE messages may also not carry justification, but they are not
+	// spammable. Because:
+	//  * QUALITY is only valid for round zero.
+	//  * PREPARE must carry justification for non-zero rounds.
+	//
+	// Therefore, we are only left with COMMIT for bottom messages as potentially
+	// spammable for rounds beyond zero.
+	//
+	// To drop such messages, the implementation below defensively uses a stronger
+	// condition of "nil justification with round larger than zero" to determine
+	// whether a message is "spammable".
+	beyondMaxLookaheadRounds := msg.Vote.Round > i.round+i.participant.maxLookaheadRounds
+	spammable := msg.Justification == nil && msg.Vote.Round > 0
+	if beyondMaxLookaheadRounds && spammable {
+		return nil
+	}
+
 	round := i.roundState(msg.Vote.Round)
 	switch msg.Vote.Step {
 	case QUALITY_PHASE:
--- a/gpbft/options.go
+++ b/gpbft/options.go
@@ -16,9 +16,12 @@ type Option func(*options) error
 type options struct {
 	delta                time.Duration
 	deltaBackOffExponent float64
+
+	initialInstance    uint64
+	maxLookaheadRounds uint64
+
 	// tracer traces logic logs for debugging and simulation purposes.
-	tracer          Tracer
-	initialInstance uint64
+	tracer Tracer
 }
 
 func newOptions(o ...Option) (*options, error) {
@@ -79,6 +82,18 @@ func WithInitialInstance(i uint64) Option {
 func WithTracer(t Tracer) Option {
 	return func(o *options) error {
 		o.tracer = t
+		return nil
+	}
+}
+
+// WithMaxLookaheadRounds sets the maximum number of rounds ahead of the current
+// round for which messages without justification are buffered. Setting a max
+// value of larger than zero would aid gPBFT to potentially reach consensus in
+// fewer rounds during periods of asynchronous broadcast as well as re-broadcast.
+// Defaults to zero if unset.
+func WithMaxLookaheadRounds(r uint64) Option {
+	return func(o *options) error {
+		o.maxLookaheadRounds = r
 		return nil
 	}
 }
--- a/sim/adversary/spam.go
+++ b/sim/adversary/spam.go
@@ -1,0 +1,85 @@
+package adversary
+
+import (
+	"github.com/filecoin-project/go-f3/gpbft"
+)
+
+var _ Receiver = (*Spam)(nil)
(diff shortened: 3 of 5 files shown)


sim/adversary/spam.go
```diff
@@ -0,0 +1,87 @@
...
+				Vote:      spamVote,
+				Signature: sig,
+				Ticket:    ticket,
+			})
+		}
+	}
+	return true, nil
+}
+
+func (r *Spam) ID() gpbft.ActorID                                              { return r.id }
```
Kubuxu, line 83: Will gofmt mess this up?

Kubuxu (approved):

sim/adversary/spam.go
masih: it shouldn't. It's standard `gofmt` formatting. If it was I believe `make lint` running default golangci-lint would have failed.

gpbft/gpbft.go
```diff
@@ -288,6 +288,20 @@ func (i *instance) receiveOne(msg *GMessage) error {
...
 	}
 
+	// Drop message that:
+	// * belong to future rounds, beyond the configured max lookahead threshold, and
+	// * carry no justification, i.e. are spammable.
+	//
+	// The only messages that are spammable are COMMIT for bottom. QUALITY and
+	// PREPARE messages may also not carry justification, but they are not spammable.
+	// Because, such messages are only valid at round 0; therefore, cannot belong
+	// to future rounds.
```
anorth, line 298: This is slightly misleading. PREPARE messages are valid at any round, but only those in round 0 are valid without justification. PREPARE for subsequent rounds must be justified.

gpbft/gpbft.go
```diff
@@ -288,6 +288,20 @@ func (i *instance) receiveOne(msg *GMessage) error {
...
+	beyondMaxLookaheadRounds := msg.Vote.Round > i.round+i.participant.maxLookaheadRounds
+	spammable := msg.Vote.Step == COMMIT_PHASE && msg.Vote.Value.IsZero()
```
anorth, line 300: Is is possible to change this predicate to `msg.Justification == nil && msg.Vote.Round > 0`? I feel it's just a little fragile in this form, and would rather that "only COMMIT bottom" is a description of the effect rather than the algorithm.

gpbft/options.go
```diff
@@ -17,8 +17,9 @@ type options struct {
 	delta                time.Duration
 	deltaBackOffExponent float64
 	// tracer traces logic logs for debugging and simulation purposes.
-	tracer          Tracer
-	initialInstance uint64
+	tracer             Tracer
```
anorth, line 20: Nit: keep tracer as the last item? Somehow group the algorithmically-relevant parameters from the coding artefacts.

gpbft/options.go
```diff
@@ -82,3 +83,14 @@ func WithTracer(t Tracer) Option {
 		return nil
 	}
 }
+
+// WithMaxLookaheadRounds sets the maximum number of rounds ahead of the current
+// round for which messages without justification are buffered. Setting a max
+// value of larger than zero would aid gPBFT to potentially reach consensus in
+// fewer rounds during re-broadcast. Defaults to zero if unset.
```
anorth, line 90: during periods of asynchrony, as well as re-broadcast

sim/adversary/spam.go
```diff
@@ -0,0 +1,87 @@
...
+			}
+
+			var ticket gpbft.Ticket
+			if len(msg.Ticket) != 0 {
+				var err error
+				ticket, err = gpbft.MakeTicket(beacon, spamVote.Instance, spamVote.Round, pubkey, r.host)
+				if err != nil {
+					panic(err)
+				}
+			}
```
anorth, line 71: Unnecessary since this doesn't send any CONVERGE messages

sim/adversary/spam.go
```diff
@@ -0,0 +1,87 @@
...
+				Signature: sig,
+				Ticket:    ticket,
+			})
+		}
+	}
+	return true, nil
+}
+
+func (r *Spam) ID() gpbft.ActorID                                              { return r.id }
+func (r *Spam) Start() error                                                   { return nil }
```
anorth, line 84: I'm a bit confused why this adversary waits to receive a COMMIT bottom from someone else (which may never happen) instead of just immediately sending the spam when started. You just need to know the instance number then you can generate as much spam as wanted right away. There's no need to derive it from other messages. (Generating more later on isn't a bad idea, but you could just set an alarm for that, or look at the clock when receiving any message).

anorth:
My comments are essentially nits so need not block you merging. I would appreciate you taking the time to explore whether a more robust predicate is possible, though.

gpbft/gpbft.go
masih: Oh i see the "vaild" wording; I will adjust thank you. What I meant was that the will not pass `validateMessage` without justification for non-zero rounds.

Issue #103, opened 2024-02-29
Location of participants in Lotus stack, multiple identities in one instance


Kubuxu:
For now, I've assumed that the f3 active participant would live in Lotus.
This might not have been as good of an assumption.

An active participant is tied to SP code and identity; that flow lives in lotus-miner/lotus-provider.
At the same time, the lotus miner is not connected with the global pubsub (AFAIK).

A single Lotus node can also host multiple providers, which, if f3 continues to live there, would necessitate either multiple concurrent instances in one Louts node or f3 being able to handle multiple identities.

As far as I know, the protocol flow is independent of our own identity, which should make running multiple identities at the same much easier. We could abstract out signing and VRF generation as part of the broadcast operation. Essentially, the instance itself stops caring about our own identity.

Kubuxu:
The design I settled on:
- gpbft.Participant is unaware of the ID it is running as the protocol is independent of our own decisions
  - this requires a slight refactor and cleanup in gpbft to remove ParticipantID
- gpbft requests a given message to be broadcasted, that message is universal (not connected with any ID), this is passed to the Host which knows as which IDs it wants to broadcast, and uses powertable from gpbft to resolve IDs to public keys
- The host then builds a payload to be signed with that key. Signing can happen over RPC boundary (necessary for offloading keys to lotus-miner). This results in serialized payloads that are to be signed with the given key and returned back to the host for broadcasting
- When these payloads are returned to be broadcasted, they are also immediately processed as incoming messages

See https://github.com/filecoin-project/go-f3/pull/188 for PR on it. The suggestion there was to separate the PR into:
1. async delivery of local messages
2. removal of ID from gpbft.Participant
3. addition of the message builder pattern

anorth:
This has proven slightly trickier than we thought. The protocol code has assumptions that it can update state as a result of its own decisions more or less synchronously. This used to be achieved by internally receiving messages sent by self synchronously. This changed to async as part of #259 but means there's now a race between receiving those messages sent to self and receiving alarms #316.

I explored an internal send-to-self of unvalidated messages, but immediately ran into not knowing the node's own power, so the state updates are impossible.

Some potential paths forward that I see:
- Attempt to remove any algorithmic assumption of being able to update state as a result of the participant's own decisions. I don't know how easy this will be. We haven't tried.
[...]

PR #259, opened 2024-05-21
FFF-103: Async delivery of local messages

adlrocha:
This PR partially addresses #103 and implements and addresses the reviews from #188 related to the broadcast interface and the async delivery of local messages.
- It introduces a `RequestBroadcast` function to the `Network` interface that delivers local messages aynchronously.
- Unifies `Broadcast` and `BroadcastSynchronous`  into a single `Broadcast` function.
- Introduces `SimNetwork` interface as an intermediary step to the removal of IDs from gpbft.Participant (coming next)

### Coming next
Two more PRs extracted from #188 with:
- removal of ID from gpbft.Participant
- addition of the message builder pattern

### DIFF
--- a/gpbft/api.go
+++ b/gpbft/api.go
@@ -38,9 +38,8 @@ type Chain interface {
 type Network interface {
 	// Returns the network's name (for signature separation)
 	NetworkName() NetworkName
-	// Sends a message to all other participants.
-	// The message's sender must be one that the network interface can sign on behalf of.
-	Broadcast(msg *GMessage)
+	// Requests that the message is signed and broadcasted, it should also be delivered locally
+	RequestBroadcast(msg *GMessage)
 }
 
 type Clock interface {
--- a/gpbft/gpbft.go
+++ b/gpbft/gpbft.go
@@ -155,8 +155,6 @@ type instance struct {
 	// This field is an alternative to plumbing an optional decision value out through
 	// all the method calls, or holding a callback handle to receive it here.
 	terminationValue *Justification
-	// Queue of messages to be synchronously processed before returning from top-level call.
-	inbox []*GMessage
 	// Quality phase state (only for round 0)
 	quality *quorumState
 	// State for each round of phases.
@@ -212,10 +210,7 @@ func newRoundState(powerTable PowerTable) *roundState {
 }
 
 func (i *instance) Start() error {
-	if err := i.beginQuality(); err != nil {
-		return err
-	}
-	return i.drainInbox()
+	return i.beginQuality()
 }
 
 // Checks whether a message is valid.
@@ -233,45 +228,15 @@ func (i *instance) Receive(msg *GMessage) error {
 	if i.terminated() {
 		return fmt.Errorf("senders message after decision")
 	}
-	if len(i.inbox) > 0 {
-		return fmt.Errorf("senders message while already processing inbox")
-	}
-
-	// Enqueue the message for synchronous processing.
-	i.enqueueInbox(msg)
-	return i.drainInbox()
+	return i.receiveOne(msg)
 }
 
 func (i *instance) ReceiveAlarm() error {
-	if err := i.tryCurrentPhase(); err != nil {
-		return fmt.Errorf("failed completing protocol phase: %w", err)
-	}
-
-	// A phase may have been successfully completed.
-	// Re-process any queued messages for the next phase.
-	return i.drainInbox()
+	return i.tryCurrentPhase()
 }
 
 func (i *instance) Describe() string {
 	return fmt.Sprintf("P%d{%d}, round %d, phase %s", i.participant.id, i.instanceID, i.round, i.phase)
-}
-
-func (i *instance) enqueueInbox(msg *GMessage) {
-	i.inbox = append(i.inbox, msg)
-}
-
-func (i *instance) drainInbox() error {
-	for len(i.inbox) > 0 {
-		// Process one message.
-		// Note the message being processed is left in the inbox until after processing,
-		// as a signal that this loop is currently draining the inbox.
-		if err := i.receiveOne(i.inbox[0]); err != nil {
-			return fmt.Errorf("failed receiving message: %w", err)
-		}
-		i.inbox = i.inbox[1:]
-	}
-
-	return nil
 }
 
 // Processes a single message.
--- a/gpbft/gpbft.go
+++ b/gpbft/gpbft.go
@@ -816,7 +816,6 @@ func (i *instance) broadcast(round uint64, step Phase, value ECChain, ticket Tic
 		Justification: justification,
 	}
 	i.participant.host.RequestBroadcast(gmsg)
-	i.enqueueInbox(gmsg)
 }
 
 // Sets an alarm to be delivered after a synchrony delay.
--- a/gpbft/gpbft.go
+++ b/gpbft/gpbft.go
@@ -815,7 +815,7 @@ func (i *instance) broadcast(round uint64, step Phase, value ECChain, ticket Tic
 		Ticket:        ticket,
 		Justification: justification,
 	}
-	i.participant.host.Broadcast(gmsg)
+	i.participant.host.RequestBroadcast(gmsg)
 	i.enqueueInbox(gmsg)
 }
 
--- a/sim/adversary/adversary.go
+++ b/sim/adversary/adversary.go
@@ -13,7 +13,8 @@ type Host interface {
 	// Sends a message to all other participants, immediately.
 	// Note that the adversary can subsequently delay delivery to some participants,
 	// before messages are actually received.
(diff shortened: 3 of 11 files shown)


sim/network.go
```diff
@@ -61,24 +61,68 @@ func (n *Network) AddParticipant(p gpbft.Receiver) {
...
+}
+
+type networkFor struct {
+	ParticipantID gpbft.ActorID
+	Signer        gpbft.Signer
+	*Network
+}
+
+// Log implements tagging Tracer interface
+func (nf *networkFor) Log(format string, args ...interface{}) {
```
masih, line 81:
```suggestion
func (nf *networkFor) Log(format string, args ...any) {
```

masih (commented):
LGTM 👍  

This PR would benefit from a review from @anorth as the reviewer of the PR from which changes here are derived from 🙏

gpbft/gpbft.go
```diff
@@ -815,7 +815,7 @@ func (i *instance) broadcast(round uint64, step Phase, value ECChain, ticket Tic
 		Ticket:        ticket,
 		Justification: justification,
 	}
-	i.participant.host.Broadcast(gmsg)
+	i.participant.host.RequestBroadcast(gmsg)
 	i.enqueueInbox(gmsg)
```
Kubuxu, line 819: This line should be removed, given that the broadcast is done async.

adlrocha: Good catch, thanks!

sim/network.go
masih: @adlrocha how come this is resolved? 😆

adlrocha: Thank you for flagging it, I fixed it somewhere else and forgot to push the one you commented: https://github.com/filecoin-project/go-f3/pull/259/commits/23a55118c16eb19fec6f76da605e352112748495#diff-4fa4e1c1a7b22e16444c976a3fecfaf2d47f9c48ca7d1ebcb955034e9daab511L148

sim/adversary/adversary.go
```diff
@@ -13,7 +13,8 @@ type Host interface {
 	// Sends a message to all other participants, immediately.
 	// Note that the adversary can subsequently delay delivery to some participants,
 	// before messages are actually received.
-	BroadcastSynchronous(sender gpbft.ActorID, msg gpbft.GMessage)
+	// The sync flag can be used to determine how the message is broadcast.
+	Broadcast(sender gpbft.ActorID, msg *gpbft.GMessage, synchronouos bool)
```
anorth, line 17: It's now much less obvious at the call sites when the special adversary powers are being invoked. I would prefer this method were named specially when the synchronous behaviour is invoked, and adversaries that aren't using it just use the same method as the normal participant.

gpbft/api.go
```diff
@@ -38,9 +38,8 @@ type Chain interface {
 type Network interface {
 	// Returns the network's name (for signature separation)
 	NetworkName() NetworkName
-	// Sends a message to all other participants.
-	// The message's sender must be one that the network interface can sign on behalf of.
-	Broadcast(msg *GMessage)
+	// Requests that the message is signed and broadcasted, it should also be delivered locally
+	RequestBroadcast(msg *GMessage)
```
anorth, line 42: Changing semantics is fine. I don't think we need to rename this method if the old behaviour is gone.

sim/network.go
```diff
@@ -61,24 +61,68 @@ func (n *Network) AddParticipant(p gpbft.Receiver) {
...
+
+type networkFor struct {
+	ParticipantID gpbft.ActorID
+	Signer        gpbft.Signer
+	*Network
+}
+
+// Log implements tagging Tracer interface
+func (nf *networkFor) Log(format string, args ...any) {
+	nf.Network.log(TraceLogic, "P%d "+format, append([]any{nf.ParticipantID}, args...)...)
```
anorth, line 82: Thanks. Please enable tracing in some test and check that the logs look reasonable. You've added a prefix here but I didn't spot where the existing prefix has been removed from call sites, so it might be duplicated now.

anorth (commented):
Not blocking, but:
- please check the logs are good
- I'd prefer the adversary special powers were more obvious when used

sim/network.go
adlrocha: Good catch, thanks. This was in preparation for the ID removal, but after working in #267 I realized with my proposal is no longer needed, so I will remove it from here.

gpbft/api.go
adlrocha: I went with this approach because naming it `Broadcast` leads to a collision among the interfaces used by `Host`.

sim/adversary/adversary.go
adlrocha: I see, I went with this approach per this comment: https://github.com/filecoin-project/go-f3/pull/188/files#r1595990496 but this makes sense. I'll revert to how it was before.

Kubuxu:
@anorth could you re-review?

sim/adversary/repeat.go
```diff
@@ -100,7 +100,7 @@ func (r *Repeat) ReceiveMessage(msg *gpbft.GMessage, _ bool) (bool, error) {
 		Ticket:        ticket,
 	}
 	for i := 0; i < echoCount; i++ {
-		r.host.Broadcast(echo)
+		r.host.Broadcast(echo.Sender, echo)
```
anorth, line 103: This can just use the normal `RequestBroadcast` method. It doesn't need to pass the sender explicitly.

sim/adversary/spam.go
```diff
@@ -71,7 +71,7 @@ func (s *Spam) spamAtInstance(instance uint64) {
 		if err != nil {
 			panic(err)
 		}
-		s.host.Broadcast(&gpbft.GMessage{
+		s.host.Broadcast(s.ID(), &gpbft.GMessage{
```
anorth, line 74: Ditto use the vanilla `RequestBroadcast` here.

sim/network.go
```diff
@@ -61,24 +61,76 @@ func (n *Network) AddParticipant(p gpbft.Receiver) {
...
-			n.queue.Insert(
-				&messageInFlight{
-					source:    msg.Sender,
-					dest:      dest,
-					payload:   *msg,
-					deliverAt: now.Add(delay),
-				})
+		if dest == sender {
+			continue
+		}
```
anorth, line 120: I think we want to deliver locally here too, with the same logic as `RequestBroadcast`. I don't think we need the sender parameter at all. I think the only difference here is the synchronous mode. You should be able to delegate to this from `RequestBroadcast` too.

gpbft/api.go
anorth: I don't think you need that method in the adversary host, though.

sim/adversary/adversary.go
```diff
@@ -10,10 +10,12 @@ type Receiver interface {
 // Endpoint with which the adversary can control the network
 type Host interface {
 	gpbft.Host
+	// Sends a message to all other participants
+	Broadcast(sender gpbft.ActorID, msg *gpbft.GMessage)
```
anorth, line 14: I think this method is unnecessary. All callers use the message sender as the sender parameter. Now that this PR is delivering messages locally, I don't think we need a sender parameter at all, not in BroadcastSynchronous either.

sim/network.go
anorth: Prefix is still here...

gpbft/api.go
adlrocha: You are right, this is no longer necessary. Thanks!

sim/network.go
adlrocha: Removed it from this review, as it adds nothing and it will make more sense once we address the review of my followup PRs. Sorry for the unnecessary noise.

adlrocha:
@anorth, thank you for your review. I think I addressed them all. Let me know if there's still something pending.

adlrocha:
@masih, @Kubuxu, actually, I realized that probably we do not need to merge the PRs as a chain because they are not strongly dependent. Let's get this one merged to `main`, and I will point to `main` from #274 and have a final review on that one before merging. Thanks! :pray:

anorth (approved):
Not sure what happened to your CI tests, but LGTM now. Thanks!

anorth:
Running tests locally:
- In one instance, panic with `fatal error: concurrent map writes`. This is most likely to be sharing of the BLS signing backend between concurrent tests
- In other instances seem to hang spinning forever (seems like what's happening in CI too, with the container then being killed_

anorth:
I think the repeat adversary is now receiving and repeating the messages that it has sent itself. This is because of the change in semantics of requesting broadcast.

adlrocha:
Thank you for the feedback, @anorth. You were right, I think I found it and fixed it in this commit: https://github.com/filecoin-project/go-f3/pull/259/commits/4a6f98ae7b44a5f3883756d4b926af17e3fad9ae. Let's wait for CI see if it is happy.

PR #264, opened 2024-05-21
Use offset committees to validate and queue/drop messages for future instances

anorth:
This covers the main parts of https://github.com/filecoin-project/go-f3/issues/151 except testing/validation. I added new tasks for follow-up items of optimising skipping rounds, and supporting parallel validation.

FYI @Kubuxu on the API change.

### DIFF
--- a/gpbft/gpbft.go
+++ b/gpbft/gpbft.go
@@ -214,20 +214,24 @@ func (i *instance) Start() error {
 	return i.beginQuality()
 }
 
-// Checks whether a message is valid.
-// An invalid message can never become valid, so may be dropped.
-// This method is read-only and inspects only immutable state, so should be safe to invoke
-// concurrently.
-func (i *instance) Validate(msg *GMessage) error {
-	return i.validateMessage(msg)
-}
-
-// Receives a validated message.
-// This method will not attempt to validate the message, the caller must ensure the message
-// is valid before calling this method.
+// Receives and processes a message.
+// Returns an error indicating either message invalidation or a programming error.
 func (i *instance) Receive(msg *GMessage) error {
+	// Check the message is for this instance, to guard against programming error.
+	if msg.Vote.Instance != i.instanceID {
+		// Return true here to cause a loud failure at the caller, which should not have
+		// passed this message here.
+		return fmt.Errorf("message for instance %d, expected %d: %w",
+			msg.Vote.Instance, i.instanceID, ErrReceivedWrongInstance)
+	}
+	// Perform validation that could not be done until the instance started.
+	if !(msg.Vote.Value.IsZero() || msg.Vote.Value.HasBase(i.input.Base())) {
+		return fmt.Errorf("message base %s, expected %s: %w",
+			&msg.Vote.Value, i.input.Base(), ErrValidationWrongBase)
+	}
+
 	if i.terminated() {
-		return fmt.Errorf("senders message after decision")
+		return ErrReceivedAfterTermination
 	}
 	return i.receiveOne(msg)
 }
@@ -260,22 +264,8 @@ func (i *instance) receiveOne(msg *GMessage) error {
 	// Drop message that:
 	//  * belong to future rounds, beyond the configured max lookahead threshold, and
 	//  * carry no justification, i.e. are spammable.
-	//
-	// The only messages that are spammable are COMMIT for bottom. QUALITY and
-	// PREPARE messages may also not carry justification, but they are not
-	// spammable. Because:
-	//  * QUALITY is only valid for round zero.
-	//  * PREPARE must carry justification for non-zero rounds.
-	//
-	// Therefore, we are only left with COMMIT for bottom messages as potentially
-	// spammable for rounds beyond zero.
-	//
-	// To drop such messages, the implementation below defensively uses a stronger
-	// condition of "nil justification with round larger than zero" to determine
-	// whether a message is "spammable".
 	beyondMaxLookaheadRounds := msg.Vote.Round > i.round+i.participant.maxLookaheadRounds
-	spammable := msg.Justification == nil && msg.Vote.Round > 0
-	if beyondMaxLookaheadRounds && spammable {
+	if beyondMaxLookaheadRounds && isSpammable(msg) {
 		return nil
 	}
 
@@ -372,25 +362,18 @@ func (i *instance) tryCurrentPhase() error {
 }
 
 // Checks message validity, including justification and signatures.
-func (i *instance) validateMessage(msg *GMessage) error {
-	// Check the message is for this instance.
-	// The caller should ensure this is always the case.
-	if msg.Vote.Instance != i.instanceID {
-		return xerrors.Errorf("message for wrong instance %d, expected %d", msg.Vote.Instance, i.instanceID)
-	}
+// An invalid message can never become valid, so may be dropped.
+// This is a pure function and does not modify its arguments.
+func ValidateMessage(powerTable *PowerTable, beacon []byte, host Host, msg *GMessage) error {
 	// Check sender is eligible.
-	senderPower, senderPubKey := i.powerTable.Get(msg.Sender)
+	senderPower, senderPubKey := powerTable.Get(msg.Sender)
 	if senderPower == nil || senderPower.Sign() == 0 {
-		return xerrors.Errorf("sender with zero power or not in power table")
+		return xerrors.Errorf("sender %d with zero power or not in power table", msg.Sender)
 	}
 
 	// Check that message value is a valid chain.
 	if err := msg.Vote.Value.Validate(); err != nil {
 		return xerrors.Errorf("invalid message vote value chain: %w", err)
-	}
-	// Check the value is acceptable.
-	if !(msg.Vote.Value.IsZero() || msg.Vote.Value.HasBase(i.input.Base())) {
-		return xerrors.Errorf("unexpected base %s", &msg.Vote.Value)
 	}
 
 	// Check phase-specific constraints.
@@ -409,7 +392,7 @@ func (i *instance) validateMessage(msg *GMessage) error {
 		if msg.Vote.Value.IsZero() {
 			return xerrors.Errorf("unexpected zero value for converge phase")
 		}
-		if !VerifyTicket(i.beacon, i.instanceID, msg.Vote.Round, senderPubKey, i.participant.host, msg.Ticket) {
+		if !VerifyTicket(beacon, msg.Vote.Instance, msg.Vote.Round, senderPubKey, host, msg.Ticket) {
 			return xerrors.Errorf("failed to verify ticket from %v", msg.Sender)
 		}
 	case DECIDE_PHASE:
@@ -426,8 +409,8 @@ func (i *instance) validateMessage(msg *GMessage) error {
 	}
 
 	// Check vote signature.
-	sigPayload := i.participant.host.MarshalPayloadForSigning(&msg.Vote)
-	if err := i.participant.host.Verify(senderPubKey, sigPayload, msg.Signature); err != nil {
+	sigPayload := host.MarshalPayloadForSigning(&msg.Vote)
+	if err := host.Verify(senderPubKey, sigPayload, msg.Signature); err != nil {
 		return xerrors.Errorf("invalid signature on %v, %v", msg, err)
 	}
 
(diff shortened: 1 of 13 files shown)


gpbft/gpbft.go
```diff
@@ -373,26 +364,19 @@ func (i *instance) tryCurrentPhase() error {
...
 	}
 
 	// Check that message value is a valid chain.
 	if err := msg.Vote.Value.Validate(); err != nil {
 		return xerrors.Errorf("invalid message vote value chain: %w", err)
 	}
-	// Check the value is acceptable.
-	if !(msg.Vote.Value.IsZero() || msg.Vote.Value.HasBase(i.input.Base())) {
-		return xerrors.Errorf("unexpected base %s", &msg.Vote.Value)
-	}
```
anorth, line 395: This can't be checked until the instance starts, so is deferred until then.

gpbft/participant.go
```diff
@@ -11,14 +12,14 @@ type Participant struct {
 	id   ActorID
 	host Host
 
-	// Instance identifier for the next Granite instance.
-	nextInstance uint64
+	// Instance identifier for the current (or, if none, next to start) GPBFT instance.
+	currentInstance uint64
```
anorth, line 16: Note change in semantics from next to current – please review this carefully.

gpbft/participant.go
```diff
@@ -66,63 +68,62 @@ func (p *Participant) Start() (err error) {
...
+//   - A message for a previous instance is dropped.
+//   - A message for the current instance is validated and delivered immediately.
+//   - A message for a future instance for which a committee is available is validated and
+//     locally queued, to be delivered when that instance begins.
+//   - A message for a future instance for which a committee is not available returns an error.
+//   - An invalid message returns an error.
+//
+// Returns whether the message was accepted for the current instance or a near future instance,
+// and/or an error if it was invalid or could not be processed.
+func (p *Participant) ReceiveMessage(msg *GMessage) (accepted bool, err error) {
```
anorth, line 87: Please give feedback on the return values here. Errors could come either from message validation-related things, which can pretty safely be ignored, or from other more serious sources. The boolean return is relevant whether or not error is nil. Roughly, if false then errors are inconsequential, and if true then errors are serious. But this breaks the usual pattern of ignoring return value with non-nil errors.

One alternative would be to return `(validation error, err error)` two possibly-nil errors, a traditional one and one indicating validation results.

Kubuxu: Another option would be to have a sentinel error value which gets wrapped with a cause.

Kubuxu, line 87: I have a slight issue with API here.
1. We know we need to leverage multithreading for signature validation. Otherwise, validating the signatures can be a bottleneck.
2. Participant does not handle concurrency
3. libp2p/pubsub validation by default is multithreaded

We could do validation from outside by fetching comities and calling `ValidateMessage,` but this validation is duplicated.

gpbft/participant.go
```diff
@@ -157,57 +159,128 @@ func (p *Participant) beginInstance() error {
...
+
+// A power table and beacon value used as the committee inputs to an instance.
+type committee struct {
+	power  *PowerTable
+	beacon []byte
+}
+
+// A collection of messages queued for delivery for a future instance.
+// The queue drops equivocations and unjustified messages beyond some round number.
+type messageQueue struct {
```
masih, line 239: Might be worth keeping it sorted with a priority queue, similar to what's done for message queue in `sim` package. 

We can revisit this with benchmarks at a later date.

gpbft/participant.go
```diff
@@ -157,57 +159,128 @@ func (p *Participant) beginInstance() error {
...
+	instanceQueue[msg.Sender] = append(instanceQueue[msg.Sender], msg)
+}
+
+func (q *messageQueue) Drain(instance uint64) []*GMessage {
+	var msgs []*GMessage
+	for _, ms := range q.messages[instance] {
+		msgs = append(msgs, ms...)
+	}
+	// Sort by step so messages will be processed in a useful order.
+	sort.Slice(msgs, func(i, j int) bool {
```
masih, line 281: Do we care about the sort stability? We probably should for testing purposes.

sim/adversary/absent.go
```diff
@@ -40,7 +40,7 @@ func (a *Absent) ValidateMessage(_ *gpbft.GMessage) (bool, error) {
 	return true, nil
 }
 
-func (a *Absent) ReceiveMessage(_ *gpbft.GMessage, _ bool) (bool, error) {
+func (a *Absent) ReceiveMessage(_ *gpbft.GMessage) (bool, error) {
```
masih, line 43:
```suggestion
func (*Absent) ReceiveMessage(*gpbft.GMessage) (bool, error) {
```

sim/adversary/decide.go
```diff
@@ -78,7 +78,7 @@ func (i *ImmediateDecide) ValidateMessage(_ *gpbft.GMessage) (bool, error) {
 	return true, nil
 }
 
-func (i *ImmediateDecide) ReceiveMessage(_ *gpbft.GMessage, _ bool) (bool, error) {
+func (i *ImmediateDecide) ReceiveMessage(_ *gpbft.GMessage) (bool, error) {
```
masih, line 81:
```suggestion
func (*ImmediateDecide) ReceiveMessage(*gpbft.GMessage) (bool, error) {
```

sim/adversary/withhold.go
```diff
@@ -118,7 +118,7 @@ func (a *WithholdCommit) ValidateMessage(_ *gpbft.GMessage) (bool, error) {
 	return true, nil
 }
 
-func (a *WithholdCommit) ReceiveMessage(_ *gpbft.GMessage, _ bool) (bool, error) {
+func (a *WithholdCommit) ReceiveMessage(_ *gpbft.GMessage) (bool, error) {
```
masih, line 121:
```suggestion
func (*WithholdCommit) ReceiveMessage(*gpbft.GMessage) (bool, error) {
```

gpbft/participant.go
masih: Could you expand further please, Kuba?

If signature validation is the only pace we would need parallelism, would it be sufficient to encapsulate it within `Signer` interface?

masih (approved):

gpbft/participant.go
Kubuxu: The validation from pubsub is multithreaded (and I would like to keep it like this). We could verify signatures separately and then take a lock to validate the rest, but at that point, we might as well make the whole validation thread safe.

masih: Summary of discussion with @Kubuxu at standup: 

There are three separate issues:
1) the need for validation as a self-contained functionality which may or may not result in a consecutive call to receive message.
2) avoiding re-validation of messages that are already validated.
3) parallel validation of messages in terms of performance concerns.

Regarding the first issue, GossipSub has a separate validate flow and handle flow, where a client provides a validation hook that is called by GossipSub upon receiving messages and is required to be safe for parallel execution. A separate validate flow in GossipSub is there to reduce the possibility of accidentally spamming the network, and only propagate messages that have at least passed validation of honest nodes. Therefore, we need some self-contained way of offering that functionality in go-f3.

Regarding the second issue, although implicit message validation by the `Receiver` interface in F3 is reasonable defensive programming, it means we will end up with messages getting validated twice: once by the gossipSub validation mechanism mentioned earlier and once internally by F3 implementation. This may be OK if validation is cheap? but ideally we should offer a way avoid this.

Regarding the third issue, we should make message validation as fast to execute as possible while keeping it safe for parallel execution. This can be tricky since validation at certain steps depends on internal state of the participant such as instance ID, phase, etc.

Solutions discussed:
* _Revert back to the previous API_: The previous gPBFT API, facilitated conditional validation on receive message. This meant that the GossipSub integration layer could signal to the receiver that a message is already validated and the implementation may skip validation. But this can be error prone.
* _Introduce dedicated `Validator` interface that returns `ValidatedMessage` type which is then taken by the `Receiver`_: this could work but might be too much typing/combersome to work with.
* _Unexported internal struct fields that can signal conditional validation_: This would work but unclear where it leaves the exported gPBFT API.

This API change blocks the gossipSub integration and we should resolve this soon. An alternative there could be to take the first solution above and revisit once both this PR and gossipSub integration are landed. Then revisit.

@Kubuxu Please let me know if there is anything that I might have missed from our discussion. 🙏

gpbft/participant.go
anorth: No harm in using a stable sort 👍 . But also this should sort by round too!

gpbft/participant.go
anorth: Thanks for the discussion. Validation is not cheap: validating aggregate signatures on message justification is probably the most CPU-expensive thing in the whole protocol. I understand the need to support parallelism and will adjust the API to support this. I'll try the `Validator`/`ValidatedMessage` approach.

gpbft/gpbft.go
```diff
@@ -218,28 +218,33 @@ func (i *instance) Start() error {
...
-}
+// Receives and validates a message.
+// Returns whether the message was valid:
+// - If false the message may be safely dropped.
+// - If true then any error indicates an internal failure to process the message.
+func (i *instance) Receive(msg *GMessage) (bool, error) {
+	// Check the message is for this instance, to guard against programming error.
+	if msg.Vote.Instance != i.instanceID {
+		// Return true here to cause a loud failure at the caller, which should not have
+		// passed this message here.
```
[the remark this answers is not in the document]
anorth: We could panic here instead (it'll be caught and wrapped to the caller). Existing panics are reserved for when protocol-level assumptions are violated, but this is an internal assumption.

gpbft/participant.go
```diff
@@ -66,63 +68,62 @@ func (p *Participant) Start() (err error) {
...
-// Returns whether the message was for the current instance, and an error if it was invalid or
-// could not be processed.
-func (p *Participant) ReceiveMessage(msg *GMessage, validated bool) (accepted bool, err error) {
-	defer func() {
-		if r := recover(); r != nil {
-			err = &PanicError{Err: r}
-		}
-	}()
+	// Fetch the committee against which to validate the message.
+	comt, err := p.getCommittee(msg.Vote.Instance)
```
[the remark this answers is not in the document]
anorth: I expect them to be cheap: the host will check if `instance - 10` is finalised, and fail if not.

gpbft/participant.go
```diff
@@ -66,63 +68,62 @@ func (p *Participant) Start() (err error) {
...
-	} else if msg.Vote.Instance >= p.nextInstance {
-		// Locally queue message for a future instance.
-		p.mqueue[msg.Vote.Instance] = append(p.mqueue[msg.Vote.Instance], msg)
+		p.handleDecision()
+		return true, nil
+	} else {
+		// Otherwise queue it for a future instance.
+		p.mqueue.Add(msg)
+		return true, nil
 	}
```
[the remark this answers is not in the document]
anorth: I pulled the return statement out of the `if/else` instead.

gpbft/participant.go
anorth: The only inspection of this queue is when it's completely drained, so I expect a single sort at the end will be more efficient than doing so incrementally.

anorth:
Okay, I have split the API into validation and receiving once again. I have also completely reworked the error return scheme to use sentinel errors. Please let me know what you think of this version.

It's not yet safe for concurrent execution, but I have noted a follow-up task for this in #151.

gpbft/participant.go
```diff
@@ -28,6 +29,16 @@ type Participant struct {
 	terminatedDuringRound uint64
 }
 
+type validatedMessage struct {
+	msg *GMessage
+}
+
```
anorth, line 35: I ended up duplicating this structure two more times in tests. I could instead just make it a public exported type in the GPBFT package. This would enable trivial bypassing of message validation. I'm unsure of its potential for accidental use.

masih: Alternatively you can move the tests that use it to `gpbft` package instead of `gpbft_test`.

gpbft/participant.go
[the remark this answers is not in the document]
masih: CC @Kubuxu to make sure we check this in host. Thanks

masih (approved):
LGTM 👍 

@Kubuxu, how does this look for GossipSub integration?

Kubuxu (approved):

PR #265, opened 2024-05-21
Order-independent fake signatures

Stebalien:
Make fake signatures produce the same results independent of order of the signers (assumes signatures are unique).

### DIFF
--- a/sim/signing/fake.go
+++ b/sim/signing/fake.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 
 	"github.com/filecoin-project/go-f3/gpbft"
+	"slices"
 )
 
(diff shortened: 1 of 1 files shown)


Kubuxu:
The BLS scheme we use is order dependant, which is why the fake signature scheme replicates it.

Stebalien:
> The BLS scheme we use is order dependant, which is why the fake signature scheme replicates it.

I see. And we expect to sort by power descending, ID ascending (same as the power table)?

Kubuxu:
Whatever is easier. We settled on power descending because we store signers as bitfields of indexes in the power table.

Stebalien:
Yeah, fair enough.

PR #266, opened 2024-05-22
Implement finality certificates

### LINKED ISSUE #5: Design/implement power table representation
The finality decision exchange protocol calls for a power table as a simple sequence of tuples of (ActorID, Power, PubKey), with a consistent serialization and CID. Entries are looked up by simple integer index for finality certificate validation. The table can be inflated into a map for runtime use by the protocol code.

See also #33 re truncation of power values.

```[tasklist]
### Tasks
- [ ] Implement an IPLD structure for the list-of-tuples
- [x] Add power table CID to core protocol messages
- [ ] Implement calculation of deltas between two tables, and application of deltas to a table
- [ ] #166
```

### LINKED ISSUE #14: Implement finality certificates
Implement the creation of finality certificates from a strong quorum of signed DECIDE messages. See [design doc](https://docs.google.com/document/d/10i9tFremOSrZou9oO5A5wvu1uOy1lvFKbv8IsvoglR0/edit).

- Compute BLS aggregate signature
- Compute power table deltas
- Compute compressed participant-set representation
- Implement certificate validation

```[tasklist]
### Dependencies
- [ ] #5
- [ ] #4
- [ ] https://github.com/filecoin-project/go-f3/issues/150
```

Stebalien:
This implements finality certificates and power table deltas. This adds:

1. Methods for applying/making power table deltas.
2. A method for validating a chain of finality certificates.

Notes:

- Finality certificates include the power-table delta between the power table used to validate the current certificate and the next certificate. They don't include the delta implied by the chain head. We'll have to address this when we implement the finality certificate exchange protocol, but I expect we'll communicate these power-deltas separate from the certificates.

- I used `PowerEntries` instead of `PowerTable` objects because I doubt we'll need the performance. But I'm happy to change that.

fixes #5
fixes #14


Kubuxu:
One early note: it should live in a separate package. The finality certificate is quite separate from the gpbft itself.

Stebalien:
@Kubuxu I've moved it to a `certs` package, and renamed `certs` to `certstore`.

I now remember why I didn't do this: I was hoping to find a way to completely disconnect gpbft and finality certificates such that finality certificates didn't depend on gpbft. But... that's not critical and something we can look into later (gpbft itself doesn't have any external dependencies anyways).

certs/certs.go
```diff
@@ -0,0 +1,293 @@
...
+		if cert.GPBFTInstance != nextInstance {
+			return nextInstance, chain, prevPowerTable, xerrors.Errorf("expected instance %d, found instance %d", nextInstance, cert.GPBFTInstance)
+		}
+		// Some basic sanity checks.
+		if err := cert.ECChain.Validate(); err != nil {
+			return nextInstance, chain, prevPowerTable, xerrors.Errorf("invalid finality certificate at instance %d: %w", cert.GPBFTInstance, err)
+		}
+
+		// We can't have a finality certificate for "bottom"
+		if cert.ECChain.IsZero() {
```
Kubuxu, line 98: Is that right? I'm not certain. I'm pretty sure the GPBFT can terminate with a bottom result.

Kubuxu: Or will the justification be for the Base?

Kubuxu (approved):
Other than that one comment. SGWM

certs/certs.go
Stebalien: From what I can tell, we cannot "decide" zero: https://github.com/filecoin-project/go-f3/blob/4f6464d9c33fae650918347bcdb06b5911922e5b/gpbft/gpbft.go#L430-L432

We can commit to zero to start another round, but we'll keep going until we decide on _at least_ the base.

certs/certs.go
```diff
@@ -0,0 +1,293 @@
...
+		return nil, xerrors.Errorf("can only create a finality certificate from a decide vote, got phase %s", justification.Vote.Step)
+	}
+
+	if justification.Vote.Round != 0 {
+		return nil, xerrors.Errorf("expected decide round to be 0, got round %d", justification.Vote.Round)
+	}
+
+	if justification.Vote.Value.IsZero() {
+		return nil, xerrors.Errorf("got a decision for bottom for instance %d", justification.Vote.Instance)
+	}
```
anorth, line 64: More preconditions you might check:
- Vote value is not empty, and does Validate()
- Signers is not empty?
- Supplemental Merkel hash is the right length?

Edit: I see some of these are captured below, but perhaps some of these easier, pure-syntax ones would be better earlier?

certs/certs.go
```diff
@@ -0,0 +1,293 @@
...
+// power table to use as-of the last valid finality certificate. If passed a non-nil `base` tipset,
+// validate that the finalized chain starts with that tipset (accept any finalized chain otherwise).
+//
+// Returns an error if it encounters any invalid finality certificates, along with the last valid
+// instance, finalized chain epochs, etc. E.g., if provided a partially valid chain of finality
+// certificates, this function will return a (possibly empty) prefix of the EC chain correctly
+// finalized, the instance of the first invalid finality certificate, and the power table that
+// should be used to validate that finality certificate, along with the error encountered.
+func ValidateFinalityCertificates(verifier gpbft.Verifier, network gpbft.NetworkName, prevPowerTable gpbft.PowerEntries, nextInstance uint64, base *gpbft.TipSet,
+	certs ...FinalityCertificate) (_nextInstance uint64, chain gpbft.ECChain, newPowerTable gpbft.PowerEntries, err error) {
```
anorth, line 87: Channelling @masih , consider introducing structs for these high arity parameter and return types.

certs/certs.go
```diff
@@ -0,0 +1,293 @@
...
+		if err != nil {
+			return nextInstance, chain, prevPowerTable, xerrors.Errorf("failed to apply power table delta for finality certificate for instance %d: %w", cert.GPBFTInstance, err)
+		}
+
+		powerTableCid, err := MakePowerTableCID(newPowerTable)
+		if err != nil {
+			return nextInstance, chain, prevPowerTable, xerrors.Errorf("failed to make power table CID for finality certificate for instance %d: %w", cert.GPBFTInstance, err)
+		}
+
+		if !bytes.Equal(cert.SupplementalData.PowerTable, powerTableCid) {
```
anorth, line 125: Nit: seeing this in use, perhaps we can make this name more obvious. E.g. SupplementalData.NextBasePowerTable? It's still tricky though...

certs/certs.go
```diff
@@ -0,0 +1,293 @@
...
+// any other parts of the certificate, just that the _value_ has been signed by a majority of the
+// power.
+func verifyFinalityCertificateSignature(verifier gpbft.Verifier, powerTable gpbft.PowerEntries, nn gpbft.NetworkName, cert FinalityCertificate) error {
+	totalPower := new(gpbft.StoragePower)
+	for i := range powerTable {
+		totalPower = totalPower.Add(totalPower, powerTable[i].Power)
+	}
+
+	signers := make([]gpbft.PubKey, 0, len(powerTable))
+	signerPower := new(gpbft.StoragePower)
```
anorth, line 149: Nit: signersPower ?

certs/certs.go
```diff
@@ -0,0 +1,293 @@
...
+		} else {
+			delta.PowerDelta = newEntry.Power
+			delta.SigningKey = newEntry.PubKey
+		}
+		diff = append(diff, delta)
+	}
+	for _, e := range oldPowerMap {
+		diff = append(diff, PowerTableDelta{
+			ParticipantID: e.ID,
+			PowerDelta:    new(gpbft.StoragePower).Neg(e.Power),
```
anorth, line 223: There's a subtle point here that a power delta that takes the result to zero is to be interpreted as a deletion. Deletion of the signing key isn't explicit here. I think it's worth documenting this interpretation, perhaps with the PowerTableDelta type. Or through the code with a function like `newPowerDeltaDelete(id, e.power)`

certs/certs.go
```diff
@@ -0,0 +1,293 @@
...
+	// only sanity check we need is the above "sort" check.
+	newPowerTable := make(gpbft.PowerEntries, 0, len(prevPowerTable)+len(deltaMap))
+	for _, pe := range prevPowerTable {
+		if diff, ok := deltaMap[pe.ID]; ok {
+			delete(deltaMap, pe.ID)
+
+			if diff.PowerDelta.Sign() != 0 {
+				pe.Power = new(gpbft.StoragePower).Add(diff.PowerDelta, pe.Power)
+				if pe.Power.Sign() == 0 {
+					continue
```
anorth, line 263: The interpretation of a deletion here is subtle, worth a comment.

certs/certs.go
```diff
@@ -0,0 +1,293 @@
...
+				}
+			}
+			if len(diff.SigningKey) > 0 {
+				pe.PubKey = diff.SigningKey
+			}
+		}
+		newPowerTable = append(newPowerTable, pe)
```
anorth, line 270: I think it's worth validating here that the signing key is non-empty. A bad input could lead to an invalid power table.

test/certificate_test.go
```diff
@@ -0,0 +1,387 @@
+package test
```
anorth, line 1: Would this be more canonically located as certs_test in the certs package? I'm curious why it's here.

anorth (commented):
Nice.

The power table delta calculation and application is complex enough to deserve a little more unit testing IMO:
- power gets added to the right entry
- change in key is processed with/without a change in power
- sequences like deletion and re-addition of an entry (requiring key to be re-specified)

certs/certs.go
Stebalien: I can but, IMO, the best solution is just to call `ValidateFinalityCertificates` on the result. It's unclear where to draw the line:

- Validate the signature? Requires the verifier and network name.
- Check the supplemental data? Requires signature validation.
- Check that we have enough power? Requires the power table.
- Validate the bitfield?
- Validate the power delta? Requires the original power table.

So instead I opted to validate information that _won't_ end up in the finality certificate (i.e., make sure the justification is actually for a decide). Well, that and the fact that we're voting on _something_.

Options are:

1. Separate out a "syntactic validation" function from `ValidateFinalityCertificate` in both places.
2. Call `ValidateFinalityCertificates` inside `NewFinalityCertificate` (need to pass a power table, network name, etc.).
3. Tell the user to call `ValidateFinalityCertificates`.

IMO, we should be calling `ValidateFinalityCertificates` regardless which reduces the usefulness of 1. However, I hesitate to go with 2 because it makes it impossible to construct a finality certificate from a justification without a verifier, etc.

certs/certs.go
Stebalien: It's not really the next _base_ power table, it's the power table used to validate the next finality certificate. I can call it "next certificate power table" or "next instance power table"?

certs/certs.go
Stebalien: Fixed in #292.

certs/certs.go
Stebalien: Fixed in #292, #291.

certs/certs.go
Stebalien: Fixed in #291.

certs/certs.go
Stebalien: Fixed in #291.

test/certificate_test.go
Stebalien: No good reason. IIRC, it was because:

1. The certificates used to live in the gpbft package.
2. I didn't want to import the simulation.

Or something like that.

I'll move it in #293 (where I'll also be adding more tests).

PR #267, opened 2024-05-22
FFF-103: remove ID from gpbft.Participant

adlrocha:
Addresses #103. Continuation (and depends on) #259

This PR: 
- Removes the `id` from `gpbft.Participant`.
- Exposes an `ID()` method to access the ID of the current participant through the host interface.

### DIFF
--- a/gpbft/gpbft.go
+++ b/gpbft/gpbft.go
@@ -236,7 +236,7 @@ func (i *instance) ReceiveAlarm() error {
 }
 
 func (i *instance) Describe() string {
-	return fmt.Sprintf("P%d{%d}, round %d, phase %s", i.participant.id, i.instanceID, i.round, i.phase)
+	return fmt.Sprintf("P%d{%d}, round %d, phase %s", i.participant.host.ID(), i.instanceID, i.round, i.phase)
 }
 
 // Processes a single message.
@@ -530,7 +530,7 @@ func (i *instance) beginConverge() {
 			panic("beginConverge called but no justification for proposal")
 		}
 	}
-	_, pubkey := i.powerTable.Get(i.participant.id)
+	_, pubkey := i.powerTable.Get(i.participant.host.ID())
 	ticket, err := MakeTicket(i.beacon, i.instanceID, i.round, pubkey, i.participant.host)
 	if err != nil {
 		i.log("error while creating VRF ticket: %v", err)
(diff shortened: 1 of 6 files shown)


gpbft/gpbft.go
```diff
@@ -552,7 +552,7 @@ func (i *instance) beginConverge() {
 			panic("beginConverge called but no justification for proposal")
 		}
 	}
-	_, pubkey := i.powerTable.Get(i.participant.id)
+	_, pubkey := i.powerTable.Get(i.participant.host.ID())
```
Stebalien, line 555: I believe the idea is to move this into a host-based `broadcast` function, and let the client compute the tickets and signatures.

gpbft/gpbft.go
```diff
@@ -796,7 +796,7 @@ func (i *instance) broadcast(round uint64, step Phase, value ECChain, ticket Tic
 	}
 
 	gmsg := &GMessage{
-		Sender:        i.participant.id,
+		Sender:        i.participant.host.ID(),
```
Stebalien, line 799: Same as above. I believe we want to remove `GMessage.Sender` and let `RequestBroadcast` sign, etc. The host may need to broadcast _multiple_ messages (one per miner).

gpbft/gpbft.go
```diff
@@ -837,13 +837,13 @@ func (i *instance) buildJustification(quorum QuorumResult, round uint64, phase P
 func (i *instance) log(format string, args ...interface{}) {
 	if i.tracer != nil {
 		msg := fmt.Sprintf(format, args...)
-		i.tracer.Log("P%d{%d}: %s (round %d, step %s, proposal %s, value %s)", i.participant.id, i.instanceID, msg,
+		i.tracer.Log("P%d{%d}: %s (round %d, step %s, proposal %s, value %s)", i.participant.host.ID(), i.instanceID, msg,
```
Stebalien, line 840: For the purpose of logging in the simulation, we can pass in a `tracer` that already has an ID attached to it.

Stebalien (commented):
I think the goal is to remove the concept of identity entirely from the instance. I.e., no `ID()` on the host (and the host may have multiple identities.

adlrocha:
Closing in favor of https://github.com/filecoin-project/go-f3/pull/274. This change makes more sense when looking at it also as part of the signature decoupling. As both changes are pretty coupled I decided to use the other PR to reduce the reviewing overhead for the team.

PR #269, opened 2024-05-22
Use deterministic latency in tests to avid intermittent failures

### LINKED ISSUE #262: Interference in parallel tests with TestHonestMultiInstance_Agreement
When enabling more test parallelism, I ran into a failing test for multi-instance agreement only when run in parallel with other tests. When `TestHonestMultiInstance_Agreement` is run in isolation, it always passes. When run in parallel with all other tests, the third iteration always fails with too many rounds. Parallelism was obtained with `t.Parallel()` in both `TestHonestMultiInstance_Agreement` and inside the nested `t.Run()`.

My IDE is running `go test -json -parallel=8 -count=1 ./... -run ./...`, and I replicate the failure with this at CLI. However, removing the `-json` flag causes the tests to pass at CLI 😕.

```
=== RUN   TestHonestMultiInstance_Agreement/async_3
=== PAUSE TestHonestMultiInstance_Agreement/async_3
=== CONT  TestHonestMultiInstance_Agreement/async_3
    multi_instance_test.go:113: 
        	Error Trace:	/Users/me/pl/src/filecoin-project/go-f3/test/multi_instance_test.go:113
        	            				/Users/me/pl/src/filecoin-project/go-f3/test/multi_instance_test.go:46
        	Error:      	Received unexpected error:
        	            	reached maximum number of 10 rounds
        	Test:       	TestHonestMultiInstance_Agreement/async_3
        	Messages:   	P0{593}, round 10, phase COMMIT
        	            	P1{593}, round 11, phase CONVERGE
        	            	P2{593}, round 10, phase COMMIT
--- FAIL: TestHonestMultiInstance_Agreement/async_3 (0.25s)
```

masih:
The intermittent failures always occur for async tests, in form of going beyond the maximum number of rounds allowed in tests (mostly 10). The number of rounds is directly impacted by the order of message delivery to each node, dictated by the latency model.

The Async test options use a latency model that is instantiated when the simulation _options_ are instantiated, not when the simulation itself is instantiated and run. This results in the reuse of the same latency model object across multiple tests.

In go tests may be run in parallel, but the same test never runs in parallel with itself when repeated multiple times via `-count=n` flag. When test are run in parallel using the same latency object across tests becomes the root cause of indeterministic behaviour, where at times a test case can require more rounds than allowed to complete.

The changes here introduce a latency model instantiator, called `latency.Modeler` which encapsulates the responsibility of constructing a latency model object used by a simulation. The simulation is then adopted to take modelers as option, not already instantiated models. This change dramatically reduces the scope of error that was previously occurring where latency models with unclean RNG state were being used by multiple test and assures that the order of message delivery in tests is deterministic regardless of the test execution parallelism.

Additionally, maximum test parallelism is enabled across all table tests and fuzz tests to reduce the total execution runtime.

Fixes #262

### DIFF
--- a/sim/cmd/f3sim/f3sim.go
+++ b/sim/cmd/f3sim/f3sim.go
@@ -31,7 +31,10 @@ func main() {
 		fmt.Printf("Iteration %d: seed=%d, mean=%f\n", i, seed, *latencyMean)
 
 		options := []sim.Option{
-			sim.WithLatencyModel(latency.NewLogNormal(*latencySeed, time.Duration(*latencyMean*float64(time.Second)))),
+			sim.WithLatencyModeler(func() (latency.Model, error) {
+				mean := time.Duration(*latencyMean * float64(time.Second))
+				return latency.NewLogNormal(*latencySeed, mean), nil
+			}),
 			sim.WithECEpochDuration(30 * time.Second),
 			sim.WithECStabilisationDelay(0),
 			sim.AddHonestParticipants(
--- a/sim/latency/latency.go
+++ b/sim/latency/latency.go
@@ -6,6 +6,9 @@ import (
 	"github.com/filecoin-project/go-f3/gpbft"
 )
 
+// Modeler instantiates a new latency Model.
+type Modeler func() (Model, error)
+
 // Model represents a latency model of cross participant communication. The model
 // offers the ability for implementation of varying latency across a simulation,
 // as well as specialised latency across specific participants.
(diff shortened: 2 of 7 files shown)


Stebalien (approved):
LGTM although I'd consider breaking this into two PRs:

1. One to change from a latency model to latency modeler.
2. One to change the tests (parallel, naming, etc.).

masih:
> I'd consider breaking this into two PRs

Sure. Test refactors and extra parallelism increase is now removed from this one. Next one coming up shortly.

PR #273, opened 2024-05-22
Commit to a next power table and additional commitments

### LINKED ISSUE #257: Commit to the next power table in the message payload
#151 makes it difficult to validate finality certificates by introducing a power lookback. Right now, given the previous power table and a finality certificate, I can generate the "next" power table and validate it by checking the `PowerTable` CID in the last finalized epoch. After this change, I won't be able to do that.

One solution is to commit to the power table used to validate the _next_ instance in the message `Payload` itself (and possibly remove it from the tipset list). This would be somewhat annoying _at the moment_ because the "next" power table can change over the course of an instance as we try to find the longest common prefix. However, given a lookback of at least one instance, this value will be consistent for all participants in the instance.

Stebalien:
In addition to committing to a power table in each tipset, agree on:

1. The next power table (used to validate the next instance).
2. Arbitrary commitments (will be used by F3 for snark-friendly power tables, and other fields).

fixes #257

### DIFF
--- a/gpbft/gpbft.go
+++ b/gpbft/gpbft.go
@@ -78,7 +78,7 @@ type Justification struct {
 	Signature []byte
 }
 
-type InstanceData struct {
+type SupplementalData struct {
 	// Merkle-tree of instance-specific commitments. Currently empty but this will eventually
 	// include things like snark-friendly power-table commitments.
 	Commitments [32]byte
@@ -87,7 +87,7 @@ type InstanceData struct {
 	PowerTable CID // []PowerEntry
 }
 
-func (d *InstanceData) Eq(other *InstanceData) bool {
+func (d *SupplementalData) Eq(other *SupplementalData) bool {
 	return d.Commitments == other.Commitments &&
 		bytes.Equal(d.PowerTable, other.PowerTable)
 }
@@ -101,7 +101,7 @@ type Payload struct {
 	// GossiPBFT step name.
 	Step Phase
 	// The common data.
-	Data InstanceData
+	SupplementalData SupplementalData
 	// The value agreed-upon in a single instance.
 	Value ECChain
 }
@@ -110,7 +110,7 @@ func (p *Payload) Eq(other *Payload) bool {
 	return p.Instance == other.Instance &&
 		p.Round == other.Round &&
 		p.Step == other.Step &&
-		p.Data.Eq(&other.Data) &&
+		p.SupplementalData.Eq(&other.SupplementalData) &&
 		p.Value.Eq(other.Value)
 }
 
@@ -130,9 +130,9 @@ func (p *Payload) MarshalForSigning(nn NetworkName) []byte {
 	_ = binary.Write(&buf, binary.BigEndian, p.Step)
 	_ = binary.Write(&buf, binary.BigEndian, p.Round)
 	_ = binary.Write(&buf, binary.BigEndian, p.Instance)
-	_, _ = buf.Write(p.Data.Commitments[:])
+	_, _ = buf.Write(p.SupplementalData.Commitments[:])
 	_, _ = buf.Write(root[:])
-	_, _ = buf.Write(p.Data.PowerTable)
+	_, _ = buf.Write(p.SupplementalData.PowerTable)
 	return buf.Bytes()
 }
 
@@ -158,8 +158,9 @@ type instance struct {
 	// For QUALITY, PREPARE, and COMMIT, this is the latest time (the phase can end sooner).
 	// For CONVERGE, this is the exact time (the timeout solely defines the phase end).
 	phaseTimeout time.Time
-	// Instance data that all participants must agree on.
-	instanceData *InstanceData
+	// Supplemental data that all participants must agree on ahead of time. Messages that
+	// propose supplemental data that differs with our supplemental data will be discarded.
+	supplementalData *SupplementalData
 	// This instance's proposal for the current round. Never bottom.
 	// This is set after the QUALITY phase, and changes only at the end of a full round.
 	proposal ECChain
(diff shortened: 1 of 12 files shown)


Stebalien:
Posting this so we can discuss with respect to #151.

gpbft/participant.go
```diff
@@ -161,10 +161,14 @@ func (p *Participant) beginInstance() error {
 	if err != nil {
 		return fmt.Errorf("failed fetching power table for instance %d: %w", p.nextInstance, err)
 	}
+	// XXX: Need the the power table for the next instance after this one:
+	// GetCommitteeForInstance(p.nextInstance + 1)
+	// XXX: Need the power table from the base tipset: GetCommitteeForEpoch(chain.Base().Epoch)
+	data := new(Data)
```
Stebalien, line 167: If we implement #257, this is where we'll need the power tables.

gpbft/participant.go
```diff
@@ -161,10 +161,13 @@ func (p *Participant) beginInstance() error {
 	if err != nil {
 		return fmt.Errorf("failed fetching power table for instance %d: %w", p.nextInstance, err)
 	}
+	// XXX: Need the the power table for the next instance after this one:
+	// GetCommitteeForInstance(p.nextInstance + 1)
+	data := new(InstanceData)
```
Stebalien, line 166: We have a few options:

1. Actually ask for the next committee.
2. Return the next power table CID as a part of `GetCommitteeForInstance(p.nextInstance)`.
3. Provide some other method.

@Kubuxu preferences?

Stebalien:
The first commit contains all the changes, the last one just renames `Payload.Value` to `Payload.Chain` (which we can drop, land in a different PR, etc. if it becomes a merge issue).

This isn't quite ready for merging yet, but I would like some feedback before I put the finishing touches (tests) on it.

gpbft/participant.go
Stebalien: I've gone with (1) for now.

Kubuxu: `GetCommitteeForInstance` is called in validation flows.

Essentially, there are two things: committees and everything else we agree on, per instance.

Stebalien:
I think we're ready for a final review and merging here unless we want to wait for power table lookback to land first.

gpbft/participant.go
anorth: FWIW I have used the word "committee" inside the participant to mean "the power table plus beacon", i.e. the stuff used to validate a message. I'm happy to change to a different term if we find one, but let's align the terminology across here. I think committee is good enough not to sweat over.

anorth: The base tipset is not part of that, since it's not known at message validation time (validation can be some number of instances ahead due to the lookback). Dropping mismatched base tipsets is a separate tiny bit of logic, but can't be in the main validation flow.

gpbft/gpbft.go
```diff
@@ -86,22 +100,24 @@ type Payload struct {
 	Round uint64
 	// GossiPBFT step name.
 	Step Phase
-	// Chain of tipsets proposed/voted for finalisation.
-	// Always non-empty; the first entry is the base tipset finalised in the previous instance.
-	Value ECChain
+	// The common data.
+	Data InstanceData
```
anorth, line 104: Can we work a bit on this naming? Perhaps AccessoryData or BaseData? It all references things that are already finalised by a prior instance, right?

Then rename InstanceData similarly (it's not data about this instance, it's about prior instances).

gpbft/gpbft.go
```diff
@@ -163,6 +181,8 @@ type instance struct {
 	// Decision state. Collects DECIDE messages until a decision can be made,
 	// independently of protocol phases/rounds.
 	decision *quorumState
+	// Instance data that all participants must agree on.
+	instanceData *InstanceData
```
anorth, line 185: Please move this up, perhaps adjacent to proposal or value, which have similar properties.

gpbft/gpbft.go
```diff
@@ -407,19 +429,24 @@ func (i *instance) validateMessage(msg *GMessage) error {
 	if msg.Vote.Instance != i.instanceID {
 		return xerrors.Errorf("message for wrong instance %d, expected %d", msg.Vote.Instance, i.instanceID)
 	}
+	// Ensure all participants are proposing the same instance data. This data is based on the
+	// _base_, so everyone should agree.
+	if !msg.Vote.Data.Eq(i.instanceData) {
+		return xerrors.Errorf("message for instance %d disagrees on the instance data", msg.Vote.Instance)
+	}
```
anorth, line 436: When you rebase against #264, this will not be part of validation, but can only be checked when receiving a message.

gpbft/participant.go
```diff
@@ -161,10 +161,14 @@ func (p *Participant) beginInstance() error {
 	if err != nil {
 		return fmt.Errorf("failed fetching power table for instance %d: %w", p.nextInstance, err)
 	}
+	data, err := p.host.GetDataForInstance(p.nextInstance)
```
anorth, line 164: I think this could be merged into GetChainForInstance (could be renamed as GetProposalForInstance). It's needed at exactly the same time and for essentially the same reason.

gpbft/participant.go
anorth: Now that I've actually read all the code here, I think this instance data should actually be fetched at the same time as the chain to be proposed. It doesn't align with the committee used for validation.

gpbft/gpbft.go
```diff
@@ -86,22 +100,24 @@ type Payload struct {
 	Round uint64
 	// GossiPBFT step name.
 	Step Phase
-	// Chain of tipsets proposed/voted for finalisation.
-	// Always non-empty; the first entry is the base tipset finalised in the previous instance.
-	Value ECChain
+	// The common data.
+	Data InstanceData
+	// The value agreed-upon in a single instance.
+	Chain ECChain
```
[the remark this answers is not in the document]
Stebalien: Can do. I've kept it as a separate commit.

gpbft/participant.go
Stebalien: Fair enough. That resolves https://github.com/filecoin-project/go-f3/pull/273#discussion_r1614016791.

Stebalien: @Kubuxu SGTY? Also, I think the line between `GetComitteeForInstance` and `GetChainForInstance` is getting a bit blurry.

Stebalien: > Also, I think the line between `GetComitteeForInstance` and `GetProposalForInstance` is getting a bit blurry.

Nvm. I now understand this a bit better (we `GetComitteeForInstance` is what we absolutely _need_ for validation).

gpbft/gpbft.go
Stebalien: By consensus: supplemental data.

PR #276, opened 2024-05-23
Skip ahead to future rounds when justified

### LINKED ISSUE #241: Skip to future rounds when justified by an incoming message
Implement skipping ahead to a future round in the current instance when in receipt of a message for a subsequent round, and: 
- at least one CONVERGE for the higher round
- a weak quorum of PREPARE messages for the higher round

masih:
Implement a rebroadcast protocol improvement where upon receiving sufficient justification a participant jumps ahead to future rounds. Justification from a future rounds is considered sufficient when the participant receives:
 * at least one `CONVERGE` message, and
 * a weak quorum of `PREPARE` messages

for that round.

The changes here introduce a test that should trigger skipping to future rounds once reboradcast is implemented. The test uses a `Deny` adversary where all messages to and from a set of targets are dropped until GST has elapsed. For now, however, the test is skipped.

Fixes #241

### DIFF
--- a/gpbft/gpbft.go
+++ b/gpbft/gpbft.go
@@ -243,8 +243,28 @@ func (i *instance) Receive(msg *GMessage) error {
 }
 
 func (i *instance) ReceiveAlarm() error {
+	phaseBeforeAlarm := i.phase
 	if err := i.tryCurrentPhase(); err != nil {
 		return fmt.Errorf("failed completing protocol phase: %w", err)
+	}
+
+	// Check if the alarm ended QUALITY phase, i.e. transition from QUALITY to PREPARE phase.
+	qualityPhaseEnded := phaseBeforeAlarm == QUALITY_PHASE && i.phase == PREPARE_PHASE
+	if qualityPhaseEnded {
+		// Check if there are any eligible future rounds to which to skip.
+
+		// TODO: Potential optimisations:
+		//        1) Consider preferring higher rounds to jump to first. FIP does not specify
+		//           any constraints here.
+		//        2) Build an incremental index of round candidates as messages arrive, allowing
+		//           for efficient jumping to the target round without scanning all state history.
+		for round, state := range i.rounds {
+			if round > i.round {
+				if chain, justification, skipToRound := i.shouldSkipToRound(round, state); skipToRound {
+					i.skipToRound(round, chain, justification)
+				}
+			}
+		}
 	}
 
 	// A phase may have been successfully completed.
@@ -341,6 +361,11 @@ func (i *instance) receiveOne(msg *GMessage) error {
 		i.log("unexpected message %v", msg)
 	}
 
+	// Check whether the instance should skip ahead to future round.
+	if chain, justification, skip := i.shouldSkipToRound(msg.Vote.Round, round); skip {
+		i.skipToRound(msg.Vote.Round, chain, justification)
+		return nil
+	}
 	// Every COMMIT phase stays open to new messages even after the protocol moves on to
 	// a new round. Late-arriving COMMITS can still (must) cause a local decision, *in that round*.
 	// Try to complete the COMMIT phase for the round specified by the message.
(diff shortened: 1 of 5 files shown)


masih:
@anorth I would love your early feedback on this please to confirm that I am on the right track. Reading the latest FIP I am infering my way around the rebroadcast changes, specifically around `beginNextRound` and how it overlaps with `skipToRound`.

gpbft/gpbft.go
```diff
@@ -567,26 +599,10 @@ func (i *instance) tryQuality() error {
 	return nil
 }
 
-func (i *instance) beginConverge() {
+func (i *instance) beginConvergeWithJustification(justification *Justification) {
```
anorth, line 602: The logical change here seems fine, so I don't think we need to rename the method, just document the parameter.

gpbft/gpbft.go
```diff
@@ -319,6 +319,11 @@ func (i *instance) receiveOne(msg *GMessage) error {
 		if err := round.converged.Receive(msg.Sender, msg.Vote.Value, msg.Ticket, msg.Justification); err != nil {
 			return fmt.Errorf("failed processing CONVERGE message: %w", err)
 		}
+		// Check if we can jump to future rounds.
+		if i.shouldSkipRound(msg, round) {
```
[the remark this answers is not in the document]
anorth: Actually, we should also trigger this from ReceiveAlarm if we continue to disallow skipping while in QUALITY phase (the alarm to exit QUALITY phase would be the trigger). But see Slack - we might be able to drop this.

masih:
@anorth this PR is ready for review. I believe to test that skip rounds functions correctly we need the rest of rebroadcast implementation. Please let me know if I have missed anything or there are tests that you would like to see done as part of this PR.  Otherwise, my vote would be to revisit the testing after rebroadcast is implemented.

gpbft/gpbft.go
```diff
@@ -243,10 +243,30 @@ func (i *instance) Receive(msg *GMessage) error {
...
+		//        2) Build an incremental index of round candidates as messages arrive, allowing
+		//           for efficient jumping to the target round without scanning all state history.
+		for round, state := range i.rounds {
+			if round > i.round {
+				if justification, skipToRound := i.shouldSkipToRound(round, state); skipToRound {
+					i.skipToRound(round, justification)
+				}
+			}
+		}
+	}
```
anorth, line 268: I think we can ditch this after all. See https://github.com/filecoin-project/FIPs/pull/1015

We'll do some other work elsewhere to ensure that all queued messages are processed before we try to jump. But not for this PR.

gpbft/gpbft.go
```diff
@@ -351,6 +376,44 @@ func (i *instance) receiveOne(msg *GMessage) error {
...
+	//
+	// Note that sipping ahead from QUALITY phase does not violate the correctness
+	// proof of gPBFT. The fact that QUALITY always terminates for a participant
+	// means this participant will eventually jump, but it may do it earlier.
+	// However, if a node skips to future rounds without executing the QUALITY phase
+	// it will only help reaching consensus for bottom.
+	//
+	// Future work may consider skipping ahead from QUALITY for faster census in
+	// certain scenarios. For now, this implementation conforms to the FIP.
+	if round <= i.round || i.phase == QUALITY_PHASE || i.phase == DECIDE_PHASE {
```
anorth, line 399: https://github.com/filecoin-project/FIPs/pull/1015

gpbft/gpbft.go
```diff
@@ -351,6 +376,44 @@ func (i *instance) receiveOne(msg *GMessage) error {
...
+		return nil, false
+	}
+	proposal := state.converged.FindMaxTicketProposal(i.powerTable)
+	if proposal.Justification == nil {
```
anorth, line 403: Impossible? Perhaps leave a note if so.

gpbft/gpbft.go
```diff
@@ -351,6 +376,44 @@ func (i *instance) receiveOne(msg *GMessage) error {
...
+		return nil, false
+	}
+	if !state.prepared.ReceivedFromWeakQuorum() {
+		return nil, false
+	}
+	if proposal.Justification.Vote.Step == PREPARE_PHASE {
+		i.log("⚠️ swaying from %s to %s by skip to round %d", &i.proposal, proposal.Chain, i.round)
+		i.candidates = append(i.candidates, proposal.Chain)
+		i.proposal = proposal.Chain
+	}
```
anorth, line 413: I would probably do this in skipToRound so that this function, which sounds like a pure predicate, doesn't have side effects. This is pretty minor in the context though.

gpbft/gpbft.go
```diff
@@ -567,26 +630,11 @@ func (i *instance) tryQuality() error {
 	return nil
 }
 
-func (i *instance) beginConverge() {
+// beginConverge initiates CONVERGE_PHASE justified by the given justification.
+func (i *instance) beginConverge(justification *Justification) {
 	i.phase = CONVERGE_PHASE
-
 	i.phaseTimeout = i.alarmAfterSynchrony()
```
anorth, line 636: Suggestion. Add an assertion that the justification is for `i.round - 1`. I think we can panic if not.

anorth (approved):
This looks good. I think we can drop the QUALITY phase stuff right away though, assuming https://github.com/filecoin-project/FIPs/pull/1015 will land.

gpbft/gpbft.go
masih: Ah I thought [previous suggestion](https://github.com/filecoin-project/go-f3/pull/276#discussion_r1612488201) was to move it to `shouldSkipToRound`. That's where I had it initially 👍

gpbft/gpbft.go
masih: Great; thank you for the FIP update. I think it's worth capturing the context for it in git history in form of a dedicated commit. Since commits get squashed on merge I'll open a separate PR to do this shortly after the merge of this one.

gpbft/gpbft.go
masih: It's possible, because `FindMaxTicketProposal` returns a zero-valued `ConvergeValue` when it doesn't find any such ticket which may have `nil` justification.

I'll leave a comment 👍

anorth: But if `shouldSkipToRound` is true there must be a ticket. This seems more like a panic case.

masih: I am sorry, I don't follow. Here we are determining whether `shouldSkipToRound` is true or not. Right?

anorth: Sorry, yes I was confused.

PR #277, opened 2024-05-24
Adjust Fuzz honest multi-instance to avoid intermittent failures

### LINKED ISSUE #251: Investigate fuzzing process hung or terminated unexpectedly
Specific fuzz tests seem to sporadically fail with error: `fuzzing process hung or terminated unexpectedly: exit status 2`. This could be a bug of sorts in the fuzzer itself as outlined in https://github.com/golang/go/issues/56238. Or possibly a panic during simulation?

See example CI fuzz failure [here](https://github.com/filecoin-project/go-f3/actions/runs/9133502075/job/25117189120).

masih:
`FuzzHonestMultiInstance_AsyncAgreement` intermittently fails on CI, most likely due to taking too long to complete a test.

To avoid intermittent failures:
* change the honest multi instance tests to cover incremental network sizes, up to 4
* reduce the length of randomly generated EC chains at each instance as this should not affect the test quality for what it is testing. But it should make it run faster due to less GC.

Additionally, add static corpus that failed locally after extended fuzz time to the fuzz test.

Fixes #251

### DIFF
--- a/test/multi_instance_test.go
+++ b/test/multi_instance_test.go
@@ -86,10 +86,12 @@ func FuzzHonestMultiInstance_SyncAgreement(f *testing.F) {
 
 func FuzzHonestMultiInstance_AsyncAgreement(f *testing.F) {
 	const (
-		instanceCount = 4000
-		honestCount   = 5
+		instanceCount = 5000
+		honestCount   = 4
 	)
 	f.Add(-7)
+	f.Add(31)
+	f.Add(33)
 	f.Fuzz(func(t *testing.T, seed int) {
 		multiAgreementTest(t, seed, honestCount, instanceCount, maxRounds*2, asyncOptions(seed)...)
 	})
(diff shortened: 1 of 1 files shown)


masih:
I have fuzzed the changes here in CI, specifically for  at fuzz time of 20 minutes and it looks good, see [here](https://github.com/filecoin-project/go-f3/actions/runs/9229817633/job/25396790317?pr=277). 

That gives me confidence that the change here would resolve the issue without affecting coverage.

Kubuxu:
Another option would be to input network size to the fuzzing function (call `t.Skip` if the network size is out of the range).

Kubuxu (approved):

anorth:
We could probably skip network sizes 1 and 2, and maybe 3. 4 is much more useful, and anything a bit larger would be handy too. But I understand the duration limitations. A quick look at optimising could be worthwhile.

PR #282, opened 2024-05-28
Reduce GOGC while fuzzing

Kubuxu:
I've run out of memory when running 60s fuzzing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ SHELL := /usr/bin/env bash
 
 all: generate test fuzz lint
 
-test: GOGC ?= 100 # Reduce GC frequency during testing, default to 1000 if unset.
+test: GOGC ?= 1000 # Reduce GC frequency during testing, default to 1000 if unset.
 test:
 	GOGC=$(GOGC) go test $(GOTEST_ARGS) ./...
 .PHONY: test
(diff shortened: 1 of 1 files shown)


Kubuxu:
> Fuzz failure on CI is interesting. Can you update your branch and push just to check for false negatives please? things might have been fixed already on main.

I think we are right on the edge of performance with these fuzzy tests which might cause the failure.

Makefile
```diff
@@ -13,6 +13,7 @@ test/cover: GOTEST_ARGS=-coverprofile=coverage.txt -covermode=atomic -coverpkg=.
 .PHONY: test/cover
 
 fuzz: FUZZTIME ?= 10s # The duration to run fuzz testing, default to 10s if unset.
+fuzz: GOGC ?= 400 # Reduce GC frequency during testing, default to 200 if unset.
```
masih, line 16:
```suggestion
fuzz: GOGC ?= 400 # Reduce GC frequency during testing, default to 400 if unset.
```

Kubuxu:
I tested with GOGC=100 and fuzzing still fails.

masih (approved):

Kubuxu:
```
--- FAIL: FuzzHonestMultiInstance_SyncAgreement (15.29s)
    fuzzing process hung or terminated unexpectedly: exit status 2
    Failing input written to testdata/fuzz/FuzzHonestMultiInstance_SyncAgreement/1c9eda1b22bf9e69
    To re-run:
    go test -run=FuzzHonestMultiInstance_SyncAgreement/1c9eda1b22bf9e69
```

PR #283, opened 2024-05-28
Reduce GC during fuzz tests

masih:
Reduce GC by reducing the size of EC chain generated by the fuzz tests. The size of EC chain makes no difference to the coverage.

In one case reduce the honest count from 5 to 4 in order to avoid intermittent failures due to fuzz tests taking too long.


masih:
OK Fuzzed the changes here on [CI for 2 minutes per test ](https://github.com/filecoin-project/go-f3/actions/runs/9272597042/job/25510883428)and it is looking good.

Kubuxu (approved):

PR #286, opened 2024-05-28
Make message validation safe for concurrent use.

anorth:
I'm interested in any suggestions of how to test this. I looked into the existing participant tests, but it seemed unlikely the mock is safe to use concurrently and difficult to test this without that. We could manually construct a harness. Is that worthwhile?

See #151

### DIFF
--- a/gpbft/participant.go
+++ b/gpbft/participant.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"sort"
+	"sync"
 )
 
 // An F3 participant runs repeated instances of Granite to finalise longer chains.
@@ -13,10 +14,16 @@ type Participant struct {
 
 	// Instance identifier for the current (or, if none, next to start) GPBFT instance.
 	currentInstance uint64
-	// Current Granite instance.
-	gpbft *instance
 	// Cache of committees for the current or future instances.
 	committees map[uint64]*committee
+	// Mutex protecting currentInstance and committees cache for concurrent validation.
+	// Note that not every access need be protected:
+	// - writes to currentInstance (which is read in validation)
+	// - reads from or writes to committees (which are written during validation)
+	mutex sync.Mutex
+
+	// Current Granite instance.
+	gpbft *instance
 	// Messages queued for future instances.
 	mqueue *messageQueue
 	// The output from the last terminated Granite instance.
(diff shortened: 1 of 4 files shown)


gpbft/participant.go
```diff
@@ -202,36 +201,52 @@ func (p *Participant) beginInstance() error {
...
 
+func (p *Participant) finishCurrentInstance() {
+	p.mutex.Lock()
+	defer p.mutex.Unlock()
+
+	p.finalised = p.gpbft.terminationValue
+	p.terminatedDuringRound = p.gpbft.round
+	delete(p.committees, p.gpbft.instanceID)
+	p.gpbft = nil
+	p.currentInstance++
```
Stebalien, line 247: The only things we actually _need_ to lock are `p.committees` and `p.currentInstance`, right? I'd consider putting everything else outside of the lock for clarity.

gpbft/participant.go
```diff
@@ -14,10 +15,16 @@ type Participant struct {
...
 	currentInstance uint64
-	// Current Granite instance.
-	gpbft *instance
 	// Cache of committees for the current or future instances.
 	committees map[uint64]*committee
+	// Mutex protecting currentInstance and committees cache for concurrent validation.
+	// Note that not every access need be protected:
+	// - writes to currentInstance (which is read in validation)
+	// - reads from or writes to committees (which are written during validation)
+	mutex sync.Mutex
```
Stebalien, line 24: nit: by convention, put this above the protected fields ("mutex hat").

gpbft/participant.go
```diff
@@ -14,10 +15,16 @@ type Participant struct {
 
 	// Instance identifier for the current (or, if none, next to start) GPBFT instance.
 	currentInstance uint64
-	// Current Granite instance.
-	gpbft *instance
 	// Cache of committees for the current or future instances.
 	committees map[uint64]*committee
+	// Mutex protecting currentInstance and committees cache for concurrent validation.
+	// Note that not every access need be protected:
+	// - writes to currentInstance (which is read in validation)
```
Stebalien, line 22: nit: and reads of `currentInstance` in validation.

Stebalien (approved):

Stebalien:
> I'm interested in any suggestions of how to test this. I looked into the existing participant tests, but it seemed unlikely the mock is safe to use concurrently and difficult to test this without that. We could manually construct a harness. Is that worthwhile?

Given that you're adding a lock around `host.GetCommitteeForInstance`, you might actually be able to use the mock. From what I can tell, the mock isn't order dependent so you should be able to:

1. Assert that `GetCommitteeForInstance` is called at most once per instance (i.e., the cache works) via the mock.
2. Repeatedly validate messages with different instances in multiple separate goroutines while the a mock test is running.
3. Ideally: Do all this with the race detector running.

masih (approved):
No blockers. I recommend reducing the scope of locking in `finishCurrentInstance` as Steven suggested.

Yes, I too think we should be able to do the mocks.

Re race detection, we should just run the tests on CI with race detector enabled by default. I will get that done.

Kubuxu:
An alternative to this might be to push out a change in the committee to the Host, which can then be used for validation (struct with a Validate method).

anorth:
Thanks for further encouragement on testing, I got it working as @Stebalien described. Confirmed races are detected if the mutex is removed.

PR #287, opened 2024-05-29
Fix race conditions and Run tests with race detection in CI

### LINKED ISSUE #288: Resolve race conditions in code and test with race detector enabled
With the need for parallel validation, make sure that:
* the current implementation does not have race conditions.
* run tests with race detector enabled.

masih:
Fix a race condition in simulation signing package key generation.

Run tests with race detection enabled in a separate CI job to avoid
slowing down the regular tests as they take significantly longer to
finish.

Fixes #288


Stebalien:
SGTM but why is this getting skipped?

Stebalien (approved):

PR #289, opened 2024-05-29
Shortcut timeout in PREPARE when outcome is determined

### LINKED ISSUE #242: Short-cut timeout in PREAPRE, COMMIT when justified by received messages
```[tasklist]
### Tasks
- [x] Skip the timeout in PREPARE after it's impossible to reach a strong quorum in agreement with the local proposal.
- [ ] Skip the timeout in COMMIT after receiving a strong quorum of anything, including bottom.
```

masih:
When sufficient justification is received do not wait for timeouts to expire in PREPARE phase.

Part of #242

### DIFF
--- a/gpbft/gpbft.go
+++ b/gpbft/gpbft.go
@@ -630,7 +630,7 @@ func (i *instance) tryQuality() error {
 // beginConverge initiates CONVERGE_PHASE justified by the given justification.
 func (i *instance) beginConverge(justification *Justification) {
 	if justification.Vote.Round != i.round-1 {
-		// For safety assert that the justification given blongs to the right round
+		// For safety assert that the justification given belongs to the right round.
 		panic("justification for which to begin converge does not belong to expected round")
 	}
 
@@ -698,17 +698,18 @@ func (i *instance) tryPrepare() error {
 	}
 
 	prepared := i.getRound(i.round).prepared
-	// Optimisation: we could advance phase once a strong quorum on our proposal is not possible.
-	foundQuorum := prepared.HasStrongQuorumFor(i.proposal.Key())
+	proposalKey := i.proposal.Key()
+	foundQuorum := prepared.HasStrongQuorumFor(proposalKey)
 	timedOut := atOrAfter(i.participant.host.Time(), i.phaseTimeout) && prepared.ReceivedFromStrongQuorum()
+	quorumNotPossible := !prepared.couldReachStrongQuorumFor(proposalKey)
 
 	if foundQuorum {
 		i.value = i.proposal
-	} else if timedOut {
+	} else if timedOut || quorumNotPossible {
 		i.value = ECChain{}
 	}
 
-	if foundQuorum || timedOut {
+	if foundQuorum || timedOut || quorumNotPossible {
 		i.beginCommit()
 	}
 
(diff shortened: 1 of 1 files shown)


masih:
@anorth  I would love your early feedback on this please 🙏

gpbft/gpbft.go
```diff
@@ -1010,6 +1011,25 @@ func (q *quorumState) HasStrongQuorumFor(key ChainKey) bool {
...
+		// for given chain.
+		return true
+	case !found:
+		// There is strong quorum of senders and no chain support for given chain,
+		// therefore, strong quorum will never be achieved.
+		return false
+	default:
+		threshold := NewStoragePower(0).Div(q.powerTable.Total, NewStoragePower(3))
+		threshold = threshold.Sub(q.sendersTotalPower, threshold)
+		return supportForChain.power.Cmp(threshold) != -1
```
anorth, line 1029: Use the existing `IsStrongQuorum` function.

anorth (commented):
The changes to tryPrepare/Commit look right to me.

These will be hard to cover with sim tests, but we should endeavour to cover them with unit tests of participant/instance #9 (won't affect line coverage, which is too crude for us here).

gpbft/gpbft.go
```diff
@@ -1086,6 +1087,24 @@ func (q *quorumState) HasStrongQuorumFor(key ChainKey) bool {
...
+		// There is no support for given chain. Strong quorum is only possible if the
+		// aggregate power of voted participants dos not exceed ⅔ of total power.
+		return !q.ReceivedFromStrongQuorum()
+	}
+
+	// A strong quorum is only feasible when the total support for the given chain,
+	// combined with the aggregate power of not yet voted participants, exceeds ⅔ of
+	// total power.
+	nonSendersTotalPower := NewStoragePower(0).Sub(q.powerTable.Total, q.sendersTotalPower)
+	strongQuorumThreshold := NewStoragePower(0).Add(nonSendersTotalPower, supportForChain.power)
```
anorth, line 1104: I don't think threshold is the right word here. Maybe `possibleSupportForKey`, or `maximumSupportForKey`?

anorth (commented):
I see you've reverted the change to commit phase. This PR can be good to merge if the title is updated and it does not automatically close #242.

gpbft/gpbft.go
```diff
@@ -1086,6 +1087,24 @@ func (q *quorumState) HasStrongQuorumFor(key ChainKey) bool {
 	return ok && supportForChain.hasStrongQuorum
 }
 
+// mayHaveStrongQuorumFor checks whether the given chain can possibly reach
+// strong quorum.
+func (q *quorumState) mayHaveStrongQuorumFor(key ChainKey) bool {
+	supportForChain, found := q.chainSupport[key]
+	if !found {
+		// There is no support for given chain. Strong quorum is only possible if the
+		// aggregate power of voted participants dos not exceed ⅔ of total power.
```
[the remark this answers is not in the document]
anorth: Edit: ok the zero value has nil bigint pointer, which doesn't work. So explicit zero.

anorth (approved):
I pushed a commit and adjusted the title and description so that I can approve this and you can merge when you are back online.

test/spam_test.go
```diff
@@ -17,7 +17,7 @@ func TestSpamAdversary(t *testing.T) {
 	t.Parallel()
 	const (
 		instanceCount = 2000
-		maxRounds     = 30
+		maxRounds     = 40
```
anorth, line 20: I think this is from your debugging the problems with the commit case. Since you're not touching that here, probably revert this.

PR #291, opened 2024-05-30
Make it possible to apply multiple power diffs at once

Stebalien:
The finality exchange protocol will need to be able to return power tables at arbitrary epochs for catch up. Rather than keeping a power table per epoch (~256KiB or 720MiB per day), I plan to keep one full power table per day and apply power table diffs to compute the power table at any point (we'll probably cache a bunch of _recent_ power tables in-memory as well).

### DIFF
--- a/certs/certs.go
+++ b/certs/certs.go
@@ -113,7 +113,7 @@ func ValidateFinalityCertificates(verifier gpbft.Verifier, network gpbft.Network
 
 		// Now compute the new power table and validate that it matches the power table for
 		// new head.
-		newPowerTable, err = ApplyPowerTableDiff(prevPowerTable, cert.PowerTableDelta)
+		newPowerTable, err = ApplyPowerTableDiffs(prevPowerTable, cert.PowerTableDelta)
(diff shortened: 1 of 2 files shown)


certs/certs.go
```diff
@@ -229,53 +229,56 @@ func MakePowerTableDiff(oldPowerTable, newPowerTable gpbft.PowerEntries) []Power
...
+			pe, ok := powerTableMap[d.ParticipantID]
+			if ok {
+				if d.PowerDelta.Sign() != 0 {
+					pe.Power = new(gpbft.StoragePower).Add(d.PowerDelta, pe.Power)
+				}
+				if len(d.SigningKey) > 0 {
+					pe.PubKey = d.SigningKey
+				}
+			} else if len(d.SigningKey) == 0 {
+				return nil, xerrors.Errorf("new power entry for %d has an empty signing key", d.ParticipantID)
```
Stebalien, line 259: @anorth validating that the key is non-empty.

Kubuxu (approved):

PR #293, opened 2024-05-30
Extra power diff tests & validation

Stebalien:
Addresses feedback from https://github.com/filecoin-project/go-f3/pull/266#pullrequestreview-2089396736

- [x] power gets added to the right entry
- [x] change in key is processed with/without a change in power
- [x] sequences like deletion and re-addition of an entry (requiring key to be re-specified)

Also, add a bunch of extra checks when applying power table diffs to make sure they're "minimal" and well-specified. This doesn't really matter for correctness (we hash the resulting power table anyways) but:

1. It aids in debugging.
2. It ensures we get _minimal_ power table diffs.

For the latter I'd generally just round-trip the diff, but that's more expensive when combining _many_ diffs (which I expect we'll need to do for the finality exchange protocol).


Stebalien:
On top of #291.

Kubuxu (approved):

PR #294, opened 2024-05-31
When starting a new instance, check for skipping rounds just once after processing all queued messages.

anorth:
See #151

### DIFF
--- a/gpbft/api.go
+++ b/gpbft/api.go
@@ -7,10 +7,11 @@ import (
 
 // Sentinel errors for the message validation and reception APIs.
 var (
-	ErrValidationTooOld      = errors.New("message is for prior instance")
-	ErrValidationNoCommittee = errors.New("no committee for instance")
-	ErrValidationInvalid     = errors.New("message invalid")
-	ErrValidationWrongBase   = errors.New("unexpected base chain")
+	ErrValidationTooOld          = errors.New("message is for prior instance")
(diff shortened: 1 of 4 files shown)


Kubuxu (approved):

gpbft/api.go
```diff
@@ -7,10 +7,11 @@ import (
...
 var (
-	ErrValidationTooOld      = errors.New("message is for prior instance")
-	ErrValidationNoCommittee = errors.New("no committee for instance")
-	ErrValidationInvalid     = errors.New("message invalid")
-	ErrValidationWrongBase   = errors.New("unexpected base chain")
+	ErrValidationTooOld          = errors.New("message is for prior instance")
+	ErrValidationNoCommittee     = errors.New("no committee for instance")
+	ErrValidationInvalid         = errors.New("message invalid")
+	ErrValidationWrongBase       = errors.New("unexpected base chain")
+	ErrValidationWrongSupplement = errors.New("unexpected supplemental data")
```
masih, line 14: Thinking out loud/Something to revisit later: I think these errors are numerous enough to justify a parent type. This is to make the API easier to use when a user wants to catch all validation errors instead of typing out every single related sentinel error.

anorth: I agree. I would love you to demonstrate how to do this. I don't think I am experienced enough in Go to make a good first guess.

masih: https://github.com/filecoin-project/go-f3/pull/307

PR #296, opened 2024-05-31
F3-245: API proposal for ReceiveFinalityCertificate

### LINKED ISSUE #245: Provide API for host to force progress from finality certificates
If a node gets really stuck, we expect the host to fetch finality certificates and use those to force a node up to a new instances without passing through the full protocol. We need some API capability for this; it might also overlap with initial bootstrapping.

Since validating a finality certificate depends on previous certificates, probably the host will do the validate then just tell the GBPFT participant where to skip to (and provide the necessary history of power tables for it to validate messages). But other orchestrations might be possible too.

FYI @Kubuxu

adlrocha:
Closes #245

This PR:
- Adds a `SkipToInstance` interface implemented by `Receiver` that can be used to force progress into a specific instance determined by a finality certificate.
- It simplifies `Participant.Start()` to leverage `SkipToInstance`  under the hood to trigger the beginning of new instances. `SkipToInstance` spawns a new instance asynchronously by setting an alarm for `time.Now()` so it is immediately picked up by the host, and the instance is started. `Start()` under other hand called `beginInstance`  asynchronously and had a lot of redundant code with `ReceiveAlarm`. By using `SkipToInstance` in `Start` we deduplicate some code and handle for all cases the beginning of new instances in the same way.

Notes:
- Depending on how finality certificates are validated and handled by the host, and how the instance (and future) power tables are computed and provided to the host, the API simplified a lot and instead of passing a `FinalityInfo`  we may be able to just pass the instance number we want to skip provided by the latest finality certificate.
- We may not need to add `ReceiveFinalityCertificate` as part of the `Receiver` api if we decide we won't use it for testings, it may be a public function from the participant consumed by the host.

## Future Work
- Host integration of `ReceiveFinality`
- Introduce timeout and condition to trigger the pulling of finality certificates.


adlrocha:
@anorth, would really appreciate your feedback here to see if this makes sense to you and to see what are you suggested next steps (maybe to start with the host integration of finality certificates)? Thanks!

gpbft/api.go
```diff
@@ -51,13 +51,26 @@ type MessageReceiver interface {
 	ReceiveAlarm() error
 }
 
+// Supports catching-up with other participants by pulling finality certificates.
+type FinalityReceiver interface {
```
anorth, line 55: I think this name will be confusing, given that this module is the thing that usually produces finality decisions (see DecisionReceiver, which is implemented by the host).

But also, I don't think we need to pass or validate any information here other than the instance to skip to. The participant will then turn around and fetch the power table, beacon etc needed for that instance when it starts. I don't think this method or its documentation should reference finality certificates or anything else about how the host decided to move to this instance: just do what it's told.

I think we can just inline this in Receiver as a method like `SkipToInstance(instance)`. In fact, if the initial instance was a parameter to `Start` rather than a configuration option, we could merge them into one method. We should consider removing `Start()` once finished here.

gpbft/participant.go
```diff
@@ -151,6 +151,49 @@ func (p *Participant) ReceiveAlarm() (err error) {
...
+	// NOTE: this may not be needed and explicitly pass the committee
+	// to this function
+	// as depending on how we handle certificates in the host, it may
+	// already store the new power tables to make them available for gpbft
+	// through the host without additional work.
+	if err := f.Power.Validate(); err != nil {
+		return fmt.Errorf("invalid power table: %w", err)
+	}
+	com := &committee{power: f.Power, beacon: f.Beacon}
+	p.committees[f.Instance] = com
```
anorth, line 187: This will be fetched when the instance starts. Don't write it here (note it would also need mutex protection)

gpbft/participant.go
```diff
@@ -151,6 +151,49 @@ func (p *Participant) ReceiveAlarm() (err error) {
 	return nil
 }
 
+// FinalityInfo is an auxiliary struct to pass relevant info from the finality certificate
+// NOTE: It is not clear to me what information is available in the finality certificate
+// and in what form. The information provided here may change, using it as a placeholder for now.
+type FinalityInfo struct {
+	Instance uint64
+	Power    *PowerTable
+	Beacon   []byte
```
anorth, line 160: I think the instance is all we need. Existing code will fetch the rest when it's needed.

gpbft/participant.go
```diff
@@ -151,6 +151,49 @@ func (p *Participant) ReceiveAlarm() (err error) {
...
+	p.gpbft = nil
+	prevInstance := p.currentInstance
+	// Set the instance from the certificate as the current one.
+	p.currentInstance = f.Instance
+	// clean all messages queued for instances below the one in the finality certificate
+	// and old committees
+	for i := prevInstance + 1; i < p.currentInstance; i++ {
+		delete(p.mqueue.messages, i)
+		delete(p.committees, i)
+	}
```
anorth, line 174: I suggest adding a `nextInstance` parameter to `finishCurrentInstance()` and then calling that method. Update it to clear the queues in the same way you have here.

gpbft/participant.go
```diff
@@ -151,6 +151,49 @@ func (p *Participant) ReceiveAlarm() (err error) {
...
+	if err := f.Power.Validate(); err != nil {
+		return fmt.Errorf("invalid power table: %w", err)
+	}
+	com := &committee{power: f.Power, beacon: f.Beacon}
+	p.committees[f.Instance] = com
+
+	// Set the alarm to begin a new instance immediately.
+	// This will fetch the chain, drain existing messages for that instance,
+	// and start the instance.
+	p.host.SetAlarm(p.host.Time())
```
anorth, line 192: Let's change `Start` to do the same thing so that every instance start is consistently asynchronous.

gpbft/participant_test.go
```diff
@@ -83,10 +83,14 @@ func newParticipantTestSubject(t *testing.T, seed int64, instance uint64) *parti
 	return &subject
 }
 
-func (pt *participantTestSubject) expectBeginInstance() {
+func (pt *participantTestSubject) expectBeginInstance(fromFinality bool) {
```
anorth, line 86: I expect these changes to revert when allowing the participant to fetch the committee as usual.

anorth (commented):
It looks like this won't be too tricky to pull off, thanks. I think we can simplify a bit from this initial proposal.

gpbft/participant.go
adlrocha: Thank you for the feedback, I didn't know to what extent we could assume the committee to be already available in the host. Question solved, thanks!

gpbft/participant.go
adlrocha: Same as above, thanks!

gpbft/participant.go
adlrocha: It took me longer to expected to find a simple way to refactor the tests without breaking them, but I think I found a way. Please let me know if it makes sense. Thanks!

adlrocha:
Thank you for your feedback and the additional context, @anorth. I think I addressed all your comments. Please let me know if there is any outstanding issue or it is good to merge.

gpbft/api.go
```diff
@@ -54,9 +54,11 @@ type MessageReceiver interface {
 
 // Interface which network participants must implement.
 type Receiver interface {
-	// Begins executing the protocol.
+	// Beings the execution of the instance number passed as an argument.
 	// The node will request the canonical chain to propose from the host.
-	Start() error
+	// This function cna be used to skip directly to an instance number,
+	// for instance, when a new valid finality certificate is received
```
masih, line 60:
```suggestion
	// SkipToInstance jumps directly to a given instance when justified by a finality certificate.
```

...or something along those lines. At the very least fix typo `cna` and end with `.`.

gpbft/participant.go
```diff
@@ -67,15 +67,12 @@ func NewParticipant(host Host, o ...Option) (*Participant, error) {
 	}, nil
 }
 
-// Fetches the preferred EC chain for the instance and begins the GPBFT protocol.
+// Start uses SkipToInstance() under the hood to trigger
```
masih, line 70:
```suggestion
// Start uses SkipToInstance under the hood to trigger
```

sim/adversary/absent.go
```diff
@@ -44,6 +44,10 @@ func (*Absent) ReceiveMessage(_ gpbft.ValidatedMessage) error {
 	return nil
 }
 
+func (*Absent) SkipToInstance(_ uint64) error {
```
masih, line 47:
```suggestion
func (*Absent) SkipToInstance(uint64) error {
```

gpbft/api.go
```diff
@@ -54,9 +54,11 @@ type MessageReceiver interface {
 
 // Interface which network participants must implement.
 type Receiver interface {
-	// Begins executing the protocol.
+	// Beings the execution of the instance number passed as an argument.
 	// The node will request the canonical chain to propose from the host.
-	Start() error
```
masih, line 59: Please make sure PR description covers the rationale for this change since it seems unrelated to the purpose of this PR?

sim/adversary/decide.go
```diff
@@ -86,6 +86,10 @@ func (*ImmediateDecide) ReceiveAlarm() error {
 	return nil
 }
 
+func (*ImmediateDecide) SkipToInstance(_ uint64) error {
```
masih, line 89:
```suggestion
func (*ImmediateDecide) SkipToInstance(uint64) error {
```

sim/adversary/deny.go
```diff
@@ -65,3 +65,4 @@ func (*Deny) ValidateMessage(msg *gpbft.GMessage) (gpbft.ValidatedMessage, error
 }
 func (*Deny) ReceiveMessage(_ gpbft.ValidatedMessage) error { return nil }
 func (*Deny) ReceiveAlarm() error                           { return nil }
+func (*Deny) SkipToInstance(_ uint64) error                 { return nil }
```
masih, line 68:
```suggestion
func (*Deny) SkipToInstance(uint64) error                 { return nil }
```

sim/adversary/repeat.go
```diff
@@ -106,4 +106,5 @@ func (r *Repeat) ReceiveMessage(vmsg gpbft.ValidatedMessage) error {
 func (r *Repeat) ID() gpbft.ActorID                                              { return r.id }
 func (r *Repeat) Start() error                                                   { return nil }
 func (r *Repeat) ReceiveAlarm() error                                            { return nil }
+func (r *Repeat) SkipToInstance(_ uint64) error                                  { return nil }
```
masih, line 109:
```suggestion
func (r *Repeat) SkipToInstance(uint64) error                                  { return nil }
```

sim/adversary/spam.go
```diff
@@ -79,4 +79,5 @@ func (s *Spam) spamAtInstance(instance uint64) {
 
 func (s *Spam) ID() gpbft.ActorID                                              { return s.id }
 func (s *Spam) ReceiveAlarm() error                                            { return nil }
+func (s *Spam) SkipToInstance(_ uint64) error                                  { return nil }
```
masih, line 82:
```suggestion
func (s *Spam) SkipToInstance(uint64) error                                  { return nil }
```

sim/adversary/withhold.go
```diff
@@ -122,6 +122,10 @@ func (*WithholdCommit) ReceiveMessage(_ gpbft.ValidatedMessage) error {
 	return nil
 }
 
+func (*WithholdCommit) SkipToInstance(_ uint64) error {
```
masih, line 125:
```suggestion
func (*WithholdCommit) SkipToInstance(uint64) error {
```

gpbft/api.go
```diff
@@ -54,9 +54,11 @@ type MessageReceiver interface {
...
+	// This function cna be used to skip directly to an instance number,
+	// for instance, when a new valid finality certificate is received
+	SkipToInstance(uint64) error
```
masih, line 61: Two questions:

* Can you please remind me where the finality certificate gets fed in that justifies the jump? My first thought looking at this API was that it should take the justification for skipping ahead as an argument but I am conscious that I might be missing something.

* What happens if there is no justification for the jump? What errors should the user expect? We probably want to document the behaviour.

adlrocha: > Can you please remind me where the finality certificate gets fed in that justifies the jump? My first thought looking at this API was that it should take the justification for skipping ahead as an argument but I am conscious that I might be missing something.

Initially, I thought so too, but gpbft participants have no knowledge of certificates, signatures, when the protocol is stuck or how to compute the corresponding power tables. Only the host knows all of this. What gpbft needs to expose is an interface that the host can use to skip to the instance once a certificate has been validated, and the new power tables have been stored in the host. I guess the next step is to wire this into the host and the certificate validation.

> What happens if there is no justification for the jump? What errors should the user expect? We probably want to document the behaviour.

Thiis is a good point. This function can't return an error anymore, same as `Start()`. With this new approach errors triggered will be handled once the instance has begun. I will push the corresponding update in my next commit.

gpbft/api.go
adlrocha: I elaborated a bit more the description. Let me know if it is clear now. Thanks!

gpbft/api.go
```diff
@@ -54,9 +54,10 @@ type MessageReceiver interface {
 
 // Interface which network participants must implement.
 type Receiver interface {
-	// Begins executing the protocol.
-	// The node will request the canonical chain to propose from the host.
-	Start() error
+	// SkipToInstance jumps directly to a given instance.
+	// This can be triggered by the reception of a valid finality certificate, or
+	// whenever a new instance for a participant want to be started.
+	SkipToInstance(uint64)
```
anorth, line 60: I didn't mean for you to actually remove Start from the interface here, just that we should consider it after this change.

If we're going to remove it, we should do so more thoroughly by removing it from Participant and all the adversary participants, and probably removing the `initialInstance` config parameter. Rather than do that now, let's put it back (still async) and follow up to remove it.

anorth (commented):
Thanks. Looks like this will work well, nearly there.

gpbft/participant_test.go
```diff
@@ -89,6 +89,8 @@ func newParticipantTestSubject(t *testing.T, seed int64, instance uint64) *parti
 func (pt *participantTestSubject) expectBeginInstance() {
 	// Prepare the test host.
 	pt.host.On("GetProposalForInstance", pt.instance).Return(pt.supplementalData, pt.canonicalChain, nil)
+	// if we begin instance from finality, the committee is already available for the instance
+	// and there is no call to GetCommitteeForInstance
```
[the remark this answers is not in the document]
adlrocha: This is a leftover from my initial approach where the committee was being manually updated when skipping the instance. Thank you for flagging.

gpbft/participant.go
```diff
@@ -67,15 +67,12 @@ func NewParticipant(host Host, o ...Option) (*Participant, error) {
...
-			err = &PanicError{Err: r}
-		}
-	}()
-
-	return p.beginInstance()
+// Start uses SkipToInstance under the hood to trigger
+// the start of the first instance.
+// This way we handle in an homogeneous way the start of
+// new instances
+func (p *Participant) Start() {
```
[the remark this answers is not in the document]
adlrocha: Restored. Would you mind elaborating a bit on why this is needed? I think I may have misunderstood the purpose of that code. I was assuming it would be handled by the alarm handler. Thanks!

gpbft/participant.go
```diff
@@ -151,6 +148,18 @@ func (p *Participant) ReceiveAlarm() (err error) {
 	return nil
 }
 
+// Triggers the start to the instance defined as an argument.
+func (p *Participant) SkipToInstance(i uint64) {
```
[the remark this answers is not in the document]
adlrocha: I was trying to add an error that covered the error handling here and I don't think  it is possible for ` SkipToInstance`  to ever return an error (unless I am missing something). Am I missing something? I am wondering if I should remove the error and panic handling here.

adlrocha:
@masih, @anorth, would really appreciate a final :heavy_check_mark: to merge this one. It would be great if I can avoid  having to rebase a few times before merging as we may progress :) Thanks!

masih (approved):
LGTM 👍

anorth (approved):

PR #299, opened 2024-06-03
CertStore: Query for the latest certificate instead of recording it

Stebalien:
I'd rather not write every certificate twice, especially as many datastores don't handle GC well. Luckily, with the datastores we care about (badger and leveldb), query orders optimize well so this query will always return the "latest" certificate.

Alternatively, I can just record the latest instance (and we can do two lookups on start). But I figured I'd collect some feedback on this approach first.


Stebalien:
Going to do this as a part of my certstore changes as it actually simplifies things a bit.

Stebalien:
Nevermind, we can still consider this separately.

Stebalien:
Eh, not worth it.

PR #300, opened 2024-06-03
Avoid loading round state for dropped messages

### LINKED ISSUE #208: `instance.roundState(round)` is a (minor) DoS vector
IIRC, the plan to avoid storing arbitrary messages was to:

1. Store at most one "quality" value from every miner.
2. ignore all messages from future rounds that lack justification from a prior round (again, at most once per miner).

However, `roundState` implicitly allocates a round object before we even look at the message. It's possible for a miner to repeatedly send the _same_ quality message but with different rounds, forcing us to allocate a bunch of round states.

This will _probably_ get fixed when we implement the above but... it's a subtle implicit allocation that's easily overlooked, so I wanted to write this down so we don't forget.

anorth:
Closes #208

### DIFF
--- a/gpbft/gpbft.go
+++ b/gpbft/gpbft.go
@@ -245,10 +245,15 @@ func (i *instance) Receive(msg *GMessage) error {
 	if i.terminated() {
 		return ErrReceivedAfterTermination
 	}
-	if err := i.receiveOne(msg); err != nil {
+	stateChanged, err := i.receiveOne(msg)
+	if err != nil {
 		return err
 	}
-	i.postReceive(msg.Vote.Round)
+	if stateChanged {
+		// Further process the message's round only if it may have had an effect.
+		// This avoids loading state for dropped messages (including spam).
+		i.postReceive(msg.Vote.Round)
+	}
 	return nil
 }
 
@@ -262,7 +267,8 @@ func (i *instance) ReceiveMany(msgs []*GMessage) error {
 	// Received each message and remember which rounds were received.
 	roundsReceived := map[uint64]struct{}{}
 	for _, msg := range msgs {
-		if err := i.receiveOne(msg); err != nil {
+		stateChanged, err := i.receiveOne(msg)
+		if err != nil {
 			if errors.Is(err, ErrValidationWrongBase) || errors.Is(err, ErrValidationWrongSupplement) {
 				// Drop late-binding validation errors.
 				i.log("dropping invalid message: %s", err)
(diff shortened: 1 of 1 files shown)


gpbft/gpbft.go
```diff
@@ -262,7 +267,7 @@ func (i *instance) ReceiveMany(msgs []*GMessage) error {
 	// Received each message and remember which rounds were received.
 	roundsReceived := map[uint64]struct{}{}
 	for _, msg := range msgs {
-		if err := i.receiveOne(msg); err != nil {
+		if _, err := i.receiveOne(msg); err != nil {
```
anorth, line 270: This path is reached only once per instance, so I don't think it's worth the check here. But I can accumulate a stateChanged here too if requested.

gpbft/gpbft.go
```diff
@@ -294,90 +299,90 @@ func (i *instance) Describe() string {
...
+			msgRound.committed.ReceiveJustification(msg.Vote.Value, msg.Justification)
 		}
 	case DECIDE_PHASE:
 		i.decision.Receive(msg.Sender, msg.Vote.Value, msg.Signature)
 		if i.phase != DECIDE_PHASE {
 			i.skipToDecide(msg.Vote.Value, msg.Justification)
 		}
-		if err := i.tryDecide(); err != nil {
-			return fmt.Errorf("failed to decide: %w", err)
-		}
```
anorth, line 361: This is redundant after implementation of skipToDecide, immediately above. `tryDecide()` will be invoked by `tryCurrentPhase()`  at the end.

gpbft/gpbft.go
```diff
@@ -391,22 +396,21 @@ func (i *instance) postReceive(roundsReceived ...uint64) {
...
 func (i *instance) shouldSkipToRound(round uint64, state *roundState) (ECChain, *Justification, bool) {
-
 	// Check if the given round is ahead of current round and this instance is not in
 	// DECIDE phase.
 	if round <= i.round || i.phase == DECIDE_PHASE {
 		return nil, nil, false
 	}
+	if !state.prepared.ReceivedFromWeakQuorum() {
+		return nil, nil, false
+	}
```
anorth, line 406: Moved this cheaper and harder-to-meet check to happen first

gpbft/gpbft.go
Stebalien: We're limited to `maxLookaheadRounds * powerLookback` anyways (except for justified messages). So I'd say this is fine.

Stebalien: Although... it should be trivial to just not add this round to `roundsReceived`, right?

Stebalien (approved):

gpbft/gpbft.go
anorth: Yes,done

PR #301, opened 2024-06-04
Add power tables to the Certificate Store

Stebalien:
This commit adds power tables to the certificate store.

1. It stores a full power table every `60*24` instances (twice a day on Filecoin).
2. From there, it applies deltas from finality certificates to recover any requested power table.

### DIFF
--- a/certstore/certstore.go
+++ b/certstore/certstore.go
@@ -3,61 +3,183 @@ package certstore
 import (
 	"bytes"
 	"context"
+	"encoding/binary"
 	"errors"
 	"fmt"
 	"math"
 	"sync"
 
 	"github.com/filecoin-project/go-f3/certs"
-	"github.com/ipfs/go-datastore"
-	"github.com/ipfs/go-datastore/namespace"
+	"github.com/filecoin-project/go-f3/gpbft"
 
 	"github.com/Kubuxu/go-broadcast"
+	"github.com/ipfs/go-datastore"
+	"github.com/ipfs/go-datastore/namespace"
 	"golang.org/x/xerrors"
 )
 
 var ErrCertNotFound = errors.New("certificate not found")
 
+const defaultPowerTableFrequency = 60 * 24 // expected twice a day for Filecoin
+
+var (
+	certStoreLatestKey = datastore.NewKey("/latestCert")
+	certStoreFirstKey  = datastore.NewKey("/firstInstance")
+)
+
 // Store is responsible for storing and relaying information about new finality certificates
 type Store struct {
-	writeLk  sync.Mutex
-	ds       datastore.Datastore
-	busCerts broadcast.Channel[*certs.FinalityCertificate]
+	writeLk             sync.Mutex
+	ds                  datastore.Datastore
+	busCerts            broadcast.Channel[*certs.FinalityCertificate]
+	firstInstance       uint64
+	powerTableFrequency uint64
+
+	latestPowerTable gpbft.PowerEntries
 }
 
-// NewStore creates a certstore
-// The passed Datastore has to be thread safe.
-func NewStore(ctx context.Context, ds datastore.Datastore) (*Store, error) {
+// Internal helper function to open a certificate store that may or may not have been created.
+func open(ctx context.Context, ds datastore.Datastore) (*Store, error) {
 	cs := &Store{
-		ds: namespace.Wrap(ds, datastore.NewKey("/certstore")),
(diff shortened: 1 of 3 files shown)


certstore/certstore.go
```diff
@@ -121,20 +174,104 @@ func (cs *Store) GetRange(ctx context.Context, start uint64, end uint64) ([]cert
...
+func (*Store) keyForPowerTable(i uint64) datastore.Key {
+	return datastore.NewKey(fmt.Sprintf("/power/%016X", i))
+}
+
 // Put saves a certificate in a store and notifies listeners.
-// It errors if adding a cert would create a gap
+// It errors if adding a cert would create a gap.
 func (cs *Store) Put(ctx context.Context, cert *certs.FinalityCertificate) error {
-	key := cs.keyForInstance(cert.GPBFTInstance)
+	if cert.GPBFTInstance < cs.firstInstance {
```
Stebalien, line 264: I'm only allowing finality certificates between the first instance (first power table) and the next instance, inclusive. In practice, this means we'll write strictly incrementing finality certificates.

certstore/certstore.go
```diff
@@ -121,20 +174,104 @@ func (cs *Store) GetRange(ctx context.Context, start uint64, end uint64) ([]cert
...
+		return xerrors.Errorf("certificate store only stores certificates on or after instance %d", cs.firstInstance)
+	}
+	cs.writeLk.Lock()
+	defer cs.writeLk.Unlock()
```
Stebalien, line 268: I've moved the lock up to the top because we really don't care about contention on write (there should only ever be one writer).

certstore/certstore.go
```diff
@@ -143,21 +280,13 @@ func (cs *Store) Put(ctx context.Context, cert *certs.FinalityCertificate) error
...
-			cert.GPBFTInstance, cs.Latest().GPBFTInstance)
-	}
-
-	if err := cs.ds.Put(ctx, key, buf.Bytes()); err != nil {
+	if err := cs.ds.Put(ctx, cs.keyForCert(cert.GPBFTInstance), buf.Bytes()); err != nil {
 		return xerrors.Errorf("putting the cert: %w", err)
 	}
-	if err := cs.ds.Put(ctx, certStoreLatestKey, buf.Bytes()); err != nil {
-		return xerrors.Errorf("putting the cert: %w", err)
+	if err := cs.writePointer(ctx, certStoreLatestKey, cert.GPBFTInstance); err != nil {
```
Stebalien, line 289: I now write the instance number of the latest finality certificate, instead of writing the actual certificate (more efficient, matches what I'm doing with the "first instance").

certstore/certstore.go
```diff
@@ -3,61 +3,114 @@ package certstore
...
+
+	cs.busCerts.Publish(latestCert)
 
 	return cs, nil
 }
 
-var certStoreLatestKey = datastore.NewKey("/latestCert")
+// NewStore initializes a new certificate store.
+// The passed Datastore has to be thread safe.
+func NewStore(ctx context.Context, ds datastore.Datastore, firstInstance uint64, initialPowerTable gpbft.PowerEntries) (*Store, error) {
```
Stebalien, line 64: We now need the first instance/power table when initialising a new store.

Stebalien:
I was initially intending to let the user set the initial power table late and, in the process, overcomplicated this. I'm going to switch to eagerly storing the power table to simplify it.

certstore/certstore.go
```diff
@@ -121,44 +246,156 @@ func (cs *Store) GetRange(ctx context.Context, start uint64, end uint64) ([]cert
...
+	}
+	if instance > lastPowerTable {
+		return nil, fmt.Errorf("cannot return future power table for instance %d > %d", instance, lastPowerTable)
+	}
+
+	// We store every `powerTableFrequency` power tables. Find the nearest multiple smaller than
+	// the requested instance.
+	startInstance := instance
+	if offset := instance % cs.powerTableFrequency; offset != 0 {
+		startInstance = max(startInstance-offset, cs.firstInstance)
```
Kubuxu, line 289: IMO, this doesn't have to be in an if, and I had to think about it for a second to check that skipping the `max` is safe.

Stebalien: We still need the max as the first instance may not be aligned. But you're right, we don't need the if.

Stebalien: Fixed.

Kubuxu (approved):

PR #305, opened 2024-06-04
Enforce size limits on TipSet key and power table CID length.

### LINKED ISSUE #303: Enforce size limits on TipSet structure
The TipSet structure has some fields that are unbounded in length, so could be used to spam very large messages and state. Set upper limits on them for message validity.
- Key
- PowerTable CID?

anorth:
Closes #303

### DIFF
--- a/gpbft/chain.go
+++ b/gpbft/chain.go
@@ -47,9 +47,9 @@ type TipSet struct {
 	// The EC epoch (strictly increasing).
 	Epoch int64
 	// The tipset's key (canonically ordered concatenated block-header CIDs).
-	Key TipSetKey `cborgen:"maxlen=38"`
+	Key TipSetKey `cborgen:"maxlen=760"` // 20 * 38B
 	// Blake2b256-32 CID of the CBOR-encoded power table.
-	PowerTable CID // []PowerEntry
+	PowerTable CID `cborgen:"maxlen=38"` // []PowerEntry
 	// Keccak256 root hash of the commitments merkle tree.
 	Commitments [32]byte
 }
--- a/gpbft/chain.go
+++ b/gpbft/chain.go
@@ -47,15 +47,15 @@ type TipSet struct {
 	// The EC epoch (strictly increasing).
 	Epoch int64
 	// The tipset's key (canonically ordered concatenated block-header CIDs).
-	Key TipSetKey
+	Key TipSetKey `cborgen:"maxlen=38"`
 	// Blake2b256-32 CID of the CBOR-encoded power table.
 	PowerTable CID // []PowerEntry
 	// Keccak256 root hash of the commitments merkle tree.
 	Commitments [32]byte
 }
 
-var EMPTY_COMMITMENTS = [32]byte{}
-
+// Validates a tipset.
+// Note the zero value is invalid.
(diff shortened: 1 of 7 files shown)


gpbft/chain.go
```diff
@@ -48,6 +54,27 @@ type TipSet struct {
 	Commitments [32]byte
 }
 
+var EMPTY_COMMITMENTS = [32]byte{}
+
+func (ts *TipSet) Validate() error {
+	if len(ts.Key) == 0 {
+		return errors.New("tipset key must not be empty")
+	}
+	if len(ts.Key) > TIPSET_KEY_MAX_LEN {
```
Stebalien, line 63: I'd consider limiting this as well in cbor-gen as well with a ``﻿`cborgen:"maxlen=38"`` struct tag.

gpbft/chain.go
```diff
@@ -48,6 +54,27 @@ type TipSet struct {
...
+		return errors.New("tipset key too long")
+	}
+	if len(ts.PowerTable) == 0 {
+		return errors.New("power table CID must not be empty")
+	}
+	if len(ts.PowerTable) > CID_MAX_LEN {
+		return errors.New("power table CID too long")
+	}
+	if !bytes.Equal(ts.Commitments[:], EMPTY_COMMITMENTS[:]) {
```
Stebalien, line 72: Doesn't really matter (we'll reject it when we compare it with our own chain) but I guess this is fine.

gpbft/chain.go
Stebalien: Or _only_ limiting it in cbor-gen.

gpbft/chain.go
```diff
@@ -238,6 +265,9 @@ func (c ECChain) Validate() error {
 		if ts.IsZero() {
```
Stebalien, line 265: Coveredby `Validate` below.

Stebalien (approved):
LGTM modulo a few suggestions. Also, we should probably have a few tests with bad CIDs/tipsets.

Kubuxu (approved):
SGWM (although @Stebalien suggestions are good as well)

gpbft/chain.go
anorth: Added a struct tag too.

gpbft/chain.go
anorth: Good point, I removed this check so it doesn't trip up later. 

This means that a message with non-empty commitments will be silently ignored after being received rather than being marked as invalid and not propagating.

gpbft/chain.go
```diff
@@ -41,13 +47,31 @@ type TipSet struct {
 	// The EC epoch (strictly increasing).
 	Epoch int64
 	// The tipset's key (canonically ordered concatenated block-header CIDs).
-	Key TipSetKey
+	Key TipSetKey `cborgen:"maxlen=38"`
```
Stebalien, line 50: 20 * 38. The power table CID has a max length of 38.

Stebalien: We should apply the same limit in the `SupplementalData` as well.

Stebalien (approved):

Stebalien:
(needs a make gen)

PR #307, opened 2024-06-05
Group validation errors as top level `ValidationError`

masih:
The validation errors are numerous enough to grant the presence of some higher level grouping for convenient API usage. This allows the users to infer whether an error was caused by validation without having to list every sentinel error associated with validation errors.

Relates to: https://github.com/filecoin-project/go-f3/pull/294#discussion_r1626602926

### DIFF
--- a/gpbft/api.go
+++ b/gpbft/api.go
@@ -1,21 +1,7 @@
 package gpbft
 
 import (
-	"errors"
 	"time"
-)
-
-// Sentinel errors for the message validation and reception APIs.
-var (
-	ErrValidationTooOld          = errors.New("message is for prior instance")
-	ErrValidationNoCommittee     = errors.New("no committee for instance")
-	ErrValidationInvalid         = errors.New("message invalid")
-	ErrValidationWrongBase       = errors.New("unexpected base chain")
-	ErrValidationWrongSupplement = errors.New("unexpected supplemental data")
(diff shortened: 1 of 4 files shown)


anorth (commented):
Could you please update the use of `errors.Is` is gpbft to now check for the supertype?

masih:
> Could you please update the use of `errors.Is` is gpbft to now check for the supertype?

As in silently drop _any_ invalid message instead of the ones with [wrong base or wrong supplement](https://github.com/filecoin-project/go-f3/blob/2affad9594bb49fb8c7d3fc4283095ff61b530f2/gpbft/gpbft.go#L272)? 

If we want to only drop messages in specific cases then the use of `erros.Is` is correct.

anorth:
> > Could you please update the use of `errors.Is` is gpbft to now check for the supertype?
> 
> As in silently drop _any_ invalid message instead of the ones with [wrong base or wrong supplement](https://github.com/filecoin-project/go-f3/blob/2affad9594bb49fb8c7d3fc4283095ff61b530f2/gpbft/gpbft.go#L272)?

Yes. Those two errors were written in because they are the only two that are ever thrown by the code being called. But now that we have a way to express dropping any validation error, that was the intent all along.

anorth (approved):

PR #317, opened 2024-06-07
Fix race condition in honest tests

masih:
Avoid modifying test case options during parallel test execution.

### DIFF
--- a/test/honest_test.go
+++ b/test/honest_test.go
@@ -71,14 +71,16 @@ func TestHonest_Agreement(t *testing.T) {
 			name := fmt.Sprintf("%s %d", test.name, participantCount)
 			t.Run(name, func(t *testing.T) {
 				t.Parallel()
-				test.options = append(test.options,
+				var opts []sim.Option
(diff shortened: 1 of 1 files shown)


test/honest_test.go
```diff
@@ -71,14 +71,14 @@ func TestHonest_Agreement(t *testing.T) {
 			name := fmt.Sprintf("%s %d", test.name, participantCount)
 			t.Run(name, func(t *testing.T) {
 				t.Parallel()
-				test.options = append(test.options,
+				opts := append(test.options,
```
Kubuxu, line 74: This append will is not guaranteed to make a copy. It could append onto the existing capacity in `test.options`.

Probably do:
```
opts := append([]sim.Option{}, test.options...)
opts = append(opts, ...)
```

Kubuxu (changes requested):
I think it works right now but it is not guaranteed to

Kubuxu (approved):
append on an empty array expression is how I usually saw it done but LGTM

PR #318, opened 2024-06-07
Document and adjust alarm delivery precedence over network messages

### LINKED ISSUE #316: Receiving alarms must take precedence over receiving broadcasted messages

masih:
Document the implicit assumption that the delivery of alarms to a participant must take precedence over broadcast messages.

Adjust simulation message queue to respect the alarm ordering.

Fixes #316

### DIFF
--- a/gpbft/api.go
+++ b/gpbft/api.go
@@ -48,6 +48,11 @@ type MessageReceiver interface {
 	// - both ErrReceivedInternalError and a cause if there was an internal error processing the message
 	// This method is not safe for concurrent use.
 	ReceiveMessage(msg ValidatedMessage) error
+	// ReceiveAlarm signals the trigger of the alarm set by Clock.SetAlarm. Note that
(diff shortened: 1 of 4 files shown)


Kubuxu:
This is hard or almost impossible to guarantee in an environment where signing is async and can fail. I will open an issue about it.

Kubuxu:
https://github.com/filecoin-project/go-f3/issues/319

Kubuxu (approved):
Codewise and resolving the immediate issue, it seems good to me, but directionally I think we want something else.

masih:
> This is hard or almost impossible to guarantee in an environment where signing is async and can fail. I will open an issue about it.

Even for alarms? Alarms have no signing. right?

Kubuxu:
Ahh, wait, I might have miss-read it, I assumed broadcast have precedence, I can for sure make the alarms take precedence.

On the other hand, in a real system, there is a possibility of a message arriving just before the alarm but IDK if it will be an issue.

PR #320, opened 2024-06-07
Add validation to pubsub

Kubuxu:
Remove the selfMessageQueue as well as it won't get used

PR #322, opened 2024-06-07
Save decisions in certstore

Kubuxu:


Kubuxu:
Based on #320

host.go
```diff
@@ -167,10 +161,38 @@ func (h *gpbftHost) SetAlarm(at time.Time) {
...
+	if err != nil {
+		return xerrors.Errorf("getting commitee for next instance %d: %w", instance+1, err)
+	}
+	powerDiff := certs.MakePowerTableDiff(current.Entries, next.Entries)
+
+	cert, err := certs.NewFinalityCertificate(powerDiff, decision)
+	if err != nil {
+		return xerrors.Errorf("forming certificate out of decision: %w", err)
+	}
+	err = h.client.certstore.Put(h.runningCtx, cert)
```
Stebalien, line 189: IMO, we should spend the extra time and validate this certificate. We can either do that inside the certstore, or here.

Stebalien (approved):

host.go
```diff
@@ -178,10 +185,43 @@ func (h *gpbftHost) SetAlarm(at time.Time) {
...
+	if err != nil {
+		return xerrors.Errorf("getting commitee for next instance %d: %w", instance+1, err)
+	}
+	powerDiff := certs.MakePowerTableDiff(current.Entries, next.Entries)
+
+	cert, err := certs.NewFinalityCertificate(powerDiff, decision)
+	if err != nil {
+		return xerrors.Errorf("forming certificate out of decision: %w", err)
+	}
+	_, _, _, err = certs.ValidateFinalityCertificates(h, h.NetworkName(), current.Entries, decision.Vote.Instance, cert.ECChain.Base())
```
Stebalien, line 213: note: you don't need to pass the base. If you pass nil, it'll just ignore it.

Stebalien (approved):

PR #324, opened 2024-06-10
Simulate skip to future instances for lagging participants

masih:
The `SkipToInstance` API is intended to be called by an honest integrator whenever a finality certificate is observed from a future instance.

As a result `TestDeny_SkipsToFuture`, previously skipped, can now be run when the simulation itself (playing the role of honest integrator) notifies participants of the existence of a future round whenever it is applicable.

The changes here introduce a new participant receiver to get the current instance, and notified of future instance when it is behind the latest completed one.

### DIFF
--- a/gpbft/participant.go
+++ b/gpbft/participant.go
@@ -89,6 +89,12 @@ func (p *Participant) CurrentRound() uint64 {
 	return p.gpbft.round
 }
 
+func (p *Participant) CurrentInstance() uint64 {
+	p.mutex.Lock()
+	defer p.mutex.Unlock()
+	return p.currentInstance
+}
+
 // Validates a message
 func (p *Participant) ValidateMessage(msg *GMessage) (valid ValidatedMessage, err error) {
 	defer func() {
--- a/sim/sim.go
+++ b/sim/sim.go
@@ -94,6 +94,18 @@ func (s *Simulation) Run(instanceCount uint64, maxRounds uint64) error {
 			// The last incomplete instance is used for testing assertions.
 			currentInstance = s.ec.BeginInstance(*decidedChain, pt)
 
+			// Simulate the behaviour of an honest integrator, where upon receiving finality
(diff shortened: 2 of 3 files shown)


gpbft/participant.go
```diff
@@ -89,6 +89,10 @@ func (p *Participant) CurrentRound() uint64 {
 	return p.gpbft.round
 }
 
+func (p *Participant) CurrentInstance() uint64 {
+	return p.currentInstance
```
anorth, line 93: Since this method is externally invoked, we can't control when that is, so I think this needs protection by the mutex (like validation).

sim/sim.go
```diff
@@ -94,6 +94,17 @@ func (s *Simulation) Run(instanceCount uint64, maxRounds uint64) error {
 			// The last incomplete instance is used for testing assertions.
 			currentInstance = s.ec.BeginInstance(*decidedChain, pt)
 
+			// Simulate the behaviour of an honest integrator, where upon receiving finality
+			// certificates from future rounds it signals gpbft to skip ahead.
```
anorth, line 98: This will react much faster than we might expect in reality, but it's probably good enough for exercising the code.

masih: Right 👍 

I have left a TODO in the inner clause to later enhance the simulation options once the finality certificate propagation protocol/mechanism is clearer. Happy to add delay now if you think this is too unrealistic.

sim/sim.go
```diff
@@ -94,6 +94,17 @@ func (s *Simulation) Run(instanceCount uint64, maxRounds uint64) error {
...
+			for _, p := range s.participants {
+				if p.CurrentInstance() < currentInstance.Instance {
+					// TODO: simulate optimistic delivery of finality certificates.
```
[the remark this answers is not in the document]
masih: Great minds. Will do.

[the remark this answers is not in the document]
masih: https://github.com/filecoin-project/go-f3/issues/327

PR #326, opened 2024-06-10
Rebroadcast last round of messages if instance has not progressed

### LINKED ISSUE #243: Rebroadcast last round messages when stuck at step longer than timeout
After some timeout that's longer than the usual step timeout, if stuck in a phase (haven't received from >2/3 of power), rebroadcast messages sent for the current and previous round that will support another node catching up. The simplest is probably to re-send everything since (inclusive) PREPARE from the prior round.

Note the DECIDE step is a special case, where only DECIDE need be rebroadcast.

masih:
When the current instance has not progressed after some time rebroadcast the last round of messages. The rebroadcast time is made configurable using a bounded exponential backoff after phase timeout expires.The rebroadcast timeout is offset by phase timeout when not expired, and by latest rebroadcast time after that.

Once the first rebroadcast is triggered successive rebroadcasts use the `Clock` alarm mechanism to daisy-chain the triggers one after another.

Introduce `Drop` adversary to simulate scenarios where for a given set of target participants messages are dropped based on some configured message loss probability. Simulate tests using the `Drop` adversary and assert that despite stochastic message loss the targeted participant reaches the expected consensus.

Fixes #243

### DIFF
--- a/gpbft/gpbft.go
+++ b/gpbft/gpbft.go
@@ -160,6 +160,21 @@ type instance struct {
 	// For QUALITY, PREPARE, and COMMIT, this is the latest time (the phase can end sooner).
 	// For CONVERGE, this is the exact time (the timeout solely defines the phase end).
 	phaseTimeout time.Time
+	// rebroadcastTimeout is the time at which the current phase should attempt to
+	// rebroadcast messages in order to further its progress.
+	//
+	// See tryRebroadcast.
+	rebroadcastTimeout time.Time
+	// rebroadcastAttempts counts the number of times messages at a round have been
+	// rebroadcasted in order to determine the backoff duration until next rebroadcast.
+	//
+	// See tryRebroadcast.
+	rebroadcastAttempts int
+	// broadcasted selectively stores messages that have been broadcast in case
+	// rebroadcast becomes necessary.
+	//
+	// See: broadcast, tryRebroadcast.
+	broadcasted *broadcastState
 	// Supplemental data that all participants must agree on ahead of time. Messages that
 	// propose supplemental data that differs with our supplemental data will be discarded.
 	supplementalData *SupplementalData
@@ -210,6 +225,7 @@ func newInstance(
 		phase:            INITIAL_PHASE,
 		supplementalData: data,
 		proposal:         input,
+		broadcasted:      newBroadcastState(),
 		value:            ECChain{},
 		candidates:       []ECChain{input.BaseChain()},
 		quality:          newQuorumState(powerTable),
@@ -697,6 +713,7 @@ func (i *instance) beginPrepare(justification *Justification) {
 	// Broadcast preparation of value and wait for everyone to respond.
 	i.phase = PREPARE_PHASE
 	i.phaseTimeout = i.alarmAfterSynchrony()
+
 	i.broadcast(i.round, PREPARE_PHASE, i.value, false, justification)
 }
 
@@ -709,17 +726,22 @@ func (i *instance) tryPrepare() error {
 	prepared := i.getRound(i.round).prepared
 	proposalKey := i.proposal.Key()
 	foundQuorum := prepared.HasStrongQuorumFor(proposalKey)
-	timedOut := atOrAfter(i.participant.host.Time(), i.phaseTimeout) && prepared.ReceivedFromStrongQuorum()
+	timedOut := atOrAfter(i.participant.host.Time(), i.phaseTimeout)
 	quorumNotPossible := !prepared.couldReachStrongQuorumFor(proposalKey)
+	phaseComplete := timedOut && prepared.ReceivedFromStrongQuorum()
 
 	if foundQuorum {
 		i.value = i.proposal
-	} else if timedOut || quorumNotPossible {
+	} else if quorumNotPossible || phaseComplete {
 		i.value = ECChain{}
 	}
 
-	if foundQuorum || timedOut || quorumNotPossible {
+	if foundQuorum || quorumNotPossible || phaseComplete {
 		i.beginCommit()
+	} else if timedOut {
+		if err := i.tryRebroadcast(); err != nil {
+			return fmt.Errorf("failed to rebroadcast at %s step: %w", i.phase, err)
+		}
 	}
 
 	return nil
@@ -751,14 +773,15 @@ func (i *instance) tryCommit(round uint64) error {
 	// A subsequent COMMIT message can cause the node to decide, so there is no check on the current phase.
 	committed := i.getRound(round).committed
 	quorumValue, foundStrongQuorum := committed.FindStrongQuorumValue()
-	timedOut := atOrAfter(i.participant.host.Time(), i.phaseTimeout) && committed.ReceivedFromStrongQuorum()
+	timedOut := atOrAfter(i.participant.host.Time(), i.phaseTimeout)
+	phaseComplete := timedOut && committed.ReceivedFromStrongQuorum()
 
 	if foundStrongQuorum && !quorumValue.IsZero() {
 		// A participant may be forced to decide a value that's not its preferred chain.
 		// The participant isn't influencing that decision against their interest, just accepting it.
 		i.value = quorumValue
 		i.beginDecide(round)
-	} else if i.round == round && i.phase == COMMIT_PHASE && (timedOut || foundStrongQuorum) {
+	} else if i.round == round && i.phase == COMMIT_PHASE && (foundStrongQuorum || phaseComplete) {
 		if foundStrongQuorum {
 			// If there is a strong quorum for bottom, carry forward the existing proposal.
 		} else {
@@ -782,7 +805,12 @@ func (i *instance) tryCommit(round uint64) error {
 			}
 		}
 		i.beginNextRound()
+	} else if timedOut {
+		if err := i.tryRebroadcast(); err != nil {
+			return fmt.Errorf("failed to rebroadcast at %s step: %w", i.phase, err)
+		}
 	}
+
 	return nil
 }
 
@@ -823,6 +851,10 @@ func (i *instance) tryDecide() error {
 			i.terminate(decision)
 		} else {
 			panic("tryDecide with no strong quorum for value")
+		}
+	} else {
+		if err := i.tryRebroadcast(); err != nil {
+			return fmt.Errorf("failed to rebroadcast at %s step: %w", i.phase, err)
 		}
 	}
 
@@ -867,10 +899,6 @@ func (i *instance) beginNextRound() {
 func (i *instance) skipToRound(round uint64, chain ECChain, justification *Justification) {
 	i.log("skipping from round %d to round %d with %s", i.round, round, i.proposal.String())
 	i.round = round
-
-	// TODO: Also update rebroadcast timeout once implemented according to the
-	//       following pseudocode borrowed from the FIP:
-	//          107:      timeout_rebroadcast ← max(timeout+1, timeout_rebroadcast)
 
 	if justification.Vote.Step == PREPARE_PHASE {
 		i.log("⚠️ swaying from %s to %s by skip to round %d", &i.proposal, chain, i.round)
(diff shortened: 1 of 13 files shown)


masih:
@anorth as always I'd appreciate your early feedback please. 🙏

gpbft/options.go
```diff
@@ -97,3 +102,32 @@ func WithMaxLookaheadRounds(r uint64) Option {
 		return nil
 	}
 }
+
+// WithRebroadcastBackoff sets the duration after the gPBFT timeout has elapsed, at
+// which all messages in the current round are rebroadcast if the round cannot be
+// terminated, i.e. a strong quorum of senders has not yet been achieved.
+//
+// The backoff duration grows exponentially up to the configured max. Defaults to
+// exponent of 1.3 with 3s base backoff growing to a maximum of 30s.
```
anorth, line 111: Nit: The comment about the defaults will likely rot since they're defined 100 lines away.

gpbft/rebroadcast.go
```diff
@@ -0,0 +1,151 @@
...
+	prepareCurrentRound,
+	preparePreviousRound,
+	convergeCurrentRound *MessageBuilder
+
+	instanceAtLatestBroadcast uint64
+	roundAtLatestBroadcast    uint64
+	phaseAtLatestBroadcast    Phase
+}
+
+func (i *instance) tryRebroadcast() error {
```
anorth, line 24: I guess this is kind of conventional in Go, but I find spreading the methods for a single type across multiple files quite confusing. I think some of the methods could be written in against the rebroadcast state instead in which case being here would be fine.

gpbft/rebroadcast.go
```diff
@@ -0,0 +1,151 @@
...
+	attempts int
+
+	decideCurrentRound,
+	commitCurrentRound,
+	commitPreviousRound,
+	prepareCurrentRound,
+	preparePreviousRound,
+	convergeCurrentRound *MessageBuilder
+
+	instanceAtLatestBroadcast uint64
```
anorth, line 19: This is all in the context of a single GPBFT instance, so this instance can never change.

gpbft/rebroadcast.go
```diff
@@ -0,0 +1,151 @@
...
+				if err := n.RequestBroadcast(i.rbs.preparePreviousRound); err != nil {
+					return err
+				}
+			}
+		}
+		return nil
+	}
+}
+
+func (i *instance) shouldRebroadcast() bool {
```
anorth, line 137: This can be a method of RBS by taking the current time as a parameter.

gpbft/rebroadcast.go
```diff
@@ -0,0 +1,151 @@
...
+
+	instanceAtLatestBroadcast uint64
+	roundAtLatestBroadcast    uint64
+	phaseAtLatestBroadcast    Phase
+}
+
+func (i *instance) tryRebroadcast() error {
+	if i.shouldRebroadcast() {
+		previousRebroadcast := i.participant.host.Time()
+		defer func() {
```
anorth, line 27: Is this defer being used explicitly to ensure this logic runs even if the rebroadcast panics? If not, I think it would be much clearer to just call rebroadcast, remember the error value, and then return it later. And if we think the panics can happen, I reckon we should make a larger change to prevent them.

gpbft/gpbft.go
```diff
@@ -378,7 +386,11 @@ func (i *instance) receiveOne(msg *GMessage) (bool, error) {
 		return true, i.tryCommit(msg.Vote.Round)
 	}
 	// Try to complete the current phase in the current round.
-	return true, i.tryCurrentPhase()
+	if err := i.tryCurrentPhase(); err != nil {
+		return true, err
+	}
+
+	return true, i.tryRebroadcast()
```
anorth, line 393: As far as I can tell from the FIP, this should only ever be triggered from the alarm timeout. I don't think we need it here.

anorth (commented):
This is more complicated than I was expecting. It's following the FIP quite closely, but I think the FIP is doing us a great disservice by giving a procedural style specification rather than just describing the properties that we need to meet. I think we should make some effort to simplify.

For recording the messages that we might possibly need to rebroadcast, I think we should start out by simply recording all messages that are sent (in the associated round state, or another round-indexed structure). I wouldn't bother indexing them by anything else. Removing messages from old rounds is an optimization, and if we're doing it, we should just remove the entire round state. When resending the messages for some round later we could filter by phase, (but it should also be ok to just resend them all).

For triggering, my (possibly naive) thought is that in tryPrepare/tryCommit, if the phase timeout has expired but the participant has not received from a strong quorum, then set another alarm for the remaining rebroadcast timeout. If we end up back in tryPrepare with that second timeout having expired, then rebroadcast and set a new rebroadcast alarm. When progress is made and we set a new phase timeout, zero the rebroadcast timeout. Similar but simpler in tryDecide, where there is no concurrent phase timeout.

gpbft/rebroadcast.go
masih: The alternative is giant LOC (by go standards) in one file, e.g. 1.4K LOC at `gpbft.go` and counting. 

I hear you and will move reciever funcs around. 

On a side note, I would like to reduce LOC in `gpbft.go` by breaking it into separate go files. That makes it a lot easier to navigate (for me) 😊

gpbft/rebroadcast.go
masih: > logic runs even if the rebroadcast panics

Yes that was my intention. It is totally up for discussion on how effective this strategy is, since if panics occur here we probably have bigger problems. Having said that I would like to avoid flurry of rebroadcasts (or total halt of rebroadcasts) if code does panic.

>  I reckon we should make a larger change to prevent them.

Yep that makes sense 👍

masih:
The current rebroadcast codepath is getting some coverage from existing test.
In the interest of work parallelisation I will put up a separate PR for more enhanced rebroadcast tests. 
@anorth I'd love your eyes on this PR to check if the implementation is correct before I spend too much time on crafting specific rebroadcast scenarios.

gpbft/gpbft.go
```diff
@@ -918,6 +983,7 @@ func (i *instance) broadcast(round uint64, step Phase, value ECChain, createTick
 	}
 
 	_ = i.participant.host.RequestBroadcast(mb)
+	i.getRound(i.round).broadcasted.record(mb)
```
anorth, line 986: This is fragile in the case of DECIDE messages. A DECIDE message doesn't fundamentally have a round, but is being recorded here against a particular round state. If `i.round` ever changes after this, the DECIDE message may be omitted from subsequent rebroadcasting, though it's the only message that should actually be sent after that point.

We discussed that perhaps it would be better to move the `map[round][]*MB` into the broadcastState rather than have it be spread across the round states, and then the decide message can be stored distinctly alongside that round map. And then you pointed out that we can do pruning of messages that will never need to be rebroadcast upon addition to that state, rather than any filtering while reading it.

anorth (commented):
This is much tighter now, thanks.

The separation in code of setting the timeout from setting the alarm for that timeout makes it a bit difficult to mentally trace, leading to what I think is a timing error.

sim/sim.go
```diff
@@ -107,6 +104,10 @@ func (s *Simulation) Run(instanceCount uint64, maxRounds uint64) error {
 				}
 			}
 
+			// Instantiate the next instance even if it goes beyond finalInstance.
+			// The last incomplete instance is used for testing assertions.
+			currentInstance = s.ec.BeginInstance(*decidedChain, pt)
```
masih, line 109: This was the cause of much head scratching, where: `StartInstance` was being called for the instance that just started instead of the one that just finished. This would then resulted in participants deciding on the wrong base. 

Hence, begin next instance after lagging participants are notified of the completion of current one.

gpbft/gpbft.go
```diff
@@ -918,6 +948,66 @@ func (i *instance) broadcast(round uint64, step Phase, value ECChain, createTick
...
+
+// tryRebroadcast checks whether the conditions for rebroadcasting messages are
+// met, and if so rebroadcasts messages from current and previous rounds.
+// Rebroadcast is needed when both of the following conditions are met:
+//   - There has been no progress since latest broadcast.
+//   - Rebroadcast backoff duration has elapsed.
+//
+// If there has been progress since latest broadcast this function resets the
+// number of rebroadcast attempts and returns immediately.
+func (i *instance) tryRebroadcast() error {
```
masih, line 962: This function is now the only one that alters the rebroadcast related parameters. Making all modifications in one place makes the code logic much easier to follow IMO, where `tryRebroadcast` is called explicitly as part of `try<phase>`.

gpbft/gpbft.go
```diff
@@ -918,6 +948,66 @@ func (i *instance) broadcast(round uint64, step Phase, value ECChain, createTick
...
+func (i *instance) rebroadcast() error {
+
+	// Rebroadcast all messages from the current and previous rounds, unless the
+	// instance has progressed to DECIDE phase. In which case, only DECIDE message is
+	// rebroadcasted.
+	//
+	// Note that the implementation here rebroadcasts more messages than FIP-0086
+	// strictly requires. Because, the cost of rebroadcasting additional messages is
+	// small compared to the reduction in need for rebroadcast.
+	for _, mbs := range i.broadcasted.messagesByRound {
```
masih, line 1003: We can optimise further here by sorting messages before broadcast. But that is throw of the dice optimisation since there are no guaranteed ordering in message broadcast. Hence the simpler approach here.

masih (commented):
Self-review.

gpbft/gpbft.go
```diff
@@ -783,6 +806,13 @@ func (i *instance) tryCommit(round uint64) error {
 		}
 		i.beginNextRound()
 	}
+
+	if timedOut {
```
anorth, line 810: Ditto this isn't true until receiving from a strong quorum, and this should be mutually exclusive with beginNextRound.

gpbft/gpbft.go
```diff
@@ -918,6 +948,66 @@ func (i *instance) broadcast(round uint64, step Phase, value ECChain, createTick
...
+
+	if i.rebroadcastTimeoutElapsed() {
+		err := i.rebroadcast()
+		i.rebroadcastAttempts++
+		i.rebroadcastTimeout = i.rebroadcastTimeout.Add(i.participant.rebroadcastAfter(i.rebroadcastAttempts))
+		return err
+	}
+
+	// Schedule the next rebroadcast.
+	i.participant.host.SetAlarm(i.rebroadcastTimeout)
```
anorth, line 981: I spent some time thinking there was a problem with overriding the alarm set for a phase timeout here. I think I've realised that this assumes that the phase timeout has already passed, and that the alarm has been delivered. I think the phase timeout having passed is true in the call paths, but are we sure that the alarm has been delivered (or if not, that it wouldn't do anything)?

Working through out loud:
1. The participant begins the prepare phase, sends a message, and sets an alarm for phase timeout.
2. The participant receives a prepared message from someone else. The phase timeout has not elapsed, so this method should not be called.
3. The participant receives a message right on the phase timeout. The message gets to the GPBFT code before the alarm (maybe a race in the host). In tryPrepare, the participant notices that the phase has timed out, so invokes this method. There is no progress since last broadcast, and rebroadcast attempts is zero, so the rebroadcast timeout variable is set. An alarm is then set for that rebroadcast timeout, which overrides the phase timeout alarm that would have been delivered immediately next.

This seems technically wrong, but inconsequential. If the alarm had arrived first, the participant still would have ended up here and set a new rebroadcast timeout. It didn't miss any opportunity to make progress based on the phase timeout.

anorth:
@masih I pushed a commit to resolve the only nontrivial issue I noticed.

anorth (approved):

gpbft/gpbft.go
masih: Good point thank you

masih: I have resolved this by only setting the rebroadcast alarm only when it is unset.

masih:
https://github.com/filecoin-project/go-f3/pull/349 should resolve the failing fuzz tests. It is submitted as a separate PR to capture the context for change in commit history.

gpbft/options.go
masih: I moved the default var definition right above the option.

gpbft/gpbft.go
```diff
@@ -1336,6 +1433,49 @@ func (c *convergeState) FindProposalFor(chain ECChain) (ConvergeValue, bool) {
...
+// instance progresses to DECIDE then only the decide message will be recorded.
+// Otherwise, all messages except QUALITY from the current and previous rounds
+// (i.e. the latest two rounds) are recorded.
+//
+// Note, the messages recorded are more than what FIL-00896 strictly requires at
+// the benefit of less reliance on rebroadcast and reduction in the possibility
+// of participants getting stuck.
+func (bs *broadcastState) record(round uint64, mb *MessageBuilder) {
+	bs.phaseAtLatestBroadcast = mb.payload.Step
+	bs.roundAtLatestBroadcast = round
```
anorth, line 1459: This `round` parameter raises questions since there is also `mb.payload.Round`, used below. 

In all the call paths into `broadcast` except the DECIDE phase, the `round` parameter comes from `i.round` and is written into `mb.payload.Round`, so the payload value could be used. For DECIDE, the parameter and payload is 0, but `i.round` is some larger value. The round will be ignored below, but I guess the reason for this is to interact with `progressedSinceLatestBroadcast`.

I think this is correct, but could be simpler. I consider `i.round` is rather meaningless once we reach the DECIDE phase. I'm hesitant to zero it since it's useful for introspection, but I think we should ignore it. How about encapsulating the complexity in `progressedSinceLatestBroadcast`? If the phase is DECIDE, return false without checking the round. Then here we can just record the payload values with no additional parameter. You might then rename the state members to be `lastBroadcastPhase|Round` to avoid confusion there too.

anorth (approved):
I'm happy for this to land as is, but I've made one suggestion that I think could make it easier to mentally follow.

gpbft/gpbft.go
masih: That makes sense thanks Alex

masih:
@anorth A quick note:

Since your latest review I fixed an issue that was causing test failures [here](https://github.com/filecoin-project/go-f3/pull/326/files#diff-758cf112195e26172e6e73622948b1719be52ca9afb3a00feabb4a5152f3d549R971-R984). The issue was that when the first rebroadcast becomes necessary in DECIDE phase, we cannot rely on `phaseTimeout` as the offset for rebroadcast alarm. Because, DECIDE phase does not update the phase timeout. 

I have resolved this by using current time as the offset of first rebroadcast alarm if it is triggered in DECIDE phase for the first time.

Thank you again for all your insightful reviews and comments 🙏

PR #328, opened 2024-06-10
Specify a non-zero participant count for the honest async BLS test

Stebalien:


test/honest_test.go
```diff
@@ -61,6 +61,7 @@ func TestHonest_Agreement(t *testing.T) {
 			name:                 "async pair bls",
 			options:              asyncOptions(1413),
 			useBLS:               true,
+			participantCounts:    blsParticipantCount,
```
Stebalien, line 64: I assume we wanted this, rigth?

masih: I must have missed it. Thanks Steven

masih (approved):

PR #330, opened 2024-06-10
Merge participant Start and SkipToInstance into a single StartInstance method.

anorth:
Removes the initialInstance config option too, in favour of the runtime parameter.

The parameter is necessary for in every case except the initial instance, i.e. it's not a one-time static value like other options. Given the single entry point, if we left the config option there as well, there would be two different initial instances at bootstrap, which would be confusing. This parameter makes the data flow of the initial instance very clear, providing it exactly when it is first needed.

### DIFF
--- a/cmd/f3/run.go
+++ b/cmd/f3/run.go
@@ -30,6 +30,10 @@ var runCmd = cli.Command{
 			Name:  "id",
 			Value: 0,
 		},
+		&cli.Uint64Flag{
+			Name:  "instance",
+			Value: 0,
+		},
 	},
 	Action: func(c *cli.Context) error {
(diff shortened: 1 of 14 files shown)


masih (approved):
This is fine, and too early to tell which approach makes up a better API UX for integrators. 

But why the runtime parameter over option? I'd be grateful if you can document this in the PR description.

Kubuxu:
I don't think we need to expose it in the command. It will get tracked internally.

anorth:
> I don't think we need to expose it in the command. It will get tracked internally.

Feel free to revert or change this in integrations etc, I just wanted to demonstrate plumbing the value all the way through.

PR #331, opened 2024-06-10
Attempt to remove assumption of receiving own messages

### LINKED ISSUE #316: Receiving alarms must take precedence over receiving broadcasted messages

anorth:
Resolves #316 (and see #103)

If this lands, we should revert #318.

### DIFF
--- a/gpbft/gpbft.go
+++ b/gpbft/gpbft.go
@@ -651,12 +651,16 @@ func (i *instance) tryConverge() error {
 		return nil
 	}
 
-	possibleDecisionLastRound := !i.getRound(i.round - 1).committed.HasStrongQuorumFor("")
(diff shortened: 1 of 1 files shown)


gpbft/gpbft.go
```diff
@@ -670,7 +674,10 @@ func (i *instance) tryConverge() error {
 		// Else preserve own proposal.
 		fallback, ok := i.getRound(i.round).converged.FindProposalFor(i.proposal)
```
anorth, line 675: Perhaps at this point it would be better to do what the FIP requests, which is to try the next ticket in sequence rather than wait to receive our own. It still might not be a viable candidate though, so we can still end up here.

anorth:
Closing in favour of #334

Issue #235, opened 2024-05-16
Test bad justifications


Stebalien:
Looking through test coverage, it appears we aren't testing messages with bad justifications:

- [x] Justification with an invalid bitfield.
- [x] Justification for the a different value.
- [x] Justification without strong quorum.
- [x] Justification with bad aggregate signature.
- [x] Justification with a powerless participant.

PR #332, opened 2024-06-11
Test bad justification: no power and invalid bitfield

Stebalien:
- Invalid bitfield
- Participant without power

Part of #235

### DIFF
--- a/sim/adversary/decide.go
+++ b/sim/adversary/decide.go
@@ -10,25 +10,39 @@ import (
 
 var _ Receiver = (*ImmediateDecide)(nil)
 
+type ImmediateDecideOption func(*ImmediateDecide)
+
+func ImmediateDecideWithNthParticipant(n uint64) ImmediateDecideOption {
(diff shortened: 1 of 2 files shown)


sim/adversary/decide.go
```diff
@@ -10,25 +10,39 @@ import (
 
 var _ Receiver = (*ImmediateDecide)(nil)
 
+type ImmediateDecideOption func(*ImmediateDecide)
```
masih, line 13: Use the term ` ImmediateDecision` instead of `ImmediateDecide`?

masih (approved):

sim/adversary/decide.go
Stebalien: Well, it's an option to `ImmediateDecide` where "decide" means "immediately send a decide message" not "immediately make a decision".

masih: And I missed that completely. Thanks Steven; nothing to see here.

PR #334, opened 2024-06-11
Relax the assumption of receiving own `CONVERGE` messages synchronously

### LINKED ISSUE #316: Receiving alarms must take precedence over receiving broadcasted messages

masih:
The gpbft implementation implicitly assumes that broadcast of `CONVERGE` messages to self are delivered immediately. In practice this assumption does not hold because of the complexity in deferred signing and async message delivery.

The changes here relax this assumption by explicitly notifying the local converge state that the self participant has begun the `CONVERGE` step, providing self proposal and justification for the proposal. The code then considers the given data whenever search in converge state does not bear any results, caused by asynchronous message delivery. Further, the code ignores the self converge value once at least one broadcast message is received.

Additionally, the changes remove zero-latency for messages to self in simulations to make a stronger assertion that synchronous message delivery to self is no longer required (neither for `GMessage` nor alarms).

Fixes #316
Reverts #318
Relates to https://github.com/filecoin-project/go-f3/issues/103#issuecomment-2159455477

### DIFF
--- a/gpbft/gpbft.go
+++ b/gpbft/gpbft.go
@@ -640,8 +640,7 @@ func (i *instance) beginConverge(justification *Justification) {
 	// Notify the round's convergeState that the self participant has begun the
 	// CONVERGE phase. Because, we cannot guarantee that the CONVERGE message
 	// broadcasts are delivered to self synchronously.
-	converged := i.getRound(i.round).converged
-	converged.NotifySelfConvergeBegun(i.proposal, justification)
+	i.getRound(i.round).converged.SetSelfValue(i.proposal, justification)
 
 	i.broadcast(i.round, CONVERGE_PHASE, i.proposal, true, justification)
 }
@@ -1223,9 +1222,8 @@ func (q *quorumState) FindStrongQuorumValue() (quorumValue ECChain, foundQuorum 
 //// CONVERGE phase helper /////
 
 type convergeState struct {
-	// self stores the self-participant converge value in order to relax the
-	// assumption that broadcast CONVERGE messages to the participant itself are
-	// delivered synchronously.
+	// Stores this participant's value so the participant can use it even if it doesn't receive its own
+	// CONVERGE message (which carries the ticket) in a timely fashion.
 	self *ConvergeValue
 	// Participants from which a message has been received.
 	senders map[ActorID]struct{}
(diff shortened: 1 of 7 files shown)


Kubuxu:
Seems good to me as an approach

gpbft/gpbft.go
```diff
@@ -1241,6 +1251,23 @@ func newConvergeState() *convergeState {
 	}
 }
 
+// NotifySelfConvergeBegun notifies the convergeState that the self participant
+// has begun the CONVERGE_PHASE. This notification ensures that the convergeState
+// of a round does not rely on messages broadcast by a participant destined for
+// itself to be delivered synchronously. See HasSelfBegunConverge.
+func (c *convergeState) NotifySelfConvergeBegun(value ECChain, justification *Justification) {
```
anorth, line 1258: Nit: rather than the terminology of notifications in the method names and comments, I suggest something simple and direct like `SetSelfValue`, `HasSelfValue`.

gpbft/gpbft.go
```diff
@@ -1281,16 +1308,34 @@ func (c *convergeState) FindMaxTicketProposal(table PowerTable) ConvergeValue {
 			}
 		}
 	}
+
+	// Check if self participant has entered CONVERGE phase.
```
anorth, line 1312: I think we can focus the documentation here a little, but the code looks good.

I would say that if the node has received any converged messages, it finds the best ticket from those. That will include the node's own messages, once they are delivered asynchronously. Fall back to the self-value if nothing has been received yet. Note that this introduces a brief possibility of a node ignoring its own value if it has not received the ticket for it yet.

I don't think we need to talk about checks on whether the phase has started (goes with renaming the method) or make statements about what is assumed to have been sent down at this level. This is a change in focus of the comments from what other code is assumed to have done to direct statements about what this code does.

gpbft/gpbft.go
```diff
@@ -1281,16 +1308,34 @@ func (c *convergeState) FindMaxTicketProposal(table PowerTable) ConvergeValue {
...
+	// broadcast first.
 	for _, value := range c.values {
 		if value.Chain.Eq(chain) {
 			return value, true
 		}
 	}
+
+	// Check if self participant has entered the CONVERGE step, and whether the chain
+	// matches the self proposal. This clause covers an edge-case where self
+	// participant has not received broadcasts about its own CONVERGE messages yet.
```
anorth, line 1334: I think these comments are unnecessary. The code does what it says it will do.

sim/latency/log_normal.go
```diff
@@ -40,10 +40,10 @@ func NewLogNormal(seed int64, mean time.Duration) *LogNormal {
 // distribution. Latency from one participant to another may be asymmetric and
 // once generated remains constant for the lifetime of a simulation.
 //
-// Note, where from and to are the same or mean configured latency is not larger
-// than zero the latency sample will always be zero.
+// Note, mean configured latency is not larger than zero the latency sample will
```
anorth, line 43:
```suggestion
// Note, when mean configured latency is not larger than zero the latency sample will
```

anorth (approved):
Nice, this is better than #331.

PR #336, opened 2024-06-11
Supplemental data is broken

Kubuxu:


masih:
Fixed by https://github.com/filecoin-project/go-f3/pull/338

Issue #229, opened 2024-05-15
Testing utilities for passive testing in network deployment


Kubuxu:
- configuration format suitable to execute test scenarios
- configuration propagation
  - Manifest version in the gpbft messge
  - Manifest fetching
  - applying the manifest 
- re-bootstrapping
- power table editing

masih:
I think we can move all of this as sub tasks under https://github.com/filecoin-project/go-f3/issues/227

Kubuxu:
The intended difference between these two is timing. This needs to land before the Lotus release.

masih:
@Kubuxu I think we can mark this as done?

PR #337, opened 2024-06-11
F3-229: Configuration manifest delivery and handling logic scaffolding

adlrocha:
> This PR is still a WIP, but pushing the draft already for visibility in case someone wants to start having a look.

Partially tackles #229. This PR mainly focuses on:
- The configuration manifest format with all the parameters required to define different test scenarios.
- A new pubsub topic in F3 used to receive manifests that will be periodically published by a central diagnostics server. 
- The basic scaffolding code to apply the manifests when a new one is received, and its upgradeEpoch is reached.

### Manifest application flow
- A centralized manifest server will be periodically publishing the latest manifest being used by the network. 
- F3 nodes are subscribed to this manifest topic and accept manifests that have sequence number over the one of the configuration they are currently using. 
- When a new manifest is received, it is stored to apply the configuration when the `UpgradeEpoch`  is reached.
- Depending on if the `ReBootstrap`  flag of the manifest is set or not, the node will restart a runner to trigger a re-bootstrap, or directly apply the configuration to the existing runner.
   -  Note: Instead of using an explicit flag, we can determine if we need to rebootstrap or not depending on the configuration changes introduced by the manifest. Mainly, parameter changes over GPBFT require a bootstrap, while changes over the power table and related to the EC backend won't need them.
- All configuration changes (i.e. new manifests) trigger a change in the the message pubsub topic where nodes listen and publish messages (that way, we can be sure that messages exchanged belong to the right configuration version). For this, we introduce a manifest suffix to the pubsub topic name that depends on the manifest version being used.


gpbft/types.go
```diff
@@ -1,8 +1,9 @@
 package gpbft
 
 import (
-	"github.com/ipfs/go-datastore"
 	"math/big"
+
+	"github.com/ipfs/go-datastore"
```
anorth, line 6: Drive-by: this datastore and network stuff doesn't really belong in the core gpbft. Could we move these methods out to pure functions closer to where they are called? The datastore key in particular is at the host's discretion.

anorth (commented):
(I was just looking at any changes to protocol code)

cmd/f3/run.go
```diff
@@ -18,6 +18,11 @@ import (
 	"golang.org/x/xerrors"
 )
 
+// ID of the default manifest server which is responsible for periodically
+// publishing it on the diagnostics topic
+// TODO: Make it configurable through a flag.
+var defaultManifestServerID, _ = peer.Decode("12D3KooWHQRSDFv4FvAjtU32shQ7znz7oRbLBryXzZ9NMK2feyyH")
```
masih, line 24: My sense is that this should be externally configurable with no hardcoded value, i.e. if unset no manifest server is taken to consideration.

masih (commented):
I would be curious to understand how much of the mechanisms introduced here will remain in code _after passive testing_.  My recommendation would be to separate passive testing related stuff as much as possible, even in a package of their own?

adlrocha:
> I would be curious to understand how much of the mechanisms introduced here will remain in code after passive testing. My recommendation would be to separate passive testing related stuff as much as possible, even in a package of their own?

Good point, let me try to move things around to make it a bit cleaner.

adlrocha:
@masih, can I have another quick look to see if this is more in line with a decoupling we may be comfortable with? Thanks!

masih:
Looking better thanks @adlrocha please keep going 👍

PR #338, opened 2024-06-11
Broadcast supplemental data as part of `GMessage`

### LINKED ISSUE #335: Supplemental data is not broadcasted
The supplemental data is not broadcasted. Messages are not received if a participant is passed non-zero supplemental data from `GetProposalForIstance`. When broadcasting is fixed, verification of aggregates fails.

masih:
The supplemental data provided by host is required for additional validation. It is already present in the payload of message builder but not the final `GMessaage`s built from it.

The changes here broadcast supplemental data as part of `GMessage` and enhances simulation to generate a unique supplemental data per instance that is required to be consistent across participant for a given instance.

Fixes #335

### DIFF
--- a/gpbft/gpbft.go
+++ b/gpbft/gpbft.go
@@ -895,10 +895,11 @@ func (i *instance) terminated() bool {
 
 func (i *instance) broadcast(round uint64, step Phase, value ECChain, createTicket bool, justification *Justification) {
 	p := Payload{
-		Instance: i.instanceID,
-		Round:    round,
-		Step:     step,
-		Value:    value,
+		Instance:         i.instanceID,
+		Round:            round,
+		Step:             step,
+		SupplementalData: *i.supplementalData,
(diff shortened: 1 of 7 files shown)


gpbft/gpbft.go
```diff
@@ -69,6 +69,8 @@ type GMessage struct {
 	Ticket Ticket
 	// Justification for this message (some messages must be justified by a strong quorum of messages from some previous step).
 	Justification *Justification
+	// SupplementalData is the additional data to aid validation.
+	SupplementalData SupplementalData
```
Kubuxu, line 73: `SupplementalData` is already part of the Vote

masih: doh; thanks.

Stebalien (approved):

anorth (approved):

Kubuxu (approved):

PR #339, opened 2024-06-12
Use consistent request broadcast pattern in simulated sync broadcasts

masih:
To implement some adversary models the simulation package introduces an interface that allows an adversary host to immediately propagate a message across the network.

Since the original implementation the core broadcast mechanisms has evolved into async `RequestBroadcast` that takes a builder. But the pattern is not fully applied to the simulation specific flavour of synchronous broadcast.

The changes here use the same builder pattern in simulation and reflect changes across adversary models.

### DIFF
--- a/sim/network.go
+++ b/sim/network.go
@@ -83,12 +83,20 @@ func (nf *networkFor) Log(format string, args ...any) {
 }
 
 func (nf *networkFor) RequestBroadcast(mb *gpbft.MessageBuilder) error {
+	return nf.requestBroadcast(mb, false)
+}
+
+func (nf *networkFor) RequestSynchronousBroadcast(mb *gpbft.MessageBuilder) error {
+	return nf.requestBroadcast(mb, true)
+}
+
+func (nf *networkFor) requestBroadcast(mb *gpbft.MessageBuilder, sync bool) error {
 	msg, err := mb.Build(nf.networkName, nf.Signer, nf.ParticipantID)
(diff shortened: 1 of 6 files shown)


sim/network.go
```diff
@@ -92,12 +92,18 @@ func (nf *networkFor) RequestBroadcast(mb *gpbft.MessageBuilder) error {
...
 
-func (n *Network) NetworkName() gpbft.NetworkName {
-	return n.networkName
+func (nf *networkFor) RequestSynchronousBroadcast(mb *gpbft.MessageBuilder) error {
+	msg, err := mb.Build(nf.networkName, nf.Signer, nf.ParticipantID)
+	if err != nil {
+		nf.Log("building message for: %d: %+v", nf.ParticipantID, err)
+		return err
+	}
+	nf.broadcast(msg, true)
```
Stebalien, line 101: nit: combine with above in a single helper function `requestBroadcast(mb *gpbft.MessageBuilder, sync bool)`.

Stebalien (approved):
This seems reasonable, but if you want deeper design feedback I don't have enough context. But if you're just looking for a LGTM, go ahead and merge it.

Stebalien:
Removing auto-marge because I'm not sure how much feedback @masih is looking for.

masih:
Thank you Steven for excellent feedback as always ♥️

anorth (approved):

PR #341, opened 2024-06-13
Feed F3 from EC chain and bootstrap

Kubuxu:


gpbft/chain.go
```diff
@@ -36,9 +35,14 @@ func MakeCid(data []byte) []byte {
 	// We construct this CID manually to avoid depending on go-cid (it's also a _bit_ faster).
 	digest := blake2b.Sum256(data)
 
-	out := make([]byte, 0, 38)
+	return DigestToCid(digest[:])
+}
+
+// DigestToCid turns a digest into CBOR + blake2b-256 CID
+func DigestToCid(digest []byte) []byte {
```
Stebalien, line 42: assert that the digest is 32 bytes.

gpbft/chain.go
```diff
@@ -103,7 +107,7 @@ func (ts *TipSet) String() string {
 		return "<nil>"
 	}
 
-	return fmt.Sprintf("%d@%s", ts.Epoch, hex.EncodeToString(ts.Key))
+	return fmt.Sprintf("@%d", ts.Epoch)
```
Stebalien, line 110: why remove the key?

host.go
```diff
@@ -1,15 +1,19 @@
...
 	"time"
 
 	"github.com/filecoin-project/go-f3/certs"
 	"github.com/filecoin-project/go-f3/gpbft"
-	"github.com/filecoin-project/go-f3/sim"
 	"golang.org/x/xerrors"
 )
 
+const instanceLookback = 5
+const finality = 900
```
Stebalien, line 15: IMO, both of these should be manifest parameters.

host.go
```diff
@@ -184,12 +281,13 @@ func (h *gpbftHost) SetAlarm(at time.Time) {
 // based on the decision received (which may be in the past).
 // E.g. this might be: finalised tipset timestamp + epoch duration + stabilisation delay.
 func (h *gpbftHost) ReceiveDecision(decision *gpbft.Justification) time.Time {
-	h.log.Infof("got decision: %+v", decision)
+	h.log.Errorf("got decision, finalized head at epoch: %d", decision.Vote.Value.Head().Epoch)
```
Stebalien, line 284: error?

host.go
```diff
@@ -108,40 +128,117 @@ func (h *gpbftRunner) ValidateMessage(msg *gpbft.GMessage) (gpbft.ValidatedMessa
...
 	if err != nil {
-		return nil, nil, xerrors.Errorf("getting power table: %w", err)
+		return nil, nil, xerrors.Errorf("getting base TS: %w", err)
+	}
+	headTs, err := h.client.ec.GetHead(h.runningCtx)
+	if err != nil {
+		return nil, nil, xerrors.Errorf("getting head TS: %w", err)
+	}
+
+	collectedChain, err := h.collectChain(baseTs, headTs)
```
Stebalien, line 156: This could be an issue if we fall really behind in gpbft and don't see the finality certificates. We could collect a very large chain.

I wonder if, instead, we should have a way to ask the client for the next N epochs after tipset A, on the chain ending in tipset B.

Stebalien: Ah, I see the TODO above now.

host.go
```diff
@@ -108,40 +128,117 @@ func (h *gpbftRunner) ValidateMessage(msg *gpbft.GMessage) (gpbft.ValidatedMessa
...
+		return nil, nil, xerrors.Errorf("getting power table for base: %w", err)
+	}
+	base.PowerTable, err = certs.MakePowerTableCID(pte)
+	if err != nil {
+		return nil, nil, xerrors.Errorf("computing powertable CID for base: %w", err)
 	}
-	ptCid, err := certs.MakePowerTableCID(pt.Entries)
+
+	suffix := make([]gpbft.TipSet, min(gpbft.CHAIN_MAX_LEN-1, len(collectedChain))) // -1 because of base
+	for i := 0; i < len(suffix) && i < len(collectedChain); i++ {
```
Stebalien, line 175:
```suggestion
	for i := range suffix {
```

host.go
```diff
@@ -108,40 +128,117 @@ func (h *gpbftRunner) ValidateMessage(msg *gpbft.GMessage) (gpbft.ValidatedMessa
...
-		PowerTable: ptCid,
+
+	supplData.PowerTable, err = certs.MakePowerTableCID(pt.Entries)
+	if err != nil {
+		return nil, nil, xerrors.Errorf("making power table cid for supplemental data: %w", err)
 	}
 
-	// TODO: use lookback to return the correct next power table commitment and commitments hash.
-	return sd, chain, nil
+	return &supplData, chain, nil
```
Stebalien, line 204: hm. Now I see why you wanted to put supplemental data in `GetCommitteeForIInstance`

host.go
```diff
@@ -108,40 +128,117 @@ func (h *gpbftRunner) ValidateMessage(msg *gpbft.GMessage) (gpbft.ValidatedMessa
...
+	if instance < instanceLookback {
+		//boostrap phase
+		ts, err := h.client.ec.GetTipsetByEpoch(h.runningCtx, h.manifest.BootstrapEpoch-finality)
+		if err != nil {
+			return nil, nil, xerrors.Errorf("getting tipset for boostrap epoch with lookback: %w", err)
+		}
+		powerTsk = ts.Key()
+	} else {
+		// TODO: optimize to use saved power tables
+		cert, err := h.client.certstore.Get(h.runningCtx, instance-instanceLookback)
```
Stebalien, line 219: Can't you just do that now or is there some complexity I'm missing.

gpbft/chain.go
Kubuxu: 100x 64characters in logs are not useful (I might replace it with truncation).

host.go
Kubuxu: temporary, to distinguish it from log spam

host.go
Kubuxu: thanks, it is a remnant from when I didn't `min` the suffix length

host.go
Kubuxu: yeah, I can, I just threw out a draft PR because it was working

gpbft/chain.go
Stebalien: Hm. The issue is that it makes it harder to tell why/how nodes disagree (they'll often agree on the epochs). But yeah, truncating it should be good enough.

f3.go
```diff
@@ -231,7 +233,22 @@ loop:
...
 }
 
-type ECBackend interface{}
+type ECBackend interface {
+	// GetTipsetByEpoch should return a tipset before the one requested if the requested
+	// tipset does not exist due to null epochs
+	GetTipsetByEpoch(ctx context.Context, epoch int64) (TipSet, error)
+	GetTipset(context.Context, gpbft.TipSetKey) (TipSet, error)
+	GetHead(context.Context) (TipSet, error)
+	GetParent(context.Context, TipSet) (TipSet, error)
```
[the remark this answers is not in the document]
Kubuxu: The idea here was that whoever implements it, can try type-checking the TipSet into their own type and save one having to fetch the tipset based on key, with a fallback of calling `Key()` and fetching the tipset.

host.go
```diff
@@ -108,40 +125,129 @@ func (h *gpbftRunner) ValidateMessage(msg *gpbft.GMessage) (gpbft.ValidatedMessa
...
+		cert, err := h.client.certstore.Get(h.runningCtx, instance-h.manifest.CommiteeLookback)
+		if err != nil {
+			return nil, nil, xerrors.Errorf("getting finality certificate: %w", err)
+		}
+		powerTsk = cert.ECChain.Head().Key
+
+		powerEntries, err = h.client.certstore.GetPowerTable(h.runningCtx, instance)
+		if err != nil {
+			// this fires every round, is this correct?
+			h.log.Infof("failed getting power table from certstore: %v, falling back to EC", err)
```
Stebalien, line 230: We should never see an error here as long as we've already inserted the certificate for the previous instance. I wonder if we have an off-by-one somewhere? Or are we not inserting the cert?

Stebalien (approved):
Code LGTM except that one thing about the certificate store.

host.go
Kubuxu: It is always off by one in the error message

gpbft/chain.go
```diff
@@ -102,8 +110,9 @@ func (ts *TipSet) String() string {
 	if ts == nil {
 		return "<nil>"
 	}
+	encTs := base32.StdEncoding.EncodeToString(ts.Key)
 
-	return fmt.Sprintf("%d@%s", ts.Epoch, hex.EncodeToString(ts.Key))
+	return fmt.Sprintf("%s@%d", encTs[:16], ts.Epoch)
```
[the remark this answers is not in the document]
Kubuxu: 👍

host.go
Stebalien: What's the specific message and/or can I reproduce it somehow?

Stebalien: OH! It's from here:

https://github.com/filecoin-project/go-f3/blob/cd30c14975edb31dfeae832596d73579ab1cad0c/host.go#L314-L317

Kubuxu: This one: https://github.com/filecoin-project/go-f3/blob/main/certstore/certstore.go#L282

You can reproduce by running:
`./f3 manifest gen` and then `./f3 run --id=0` and `./f3 run --id=1` in two different terminals.

PR #344, opened 2024-06-14
Follow golang naming convention for constants

masih:
The codebase inconsistently follows the MaxCaps golang naming convention for constant.

Consistently use MaxCaps for all constants, except for PHASEs to keep the code easier to read.

### DIFF
--- a/gpbft/gpbft.go
+++ b/gpbft/gpbft.go
@@ -52,7 +52,7 @@ func (p Phase) String() string {
 	}
 }
 
-const DOMAIN_SEPARATION_TAG = "GPBFT"
+const DomainSeparationTag = "GPBFT"
 
 // A message in the Granite protocol.
 // The same message structure is used for all rounds and phases.
@@ -132,7 +132,7 @@ func (p *Payload) MarshalForSigning(nn NetworkName) []byte {
 	root := merkle.Tree(values)
 
 	var buf bytes.Buffer
-	buf.WriteString(DOMAIN_SEPARATION_TAG)
+	buf.WriteString(DomainSeparationTag)
 	buf.WriteString(":")
 	buf.WriteString(string(nn))
 	buf.WriteString(":")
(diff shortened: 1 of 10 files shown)


Stebalien:
I'd say to wait for code freeze.

masih:
> I'd say to wait for code freeze.

Would it make that much of a difference? There are not that many PRs open and it's unlikely for open PRs/ongoing-work to have renamed constants meaning hopefully no conflicts, or easy to fix ones at that. I would just rip the bandaid.

Having said that, I have no strong opinions.

Stebalien:
There aren't that many open PRs, but this will conflict with anything anyone is working on locally. Given that we can make this change programatically at any time, I'd prefer to wait.

But I also don't have strong opinions on this.

gpbft/gpbft.go
```diff
@@ -20,38 +20,38 @@ import (
...
-	COMMIT_PHASE
-	DECIDE_PHASE
-	TERMINATED_PHASE
+	InitialPhase Phase = iota
+	QualityPhase
+	ConvergePhase
+	PreparePhase
+	CommitPhase
+	DecidePhase
+	TerminatedPhase
```
anorth, line 29: Is there any syntax or construct we can use in Go that will give permission to use ALL_CAPS for these constants? For these ones in particular I think their use is much clearer in all caps.

anorth (approved):
I don't like the change for the phase names in particular, but if this is a clearly established convention I'm not going to argue.

anorth:
I removed from the merge queue just to leave space for a second thought about when to merge this or if we indeed want to change everything.

masih:
Marked as draft to revisit just before code freeze as recommended by the team.

gpbft/gpbft.go
masih: We could keep those as all caps if that makes it clear. I have no strong opinions.

Kubuxu (approved):

PR #348, opened 2024-06-18
F3-229: manifest provider, ec decoupling from f3, and dynamic manifest update

adlrocha:
Partially tackles #229 

## Background
In #337 I did all of the exploration for the implementation end-to-end of #229, but as the PR was growing too big, reviewing would have been quite cumbersome, and rebasing was becoming a bit of a pain, I decided to start dividing the work into smaller self-contained PRs that can (hopefully) be reviewed and merged faster. 

(This will really make my life easier every time that I need to rebase :) ) 

## Implementation
This PR: 
- Decoupled `EcBackend`  and ` FakeEC`  from f3 into its own independent package so it can be used for the integration tests that I was writing in #337 without circular dependencies.
- Implement a `ManifestProvider` interface that abstracts the specific manifest implementation used under the hood. This will hopefully simplify the removal of all the passive testing-specific logic in the future. 
  - There are two implementation of `ManifestProvider`: `Static` , which doesn't accept any runtime changes and is just a bunch of accessors to the underlying manifest, and `DynamicManifest`  which will be the one used to trigger configuration changes in passive testing. 
  - To simplify the review, I haven't pushed the whole implementation of `DynamicManifest`, this will come in the next PR. 
- As suggested by Alex [here](https://github.com/filecoin-project/go-f3/pull/337#discussion_r1635511935), moved network-specific types out of the `gpbft`  protocol 

## Coming Next
- [x] Implementation of `DynamicManifest`  and integration in f3 to trigger configuration rebootstraps.
- [x] Implementation of `ManifestSender`  responsible for periodically broadcasting the current manifest used by the network.
- [x]  Integration tests for f3
- [x]  Implementation of configuration changes without rebootstrap.


ec/ec.go
```diff
@@ -0,0 +1,26 @@
+package ec
+
+import (
+	"context"
+	"time"
+
+	"github.com/filecoin-project/go-f3/gpbft"
+)
+
+type ECBackend interface {
```
masih, line 10: Avoid repeating package name in type names.

ec/ec.go
```diff
@@ -0,0 +1,26 @@
...
+	// tipset does not exist due to null epochs
+	GetTipsetByEpoch(ctx context.Context, epoch int64) (TipSet, error)
+	GetTipset(context.Context, gpbft.TipSetKey) (TipSet, error)
+	GetHead(context.Context) (TipSet, error)
+	GetParent(context.Context, TipSet) (TipSet, error)
+
+	GetPowerTable(context.Context, gpbft.TipSetKey) (gpbft.PowerEntries, error)
+}
+
+type TipSet interface {
```
masih, line 21: There is a [`TipSet` type](https://github.com/filecoin-project/go-f3/blob/650861a1a3c36f9103b9010a42ce8e6b0005cb69/gpbft/chain.go#L54) already in `gpbft` package. It would be nice to unify them if they intend to covey the same entity, or rename this one to express that they are indeed different. 

This can be done in a separate PR if it would touch too many lines.

f3.go
```diff
@@ -88,10 +89,10 @@ func (mc *client) Logger() Logger {
 
 // New creates and setups f3 with libp2p
 // The context is used for initialization not runtime.
-func New(ctx context.Context, id gpbft.ActorID, manifest Manifest, ds datastore.Datastore, h host.Host,
-	ps *pubsub.PubSub, sigs gpbft.SignerWithMarshaler, verif gpbft.Verifier, ec ECBackend, log Logger) (*F3, error) {
-	ds = namespace.Wrap(ds, manifest.NetworkName.DatastorePrefix())
-	cs, err := certstore.OpenOrCreateStore(ctx, ds, 0, manifest.InitialPowerTable)
+func New(ctx context.Context, id gpbft.ActorID, manifest manifest.ManifestProvider, ds datastore.Datastore, h host.Host, manifestServer peer.ID,
+	ps *pubsub.PubSub, sigs gpbft.SignerWithMarshaler, verif gpbft.Verifier, ec ec.ECBackend, log Logger) (*F3, error) {
```
masih, line 93: 😮‍💨  11 args. I would introduce options here once the implementation settles (but before code freeze).

f3.go
```diff
@@ -88,10 +89,10 @@ func (mc *client) Logger() Logger {
...
+	ds = namespace.Wrap(ds, manifest.DatastorePrefix())
```
masih, line 94: Thinking out loud: should `Manifest` have something as low level as datastore prefix key? We might want to infer it from some other parameter as it seems like lower level impelentation detail as far as `Manifest` goes.

gpbft/types_test.go
```diff
@@ -1,36 +0,0 @@
...
-		t.Run(test.name, func(t *testing.T) {
-			gotDsPrefix := test.subject.DatastorePrefix().String()
-			require.True(t, strings.HasPrefix(gotDsPrefix, "/f3"))
-			require.True(t, strings.HasSuffix(gotDsPrefix, string(test.subject)))
-			gotPubSubTopic := test.subject.PubSubTopic()
-			require.True(t, strings.HasPrefix(gotPubSubTopic, "/f3/granite/0.0.1/"))
-			require.True(t, strings.HasSuffix(gotPubSubTopic, string(test.subject)))
-		})
-	}
-}
```
masih, line 36: `git mv` to preserve commit history?

manifest/manifest.go
```diff
@@ -0,0 +1,150 @@
...
+	Run(context.Context, chan error)
+	// Returns the list of gpbft options that should be used
+	GpbftOptions() []gpbft.Option
+	// suscribe to manifest updates
+	Subscribe() <-chan struct{}
+	// Set callback for manifest changes
+	SetManifestChangeCb(OnManifestChange)
+
+	// Manifest accessors
+	MsgPubSubTopic() string
```
masih, line 32: I recommend avoiding `Msg` shortened word at this interface api. To use consistent terminology with GPBFT, would it make sense to call this `BroadcastTopic` or simply drop `Msg`?

manifest/manifest.go
```diff
@@ -0,0 +1,150 @@
...
+type OnManifestChange func(ctx context.Context, initialInstance uint64, rebootstrap bool, errCh chan error)
+
+type ManifestProvider interface {
+	// Run starts any background tasks required for the operation
+	// of the manifest provider.
+	Run(context.Context, chan error)
+	// Returns the list of gpbft options that should be used
+	GpbftOptions() []gpbft.Option
+	// suscribe to manifest updates
+	Subscribe() <-chan struct{}
```
masih, line 27: * Document what the expected semantics is: e.g. how many subscribers can subscribe here?
* What about unsubscribe? A pattern you can use is to return an unsubscribe function along with the channel.

manifest/manifest.go
```diff
@@ -0,0 +1,150 @@
...
+	// Set callback for manifest changes
+	SetManifestChangeCb(OnManifestChange)
```
masih, line 29: Without the comment above this function it would have taken me loner than it should to infer `Cb` is short for "Callback". 
* I recommend a clearer name and documentation of semantics (similar to what I recommended for `Subscribe`)
* At API design level, it seems odd to have a subscription mechanism to receive notifications, _and_ a hook registry to execute once manifest changes. I am curious if both are necessary and whether the API surface here can be reduced.

manifest/manifest.go
```diff
@@ -0,0 +1,150 @@
...
+
+	// Manifest accessors
+	MsgPubSubTopic() string
+	NetworkName() gpbft.NetworkName
+	BootstrapEpoch() int64
+	DatastorePrefix() datastore.Key
```
masih, line 35: Ditto re implementation detail. Is this something that needs to be exposed at interface level?

manifest/manifest.go
```diff
@@ -0,0 +1,150 @@
...
+	GpbftOptions() []gpbft.Option
+	// suscribe to manifest updates
+	Subscribe() <-chan struct{}
+	// Set callback for manifest changes
+	SetManifestChangeCb(OnManifestChange)
+
+	// Manifest accessors
+	MsgPubSubTopic() string
+	NetworkName() gpbft.NetworkName
+	BootstrapEpoch() int64
```
masih, line 34: Thinking out loud: Would it make sense to set bootstrap `TipSet`, where the tipset encapsulates all the info needed for the initial bootstrap (e.g. power table, supplemental data, etc.). That should simplify this API and get rid of having one api per bootstrap-info "stuff".

manifest/manifest.go
```diff
@@ -0,0 +1,150 @@
...
+
+type Version string
+
+type GpbftConfig struct {
+	Delta                time.Duration
+	DeltaBackOffExponent float64
+	MaxLookaheadRounds   uint64
+}
+
+type EcConfig struct {
```
masih, line 48: Does this belong to `ec` package?

manifest/manifest.go
```diff
@@ -0,0 +1,150 @@
...
+	ECFinality       int64
+	ECDelay          time.Duration
+	CommiteeLookback uint64
+
+	//Temporary
```
masih, line 53: Can you expand on this please? As in to be removed after passive testing? If so, i would leave an explicit `// TODO: `

manifest/manifest.go
```diff
@@ -0,0 +1,150 @@
...
+	// Updates to perform over the power table retrieved by the host
+	// starting from BootstrapEpoch.
+	PowerUpdate []certs.PowerTableDelta
+	// Config parameters for gpbft
+	*GpbftConfig
+	// EC-specific parameters
+	*EcConfig
+}
+
+func LocalnetManifest() Manifest {
```
masih, line 84: use the phrase `localDevnet` for consistency with the rest of the filecoin ecosystem?

manifest/manifest.go
```diff
@@ -0,0 +1,150 @@
...
+	CommiteeLookback uint64
+
+	//Temporary
+	ECPeriod            time.Duration
+	ECBoostrapTimestamp time.Time
+}
+
+// Manifest identifies the specific configuration for
+// the F3 instance currently running.
+type Manifest struct {
```
masih, line 60: Dumb question: Why both `ManifestProvider` interface with  verbose per-field API, and `Manifest` struct? Could `ManifestProvider` be a simple one liner interface API that returns `Manifest`?

manifest/manifest.go
```diff
@@ -0,0 +1,150 @@
...
+	*EcConfig
+}
+
+func LocalnetManifest() Manifest {
+	rng := make([]byte, 4)
+	_, _ = rand.Read(rng)
+	m := Manifest{
+		NetworkName:    gpbft.NetworkName(fmt.Sprintf("localnet-%X", rng)),
+		BootstrapEpoch: 1000,
+		EcConfig: &EcConfig{
```
masih, line 90: Extract to some default EC config `var`.

manifest/manifest.go
```diff
@@ -0,0 +1,150 @@
...
+			ECFinality:       900,
+			CommiteeLookback: 5,
+			ECDelay:          30 * time.Second,
+
+			ECPeriod: 30 * time.Second,
+		},
+	}
+	m.ECBoostrapTimestamp = time.Now().Add(-time.Duration(m.BootstrapEpoch) * m.ECPeriod)
```
masih, line 98: Can I ask what this timestamp is used for?

manifest/manifest.go
```diff
@@ -0,0 +1,150 @@
...
+	NetworkName gpbft.NetworkName
+	// Initial power table used when bootstrapping. This is used for test
+	// networks or if we want to force a specific power table when
+	// rebootstrapping. If nil, the power table is fetched from the host
+	InitialPowerTable gpbft.PowerEntries
+	// Updates to perform over the power table retrieved by the host
+	// starting from BootstrapEpoch.
+	PowerUpdate []certs.PowerTableDelta
+	// Config parameters for gpbft
+	*GpbftConfig
```
masih, line 79: Why top level field? Ditto for `EcConfig`.

manifest/manifest.go
```diff
@@ -0,0 +1,150 @@
...
+	NetworkName() gpbft.NetworkName
+	BootstrapEpoch() int64
+	DatastorePrefix() datastore.Key
+	InitialPowerTable() []gpbft.PowerEntry
+	EcConfig() *EcConfig
+}
+
+type Version string
+
+type GpbftConfig struct {
```
masih, line 42: Simply take `gopbft.Option` as a field of `Manifest`?

manifest/manifest.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+func (m Manifest) PubSubTopic() string {
+	return "/f3/granite/0.0.1/" + string(m.NetworkName)
+}
+
+func (m Manifest) GpbftOptions() []gpbft.Option {
+	var opts []gpbft.Option
+
+	if m.Delta != 0 {
```
masih, line 139: These `if` checks for default are technically incorrect? i.e. don't cover a case where user explicitly sets something to zero value. 

I recommend thinking about if that matters here, and if so use pointers for fields instead and check if they are `nil`. Better yet, avoid it all by taking options directly

manifest/manifest_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+	m.Delta = 1
+	m.NetworkName = "test2"
+	v1, err = m.Version()
+	require.NoError(t, err)
+	require.NotEqual(t, v1, v2)
+
+}
+
+func TestNetworkName(t *testing.T) {
```
masih, line 88: `TestManifest_NetworkName` ? This is a convention that so far has worked well for me. Ditto for the rest of the tests. At a very least i would include `Manifest` in the test name, since that's what it is testing.

manifest/static.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"context"
+
+	"github.com/filecoin-project/go-f3/gpbft"
+	"github.com/ipfs/go-datastore"
+)
+
+// Static manifest provider that doesn't allow any changes
+// in runtime to the initial manifest set in the provider
+type Static struct {
```
masih, line 12: `StaticManifestProvider`?

manifest/static.go
```diff
@@ -0,0 +1,56 @@
...
+	manifest *Manifest
+}
+
+var _ ManifestProvider = (*Static)(nil)
+
+func NewStaticManifest(m *Manifest) ManifestProvider {
```
masih, line 18: Ditto re naming, i.e. include `Provider` suffix.

passive/dynamic_manifest.go
```diff
@@ -0,0 +1,90 @@
...
+func (m *DynamicManifest) NetworkName() gpbft.NetworkName {
+	return m.manifest.NetworkName
+}
+
+func (m *DynamicManifest) InitialPowerTable() []gpbft.PowerEntry {
+	return m.manifest.InitialPowerTable
+}
+
+func (m *DynamicManifest) Subscribe() <-chan struct{} {
+	return m.manifestUpdates
```
masih, line 72: This means only one of the subscribers will receive the update. Since the intended semantics of `Subscribe` is undocumented I don't know if this is intended.

ec/ec.go
adlrocha: chiming @Kubuxu in here for visibility and to give additional context. This type comes from his EC integration, and I just moved it to its own independent package.

f3.go
adlrocha: Agree. Definitely a debt we will want to pay :)

gpbft/types_test.go
adlrocha: Unfortunately, this is not exactly a `git mv` but an aggregation of different tests into the same way. Can history be preserved then? (AFAIK. `git mv` is only for files, right?

manifest/manifest.go
adlrocha: +1. Dropping `Msg` works. The reason for the prefix is because before we had a different pubsub topic for the manifest (when everything was way less decoupled).

manifest/manifest.go
adlrocha: Chiming @Kubuxu in again. This is not temporary from the passive testing point of view, but I guess the `FakeEC`  one which uses strict timing assumptions.

manifest/manifest.go
adlrocha: fixed.

manifest/manifest.go
adlrocha: Good catch, thanks! Initially (before the PRs for EC and other dependencies), I was thinking that for testing providers may need different underlying manifests (so the `StaticProvider` would use a different underlying Manifest that is not the same to the one used by the `DynamicProvider`). It is clear now that this won't be the case, so I will go with your suggestion of directly exposing the Manifest. Thanks!

manifest/manifest.go
adlrocha: Fixed by exposing the whole Manifest as discussed in the other comment.

manifest/manifest.go
adlrocha: This may not be possible (unless I am missing something). We need the Manifest to be serializable so it can me persisted and moved around, and using `gpbft.Option` required implementing all the mechanics to make them JSON serializable.

manifest/manifest.go
adlrocha: I updated it a bit allowing zero to be set except for `Delta`  which doesn't make much sense.

manifest/manifest.go
adlrocha: Make the access less verbose, but no strong opinions here.

manifest/manifest.go
adlrocha: This is the EC configuration propagated by the manifest. The manifest is used by the fake EC backend. If I include it as part of that package it leads to a circular dependency.

manifest/manifest.go
adlrocha: I think you may be right, it is not clear yet if I'll need the `Subscribe()`  in the end. Before I needed it to notify the runner that there was a new manifest so it can pick up the new configuration. Now that you mention it, I may be able to integrate this as part of the F3 callback. Unfortunately, that logic comes in the next PR, so if you don't mind, let's revisit this one with the whole logic implemented in the next PR. Thanks!

gpbft/types_test.go
masih: I would move the file, then rename it then add the other tests. but no biggy :)

ec/ec.go
Stebalien: These are, unfortunately, _different_ tipsets.

1. This is an abstraction over the client's concept of a tipset and doesn't, e.g., carry a commitment to a power table etc.
2. The one in the `gpbft` package contains additional commitments, etc.

Honestly, we may just want to rename the one in the `gpbft` package to... ECBlock/ECBlockKey? Or something like that? But that might be even more confusing.

manifest/manifest.go
Kubuxu: It is used for producing new blocks/new head over time. We should probably abstract out time for tests.

manifest/manifest.go
Kubuxu: I think we should try removing them; they should only be used as part of the FakeEC, which is outliving its purpose. If we have to hardcode them or make the standalone node inoperable, I'm fine with that (as long as we have tests that Alfonso is adding).

I'm fine with us breaking the standalone node for cleanup purposes, especially if it is FakeEC as it will need a lightweight FilecoinAPI client in there.

manifest/manifest.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+// Manifest identifies the specific configuration for
+// the F3 instance currently running.
+type Manifest struct {
+	// Sequence number of the manifest.
+	// This is used to identify if a new config needs to be applied
+	Sequence uint64
+	// Initial instance to use for gpbft
+	InitialInstance uint64
```
Kubuxu, line 80: This is the property of the whole F3, not only gpbft. We would have teardown F3 and certstore to change it.

manifest/manifest.go
```diff
@@ -0,0 +1,161 @@
...
+	BootstrapEpoch int64
+	// Flag to determine if the peer should rebootstrap in this configuration
+	// change at BootstrapEpoch
+	ReBootstrap bool
+	// Network name to apply for this manifest.
+	NetworkName gpbft.NetworkName
+	// Initial power table used when bootstrapping. This is used for test
+	// networks or if we want to force a specific power table when
+	// rebootstrapping. If nil, the power table is fetched from the host
+	InitialPowerTable gpbft.PowerEntries
```
Kubuxu, line 91: InitialPowerTable does not exist anymore. It is fetched from EC. This is another remnant of FakeEC.

manifest/manifest.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+// Version that uniquely identifies the manifest.
+func (m Manifest) Version() (Version, error) {
+	b, err := json.Marshal(m)
+	if err != nil {
+		return "", xerrors.Errorf("computing manifest version: %w", err)
+	}
+	return Version(hex.EncodeToString(gpbft.MakeCid(b))), nil
+}
```
Kubuxu, line 120: I don't like relying on stable serialization of JSON. I see that we use it for the topic name. Can we use the sequence number of it instead?

passive/dynamic_manifest.go
```diff
@@ -0,0 +1,222 @@
...
+		case <-ticker.C:
+			if m.nextManifest != nil {
+				ts, err := m.ec.GetHead(ctx)
+				if err != nil {
+					log.Errorf("error fetching chain head: %+v", err)
+					continue
+				}
+
+				// if the upgrade epoch is reached or already passed.
+				if ts.Epoch() >= m.nextManifest.BootstrapEpoch {
```
Kubuxu, line 94: What about the case where we didn't boostrap yet? we might want to swap out the manifest either way

Kubuxu (commented):
Notes:
- changing the pubsub topic name will run into issues in Lotus because Lotus has a whitelist of pubsub topics which are allowed to be used, not sure why exactly but it will be a problem

manifest/manifest.go
adlrocha: Fixed

test/f3_test.go
```diff
@@ -0,0 +1,545 @@
...
+	ec             *ec.FakeEC
+
+	manifestSender *passive.ManifestSender
+}
+
+// signals the update to the latest manifest in the environment.
+func (e *testEnv) updateManifest() {
+	e.manifestSender.UpdateManifest(&e.manifest)
+	// wait for the manifest to propagate
+	time.Sleep(2 * ManifestSenderTimeout)
```
masih, line 227: What indicates that the manifest has propagated? Wrap that logic in a `require.Eventually` instead of arbitrary wait time.

masih (commented):
Blockers:
* flow of `setupGpbftRunner` when errors occur
* arbitrary timeouts in tests.

Left some suggestions. Thank you for powering through this @adlrocha 💪

f3.go
```diff
@@ -254,102 +326,148 @@ func (m *F3) GetCert(ctx context.Context, instance uint64) (*certs.FinalityCerti
...
+	case <-ctx.Done():
+		if ctx.Err() != nil {
+			return ctx.Err()
 		}
-		err := runner.Run(startInstance, ctx)
-		m.log.Errorf("running host: %+v", err)
-		runnerErrCh <- err
-	}()
+	case err = <-runnerErrCh:
+		return err
```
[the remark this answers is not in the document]
masih: Right you are.

f3.go
```diff
@@ -183,35 +202,88 @@ func (m *F3) setupPubsub(runner *gpbftRunner) error {
...
+	if err != nil {
+		errCh <- xerrors.Errorf("subscribing to topic: %w", err)
+		return
+	}
+
+	messageQueue := make(chan gpbft.ValidatedMessage, 20)
+	m.client.messageQueue = messageQueue
+
+	go func() {
+		latest := m.certStore.Latest()
```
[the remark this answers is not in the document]
masih: I think not doing it means instance gets started even if context is cancelled. We probably want to check the context eitehr here or earlier in Run.

f3.go
```diff
@@ -183,35 +202,88 @@ func (m *F3) setupPubsub(runner *gpbftRunner) error {
...
+		return
+	}
+
+	m.msgSub, err = m.client.topic.Subscribe()
+	if err != nil {
+		errCh <- xerrors.Errorf("subscribing to topic: %w", err)
+		return
+	}
+
+	messageQueue := make(chan gpbft.ValidatedMessage, 20)
```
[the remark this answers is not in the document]
masih: Can you add the above comment to the code please?

f3.go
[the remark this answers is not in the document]
adlrocha: You are right, fixed.

adlrocha:
@masih, I think I addressed all of your comments, rebased again, and fixed tests. If you could have a final look and don't have any big blocker it would be great if we could get this one merged. It is getting harder and harder to maintain this branch with every new commit as the PR is getting huge :see_no_evil:. Thanks :pray:

manifest/dynamic_manifest.go
```diff
@@ -0,0 +1,226 @@
...
+		select {
+		case <-ctx.Done():
+			break loop
+
+		default:
+			var msg *pubsub.Message
+			msg, err = manifestSub.Next(ctx)
+			if err != nil {
+				if ctx.Err() != nil {
+					break
```
masih, line 144: Should this be `break loop`?

manifest/dynamic_manifest.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+
+	manifestSub, err := m.manifestTopic.Subscribe()
+	if err != nil {
+		errCh <- xerrors.Errorf("subscribing to topic: %w", err)
+		return
+	}
+
+loop:
+	for {
```
masih, line 134: It would make the logic simpler to add a condition here to check if contex has error or not: `for ctx.Err() == nil {`

manifest/dynamic_manifest.go
```diff
@@ -0,0 +1,226 @@
...
+				log.Errorf("invalid manifestValidatorData: %+v", msg.ValidatorData)
+				continue
+			}
+
+			m.acceptNextManifest(manifest)
+		}
+	}
+
+	manifestSub.Cancel()
+	if err := m.teardownManifestPubsub(); err != nil {
```
masih, line 160: Cancel and tear down could go in a `differ` function to further simplify the code path if this function will keep doing its thing until context is done.

manifest/dynamic_manifest.go
```diff
@@ -0,0 +1,226 @@
...
+// Checks if we should accept the manifest that we received through pubsub
+// and sets nextManifest if it is the case
+func (m *DynamicManifestProvider) acceptNextManifest(manifest *Manifest) {
+	m.lk.Lock()
+	defer m.lk.Unlock()
+
+	if manifest.Sequence <= m.manifest.Sequence {
+		return
+	}
+	// TODO: Any additional logic here to determine what manifests to accept or not?
```
masih, line 185: Double checking that the manifest signature/sender is confirmed to be the passive test control server before we hit this.

masih (approved):
The failing tests with race detector need to be resolved. 
Otherwise Looks good thanks @adlrocha 👍 

Left some suggestions.

PR #354, opened 2024-06-19
Wait for boostrap epoch

Kubuxu:
Also fix a off-by-one in GetTipsetByEpoch


Kubuxu:
@Stebalien this is a bit nasty, I need it to unblock me on lotus

f3.go
```diff
@@ -122,8 +116,21 @@ func New(ctx context.Context, id gpbft.ActorID, manifest Manifest, ds datastore.
 		},
 	}
 
+	cs, err := certstore.OpenStore(ctx, ds)
+	if err != nil {
+		log.Warnf("opening CertStore, it might not have been created: %+v", err)
```
Stebalien, line 121: We should distinguish between the two cases.

f3.go
```diff
@@ -170,11 +177,69 @@ func (m *F3) teardownPubsub() error {
...
+			aim := time.Until(head.Timestamp().Add(m.Manifest.ECPeriod))
+			// correct for null epochs
+			for aim < 0 {
+				aim += m.Manifest.ECPeriod
+			}
+
+			select {
+			case <-time.After(aim):
+			case <-ctx.Done():
+				return nil
```
Stebalien, line 207: This should be an error.

f3.go
```diff
@@ -170,11 +177,69 @@ func (m *F3) teardownPubsub() error {
...
+	cs, err := certstore.CreateStore(ctx, m.ds, initialInstance, initialPowerTable)
+	if err != nil {
+		return xerrors.Errorf("creating certstore: %w", err)
+	}
+	m.setCertStore(cs)
+	return nil
+}
+
 // Run start the module. It will exit when context is cancelled.
 func (m *F3) Run(initialInstance uint64, ctx context.Context) error {
```
Stebalien, line 231: unrelated, we should get the initial instance from the certificate store.

f3.go
Stebalien: Actually, I'd just _never_ open the cert store here. Instead:

1. On start, always wait until we've passed the bootstrap epoch.
2. Then open the store.

f3.go
```diff
@@ -170,11 +177,69 @@ func (m *F3) teardownPubsub() error {
...
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
+	if m.CertStore == nil {
```
Stebalien, line 235: Make the certstore private or someone will trip over it.

Stebalien (approved):
This will probably be fine given that nothing really happens outside of `Run` (assuming that's the case?). I don't like, but it'll probably be fine.

Kubuxu:
Merging to unblock further work

PR #355, opened 2024-06-19
Relax the need for synchronous broadcast of `COMMIT` messages to self

### LINKED ISSUE #351: `beginNextRound` constains assumption on recieving local messages
https://github.com/filecoin-project/go-f3/blob/ad199a817719535579eff6092ce9d66f00171b9a/gpbft/gpbft.go#L854-L858

@adlrocha observes a panic from GPBFT in that piece of code.
Commit might require a similar treatment as converge: https://github.com/filecoin-project/go-f3/pull/334

Part of https://github.com/filecoin-project/go-f3/issues/319

masih:
The message broadcast mechanism cannot guarantee that broadcasts to self are delivered synchronously. This means it is possible for next round with justification for non-bottom value from self to begin before the justification for it is delivered.

Relax this assumption by directly updating the committed quorum state when the participant itself finds a strong quorum of PREPARE and builds a justification for it. This change guarantees that regardless of the order of message delivery in broadcast the next round can begin when initiated by a justification from self.

Fixes #351

### DIFF
--- a/gpbft/gpbft.go
+++ b/gpbft/gpbft.go
@@ -759,6 +759,12 @@ func (i *instance) beginCommit() {
 		if quorum, ok := i.getRound(i.round).prepared.FindStrongQuorumFor(i.value.Key()); ok {
 			// Found a strong quorum of PREPARE, build the justification for it.
 			justification = i.buildJustification(quorum, i.round, PREPARE_PHASE, i.value)
+
+			// Update the round's committed quorumState with the self participant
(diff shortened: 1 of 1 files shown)


Kubuxu (approved):

masih:
On second look `beginNextRound` cannot be called unless a strong quorum is found or message from strong quorum of senders is received.

Kubuxu:
Ahh, so this was caused by the zero power table.

PR #359, opened 2024-06-21
Implement a GPBFT emulator and test state transitions

masih:
Implement an emulator to test the core GPBFT protocol, with fine-grained control and assertion over message-by-message progress of a participant.

Add utility functions to reduce the boilerplate code during testing and implement state transition tests that assert state transition occurs:
 * for proposal when there is strong evidence of prepare.
 * for base when there is no strong support for any proposal
 * for base by timeout alone for a sole participant.

Part of #9

### DIFF
--- a/gpbft/gpbft_test.go
+++ b/gpbft/gpbft_test.go
@@ -1,0 +1,222 @@
+package gpbft_test
+
+import (
+	"testing"
+
+	"github.com/filecoin-project/go-f3/emulator"
+	"github.com/filecoin-project/go-f3/gpbft"
+)
+
+var (
+	tipset0 = gpbft.TipSet{Epoch: 0, Key: []byte("bigbang")}
+	tipSet1 = gpbft.TipSet{Epoch: 1, Key: []byte("fish")}
+	tipSet2 = gpbft.TipSet{Epoch: 2, Key: []byte("lobster")}
+)
+
+func TestGPBFT_WithEvenPowerDistribution(t *testing.T) {
+	newInstanceAndDriver := func(t *testing.T) (*emulator.Instance, *emulator.Driver) {
+		driver := emulator.NewDriver(t)
+		instance := emulator.NewInstance(t,
+			0,
+			gpbft.PowerEntries{
+				gpbft.PowerEntry{
+					ID:    0,
+					Power: gpbft.NewStoragePower(1),
+				},
+				gpbft.PowerEntry{
+					ID:    1,
+					Power: gpbft.NewStoragePower(1),
+				},
+			},
+			tipset0, tipSet1, tipSet2,
+		)
+		driver.Start(instance)
+		return instance, driver
+	}
+
+	t.Run("Decides proposal on strong quorum", func(t *testing.T) {
+		instance, driver := newInstanceAndDriver(t)
+		driver.RequireQuality()
+		driver.RequireNoBroadcast()
+
+		driver.RequireDeliverMessage(&gpbft.GMessage{
+			Sender: 1,
+			Vote:   instance.NewQuality(instance.Proposal()),
+		})
+		driver.RequirePrepare(instance.Proposal())
+		driver.RequireDeliverMessage(&gpbft.GMessage{
+			Sender: 1,
+			Vote:   instance.NewPrepare(0, instance.Proposal()),
+		})
+
+		evidenceOfPrepare := driver.NewJustification(0, gpbft.PREPARE_PHASE, instance.Proposal(), 0, 1)
+		driver.RequireCommit(0, instance.Proposal(), evidenceOfPrepare)
+		driver.RequireDeliverMessage(&gpbft.GMessage{
+			Sender:        1,
+			Vote:          instance.NewCommit(0, instance.Proposal()),
+			Justification: evidenceOfPrepare,
+		})
+
+		evidenceOfCommit := driver.NewJustification(0, gpbft.COMMIT_PHASE, instance.Proposal(), 0, 1)
+		driver.RequireDecide(instance.Proposal(), evidenceOfCommit)
+		driver.RequireDeliverMessage(&gpbft.GMessage{
+			Sender:        1,
+			Vote:          instance.NewDecide(0, instance.Proposal()),
+			Justification: evidenceOfCommit,
+		})
+
+		driver.RequireDecision(instance.Proposal())
+	})
+
+	t.Run("Decides base on lack of quorum", func(t *testing.T) {
+		instance, driver := newInstanceAndDriver(t)
+		driver.RequireQuality()
+		driver.RequireNoBroadcast()
+
+		baseChain := instance.Proposal().BaseChain()
+		alternativeProposal := baseChain.Extend([]byte("barreleye"))
+
+		driver.RequireDeliverMessage(&gpbft.GMessage{
+			Sender: 1,
+			Vote:   instance.NewQuality(alternativeProposal),
+		})
+		driver.RequireDeliverAlarm()
+		driver.RequirePrepare(baseChain)
+		driver.RequireDeliverMessage(&gpbft.GMessage{
+			Sender: 1,
+			Vote:   instance.NewPrepare(0, baseChain),
+		})
+		evidenceOfPrepare := driver.NewJustification(0, gpbft.PREPARE_PHASE, baseChain, 0, 1)
+
+		driver.RequireCommit(0, baseChain, evidenceOfPrepare)
+		driver.RequireDeliverMessage(&gpbft.GMessage{
+			Sender:        1,
+			Vote:          instance.NewCommit(0, baseChain),
+			Justification: evidenceOfPrepare,
+		})
+
+		evidenceOfCommit := driver.NewJustification(0, gpbft.COMMIT_PHASE, baseChain, 0, 1)
+		driver.RequireDecide(baseChain, evidenceOfCommit)
+		driver.RequireDeliverMessage(&gpbft.GMessage{
+			Sender:        1,
+			Vote:          instance.NewDecide(0, baseChain),
+			Justification: evidenceOfCommit,
+		})
+		driver.RequireDecision(baseChain)
+	})
+
+	t.Run("Converges on base when quorum not possible", func(t *testing.T) {
+		instance, driver := newInstanceAndDriver(t)
+		driver.RequireQuality()
+		driver.RequireNoBroadcast()
+
+		baseChain := instance.Proposal().BaseChain()
+		alternativeProposal := baseChain.Extend([]byte("barreleye"))
+
+		driver.RequireDeliverMessage(&gpbft.GMessage{
+			Sender: 1,
+			Vote:   instance.NewQuality(alternativeProposal),
+		})
+		driver.RequireDeliverAlarm()
+		driver.RequirePrepare(baseChain)
+		driver.RequireDeliverMessage(&gpbft.GMessage{
+			Sender: 1,
+			Vote:   instance.NewPrepare(0, alternativeProposal),
+		})
+
+		driver.RequireCommitForBottom(0)
+		driver.RequireDeliverMessage(&gpbft.GMessage{
+			Sender: 1,
+			Vote:   instance.NewCommit(0, gpbft.ECChain{}),
+		})
+
+		evidenceOfCommitForBottom := driver.NewJustification(0, gpbft.COMMIT_PHASE, gpbft.ECChain{}, 0, 1)
+
+		driver.RequireConverge(1, baseChain, evidenceOfCommitForBottom)
+		driver.RequireDeliverMessage(&gpbft.GMessage{
+			Sender:        1,
+			Vote:          instance.NewConverge(1, baseChain),
+			Justification: evidenceOfCommitForBottom,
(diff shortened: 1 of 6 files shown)


gpbft/gpbft_test.go
```diff
@@ -0,0 +1,197 @@
...
+	), progressTimeout, "DECIDE with evidence of COMMIT for base not reached")
+	em.RequestBroadcastFrom(1, instance.NewMessageBuilder(gpbft.Payload{
+		Step:  gpbft.DECIDE_PHASE,
+		Value: baseChain,
+	}, em.GetLastSubjectBroadcast().Justification, true))
+
+	em.NextUntil(em.HasDecided(instance.ID, baseChain), progressTimeout, "Did not decide on base")
+}
+
+func TestGPBFT_SoleParticipantDecidesOnBaseByTimeout(t *testing.T) {
```
anorth, line 154: There aren't really any timeouts here. A participant observing all of the power vote for something should immediately progress to the next step in every case (and we want the test to show that no alarm was needed).

anorth (commented):
Just noting a couple of things ahead of us discussing in person.

gpbft/emulator/emulator.go
```diff
@@ -0,0 +1,153 @@
...
+	return e.bus.Len() > 0
+}
+
+func (e *Emulator) NextUntil(match Matcher, timeout time.Duration, msgAndArgs ...any) {
+	if match() {
+		return
+	}
+
+	timer := time.NewTimer(timeout)
+	defer func() { timer.Stop() }()
```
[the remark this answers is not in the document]
Kubuxu: I think it primarily prevents a no-progress from stopping the test execution without relying on the global test timeout.

gpbft/gpbft_test.go
```diff
@@ -0,0 +1,197 @@
...
+	"time"
+
+	"github.com/filecoin-project/go-f3/certs"
+	"github.com/filecoin-project/go-f3/gpbft"
+	"github.com/filecoin-project/go-f3/gpbft/emulator"
+	"github.com/stretchr/testify/require"
+)
+
+func TestGPBFT_DecidesProposalOnStrongEvidenceOfPrepare(t *testing.T) {
+	const progressTimeout = 2 * time.Second
```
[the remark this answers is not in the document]
Kubuxu: As I mentioned above, but @masih should confirm, it is a "test should finish in 2 seconds, or it is a failure".

gpbft/emulator/emulator.go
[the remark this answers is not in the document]
masih: @anorth it asserts that the condition occurs within a given timeout. 

> a worrying sign

Why? go tests have a default timeout of 10 minutes for a test to pass. There should be _some_ timeout when testing anything. Here, the code asserts that a condition becomes true within a given amount of time.

gpbft/gpbft_test.go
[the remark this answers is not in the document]
masih: See above.

gpbft/gpbft_test.go
masih: Right, the name of the test is misleading. 

> we want the test to show that no alarm was needed

That's _a_ scenario yes but this test should pass even with alarm triggering. Because we cannot guarantee that (or rather the specification does not strictly require that) alarms arrive after broadcasts to self or vice versa.

gpbft/emulator/emulator.go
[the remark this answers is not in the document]
anorth: We discussed this in person, and the main reason it was worrying to me is because of what it implied about how the tests were run: non-deterministically. I think we've agreed to change that.

But for posterity I'll note here that I would still try to avoid timeouts and make the tests clearly pass or fail. Is it _really_ any number of messages before the condition is met? Can we not deterministically detect a lack of progress? The simulator-based tests have far less control over how the participant are driven yet they don't need a timeout.

A timeout either signals that a test is flaky, or makes it so if it wasn't already. It's an assertion about hardware and environment. Any timeout low enough to be useful (i.e. signal "we broke something if this is exceeded") is also low enough to be accidentally exceeded in the wrong environment. So even if the test was deterministic, the timeout adds a spurious failure mode. Go's 10-minute default has some utility in preventing an infinite loop from locking up your CI machine for hours or days, but it's set to a ridiculous high value. I wouldn't have objected so much to a 10-minute timeout here either. But 2 seconds appears to be attempting to set an actual expectation about how long the code should run for. I don't think one should ever try to code "test should finish in 2 seconds, or it is a failure". (Benchmarks are a different thing, and use benchmark tools, not timeouts either).

gpbft/gpbft_test.go
anorth: I agree these are two different scenarios, both of which should be tested. (In the alarm case, the node must still wait for the message in phases except quality and converge).

gpbft/emulator/emulator.go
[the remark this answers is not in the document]
masih: Thank you for capturing this, Alex.

I agree with the sentiment of the comment above. However, I continue to believe that in this case timeout is relevant. It is a scaled down version of the Go's default 10m timeout, offering sufficient headroom for me to be confident that it will not impact the test flakiness. Because, `Next` is sufficiently fast to result in thousands of iterations (i.e. broadcasts or alarm triggers) per second, and that means it should be a long enough timeout in any reasonable environment despite the "2s" itself being a short period of time. 

Based on what we discussed this timeout will go away from the PR in any case since the automatic dispatch of events via bus in random order will be removed from the PR.

gpbft/emulator/emulator.go
```diff
@@ -0,0 +1,93 @@
...
+// Returns the final built and delivered message along with any errors that may
+// occur as a result of message delivery. If the building of the message fails
+// nil message is returned with error.
+func (e *Emulator) ReceiveMessage(from gpbft.ActorID, mb *gpbft.MessageBuilder) (*gpbft.GMessage, error) {
+	if from == e.host.id {
+		return nil, fmt.Errorf("subject impersonation is not supported")
+	}
+	mb.SetNetworkName(e.host.NetworkName())
+	mb.SetSigningMarshaler(e.host.adhocSigning)
+	msg, err := mb.Build(e.host.adhocSigning, from)
```
anorth, line 67: I was expecting to make assertions directly on the thing the participant sent to us: the MessageBuilder. Initially I was confused why this is being built into a message first. I since realised it's for delivery of messages back to the participant. If you didn't do this, I think you could remove even the CRC signing and replace the host signature verification calls with simple mocks that just return true/false (and maybe check arguments, for one or two tests).

gpbft/gpbft_test.go
```diff
@@ -0,0 +1,457 @@
...
+	require.NoError(err)
+
+	// Start the instance and trigger the first alarm because the GPBFT
+	// implementation in this repo implicitly uses the alarm to actually start the
+	// instance.
+	require.NoError(em.Start(instance))
+	require.AlarmReceived(em.ReceiveAlarm())
+
+	// Expect QUALITY step from the emulator.
+	require.Quality(require.MessageReceived(em.ReceiveBroadcast()), instance)
```
anorth, line 45: Interesting, this is somewhat inverted from what I expected. I expected:
1. Assert that QUALITY was sent
2. Deliver a QUALITY message

This is kind of inferring that a message was sent by delivering an unknown message that the participant sent earlier and then checking that it exists and was QUALITY. It works, but this scheme will always deliver participants own message first, and seems awkward to re-order.

gpbft/gpbft_test.go
```diff
@@ -0,0 +1,457 @@
...
+	)
+
+	// Expect PREPARE step with the original instance proposal, since strong quorum
+	// of senders is met during QUALITY phase.
+	receivedPrepare := require.MessageReceived(em.ReceiveBroadcast())
+	require.Prepare(receivedPrepare, instance)
+
+	// Send PREPARE message to the emulator for the instance proposal which should
+	// facilitate evidence of strong quorum of PREPARE.
+	sentPrepare := require.MessageReceived(
```
anorth, line 59: Received/sent names are confusing here. receivedPrepare is the on the participant sent, not the one it received from someone else (I think).

gpbft/emulator/assertions.go
```diff
@@ -0,0 +1,131 @@
...
+	r.Zero(subject.Vote.Round)
+	r.Equal(vote, subject.Vote.Value)
+	r.Equal(instance.id, subject.Vote.Instance)
+	r.Equal(instance.supplementalData, subject.Vote.SupplementalData)
+	r.Equal(justification, subject.Justification)
+	r.Empty(subject.Ticket)
+}
+
+// Consensus asserts that the given instance has reached consensus for the expected chain.
+func (r *Assertions) Consensus(instance *Instance, expect gpbft.ECChain) {
```
anorth, line 127: "Consensus" isn't the right term for a single instance. I suggest `Decision` or `Decided`.

anorth (approved):
Thanks, this is better. It's still doing more than I expected, but I don't think there's anything wrong, and it should be able to exercise all the paths we're going for. More helpers will be able to make the test more concise in cases where we're not trying to check everything at every step, but move the participant quickly into some state to then check a particular transition.

gpbft/emulator/chain.go
```diff
@@ -0,0 +1,51 @@
+package emulator
```
anorth, line 1: So I suggest moving this out of the gpbft package and as a peer of `sim`. If we were using Rust crates or some other decent dependency/package system, one would want to be able to depend on the protocol code without necessarily pulling this in.

anorth:
To check I wasn't missing some essential complication, I hacked up some of the directions I suggested above in https://github.com/filecoin-project/go-f3/pull/385. I think a little more change is needed before I'd consider it a good final shape, namely restoring some of the removed assertions in place of comparing MessageBuilder equality.

Feel free to take or leave any of those ideas as you wish. I think they simplify things a bit, but I'll continue to approve your current direction anyway.

gpbft/emulator/host.go
```diff
@@ -0,0 +1,56 @@
...
+
+import (
+	"time"
+
+	"github.com/filecoin-project/go-f3/gpbft"
+)
+
+var _ gpbft.Host = (*host)(nil)
+
+type host struct {
```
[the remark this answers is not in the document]
masih: The term "mock" has a contextual meaning in code which I don't think alines with what this struct does. "fake" is maybe a better term.

gpbft/emulator/emulator.go
```diff
@@ -0,0 +1,93 @@
...
+func (e *Emulator) ReceiveAlarm() (bool, error) {
+	if e.host.maybeTriggerAlarm() {
+		return true, e.subject.ReceiveAlarm()
+	}
+	return false, nil
+}
+
+// ReceiveMessage delivers a message to the emulated participant from the given
+// sender. The sender must not be the ID of the honest participant as
+// impersonation of messages to the participant is not supported.
```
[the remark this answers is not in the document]
masih: > Why this restriction?

My thought was to avoid emulating unrealistic scenarios, where a tester (or attacker if you like) can sign messages on behalf of the honest participant. We might want to test message replay, but I think that's different from impersonation.

> drop this, probably by removing `host.id` entirely.

Although it is true that a participant does not know its own ID, but the concept of ID is (for better or worse) present across the core implementation, used as a way to index received messages, etc. 

It is also the way the Power Table is indexed. From the emulation's perspective we want to be able to send a message on behalf of a node with specific storage power. Hence the presence of ID.

gpbft/emulator/emulator.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	return msg, e.doBroadcast(msg)
+}
+
+// ReceiveBroadcast delivers one pending broadcast, if any, requested by the
+// emulated participant to itself. Returns the final built and delivered message
+// along with any errors that may occur as part of message delivery. If there are
+// no pending broadcasts returns nil as both the message and error.
+func (e *Emulator) ReceiveBroadcast() (*gpbft.GMessage, error) {
+	if len(e.host.pendingBroadcasts) != 0 {
```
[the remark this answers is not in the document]
masih: > you are doing something with these received messages

My take is that any lower level emulation goes into the realms of what the existing `MockHost` is for. At the emulation level my aim is to test GPBFT while obeying the rules of the game: sign messages as dictated by the signer, progress state transition as the FIP dictates, avoid impersonating the participant itself, and so on.

gpbft/gpbft_test.go
masih: Thanks for pointing that out. I'll see what can do. 

The way this makes sense in my head is that the writer of the test is the "subject" of `sent` `received` naming prefix of the variables. For example, `sentPrepare` means the test sent a prepare message. The test can never send a message on behalf of the participant. It can receive it (by delivering a self-broadcast). Hence the `recievedPrepare` naming.

gpbft/gpbft_test.go
masih: >It works, but this scheme will always deliver participants own message first, and seems awkward to re-order.

Happy to add inspection of what the honest participant requested the broadcast of before delivering it. I don't think it makes much difference because we assert that what we delivered was what we expected to deliver immediately after it is delivered.

PR #362, opened 2024-06-21
Consolidate signing into one package

masih:
The signing implementation in simulation seem to have become useful across production scope code, among other places. Instead of having a split singing package, one at top level only containing BLS and one at `sim` mixing it all consolidate them both.

Move tests as necessary and split test files as placeholder for writing more tests. Simplify the test suit boilerplate that now has become redundant because of the new simpler package structure.


Stebalien:
These packages have different goals, from what I can tell. The `sim/signing` package exists to manage and generate temporary keys.

Honestly, I expect a lot of this will change once we move most of the signing to the client. But @Kubuxu might be able to shed more light on that.

masih:
> These packages have different goals, from what I can tell. The sim/signing package exists to manage and generate temporary keys.

I more or less agree. The current state of `main` however uses Fake signer (in `sim/signing`) across other production scope packages.

Would separating "generation of temporary keys" from the rest and leaving it in `sim` unblock this PR @Stebalien ?

Stebalien:
My understanding is that the current situation is temporary and that `main.go` will drop the dependency, right?

masih:
Ok

Kubuxu:
Yes, the main will drop the dependency in the future. We still might want to use a fake signer for wider integration tests, though.

masih:
This is a quick refactor. Seems like it's best to to reopen it once there is a clearer idea on where we stand with main branch.

PR #375, opened 2024-06-25
feat: use the scaled power for winning tickets

Stebalien:
This brings us in-line with the FIP.


Stebalien:
Note: we can't throw away the unscaled powers because we need them for:

1. Re-calculating power scaling (unfortunately, due to how our tests are written, we need the ability to _add_ entries to the power table).
2. Commit to the power table, etc.

PR #377, opened 2024-06-27
CertStore: Implement a friendlier subscribe method

Stebalien:
I ended up not needing this but I wanted to get some feedback on it as it's significantly friendlier than the current `SubscribeForNewCerts` method. Basically, this one _doesn't_ force the subscriber to implement resubscribe/catchup logic manually.


Kubuxu:
I feel that in most cases the cert subscriber doesn't care about the complete sequence of certs, only the most recent one. I would for sure keep this branch around though as the code might be useful in the future.

Stebalien:
Yeah, that's where I ended up. However, light clients (and lotus APIs in general) will likely want this kind of API.

Stebalien:
Closing in favor of https://github.com/filecoin-project/go-f3/pull/646. Every time I've wanted to subscribe for certificates, I've only cared about the latest. Especially because certificates are usually stored inside the certstore anyways.

PR #382, opened 2024-06-27
Correctly use ECPeriod & ECDelay

Stebalien:
ECDelay is how long we wait before finalizing a tipset, but it isn't necessarily the time between instances (it's actually 2x right now in lotus). ECPeriod is the time between instances.

The new logic waits `ECDelay + ECPeriod * multiplier` where I've reduced each multiplier by 1 to account for the added delay.

Additionally:

1. We now use `now+ECPeriod` when we fail to get the tipset from lotus.
2. ECDelay defaults to 60s (same as lotus).

NOTE: I'm not sure if this new logic is correct, but the old logic definitely had some issues so I'm opening this PR for discussion.


masih (approved):

masih:
Thank you for this Steven, this looks good. Though I'd like to take fresh look at it again tomorrow.

Stebalien:
Fix for the failed test: https://github.com/filecoin-project/go-f3/pull/384

PR #385, opened 2024-06-28
Demo removing signing, message-building and semi-automated delivery own messages

anorth:
Demo some of the things I discussed in #359 to check for myself that they make sense.

Don't merge this PR - take or leave whatever you want from it.


gpbft/emulator/emulator.go
```diff
@@ -0,0 +1,100 @@
...
+// Emulator.ReceiveBroadcast or Emulator.DeliverAlarm.
+type Driver struct {
+	subject *gpbft.Participant
+	host    *mockHost
+	require *Assertions
+}
+
+// New instantiates a new Emulator with the given GPBFT options. See
+// Emulator.Start.
+func New(assertions *Assertions, o ...gpbft.Option) *Driver {
```
masih, line 22: May be a YAGNI but the reason i did not do this was to avoid having emulator package depend on `testing`. This is so that the emulator can be used at production level code in the future for non-testing scenarios: for example using emulator as a light-weight tool to research core protocol improvements. That would separate the emulated behaviour from what it considers "correct". 

If YAGNI, then I would take `testing.T` here straight up.

gpbft/gpbft_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+	em := emulator.New(require)
+
+	em.Start(instance)
+
+	// Expect subject to emit QUALITY
+	proposal := instance.GetProposal()
+	// TODO: replace verifyBroadcast with VerifyQuality, VerifyPrepare etc that
+	// have similar inspections to those removed from Assertions
+	em.VerifyBroadcast(instance.NewQuality(proposal))
```
masih, line 42: This style is much more readable than mine; thanks Alex

masih (commented):
Many thanks for taking the time to put this together. I'll cherry pick thing from it onto the original PR 🙏

gpbft/emulator/emulator.go
anorth: I think I would use the simulator for protocol R&D (that's what it was first built for). This is for unit tests.

masih: You are probably right, thanks

PR #391, opened 2024-07-01
F3-390: Pause callback for dyncamic manifest

### LINKED ISSUE #390: Pause F3 until a new configuration manifest is received
We should include in the manifest a flag to say F3 to pause until the next manifest with a new configuration is received. While F3 is paused, the runner and the pubsub topic for F3 will be disabled, and the the node will only be listening to the reception of new manifests.

//cc @masih

adlrocha:
Closes #390 

This PR:
* Adds a `Pause` flag to dynamic manifest to pause F3 when a new manifest with the flag enabled is received.
* Simplifies the locking strategy so the certstore and the runner share a common lock instead of using independent ones (they are used in the same code paths, when initializing a node, and triggering a configuration change by a dynamic manifest). 
* Slightly refactors the callback logic to make it more readable.
* Make `message_builder` thread-safe to avoid some races that were appearing since the message signature was exposed externally


Kubuxu:
> Make message_builder thread-safe to avoid some races that were appearing since the message signature was exposed externally

I'm not sure why this would happen but it is a solution for sure. :P

Kubuxu (approved):
SGWM

adlrocha:
> I'm not sure why this would happen but it is a solution for sure. :P

Unfortunately, it happened :( and the race detector kept complaining haha

PR #414, opened 2024-07-06
Test rebroadcast and fix rebroadcast reset issue

masih:
Fix a bug where rebroadcast parameters are never reset within a round, because of the way by which "progress" was determined: The previous implementation stored last seen round/phase on broadcast and compared it to the current round to infer whether progress has been made. This approach would always result in no progress since on phase/round change
the broadcast state would update the references based on the last broadcast. Therefore the current round/phase always matched the last seen round/phase in broadcast state.

The changes here resolve the issue by explicitly resetting the rebroadcast params upon progress by explicitly resetting rebroadcast params on phase change. Based on a few retries this makes up the most readable code IMHO.

The changes also add a test that assert the right set of messages are broadcasted relative to phase and at only once enough alarms are triggered.

Part of #9

### DIFF
--- a/emulator/driver.go
+++ b/emulator/driver.go
@@ -1,11 +1,17 @@
 package emulator
 
 import (
+	"bytes"
+	"context"
 	"testing"
 
 	"github.com/filecoin-project/go-f3/gpbft"
 	"github.com/stretchr/testify/require"
 )
+
+// ValidTicket is a sentinel value to generate and set a valid gpbft.Ticket when
+// delivering messages via Driver.deliverMessage.
+var ValidTicket gpbft.Ticket = []byte("filled in by driver")
 
 // Driver drives the emulation of a GPBFT instance for one honest participant,
 // and allows frame-by-frame control over progress of a GPBFT instance.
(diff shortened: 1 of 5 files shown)


Kubuxu:
```
--- FAIL: TestGPBFT_WithEvenPowerDistribution (0.02s)
    --- FAIL: TestGPBFT_WithEvenPowerDistribution/Rebroadcasts_selected_messages_on_timeout (0.00s)
        driver_assertions.go:54: 
            	Error Trace:	/home/runner/work/go-f3/go-f3/emulator/driver_assertions.go:54
            	            				/home/runner/work/go-f3/go-f3/emulator/driver_assertions.go:46
            	            				/home/runner/work/go-f3/go-f3/gpbft/gpbft_test.go:335
            	Error:      	Not equal: 
            	            	expected: 0x3
            	            	actual  : 0x2
            	Test:       	TestGPBFT_WithEvenPowerDistribution/Rebroadcasts_selected_messages_on_timeout
FAIL
```

Stebalien (approved):

PR #416, opened 2024-07-06
Improve lifecycle management

Stebalien:
1. Merge the client and the runner. The distinction was unclear and the client/runner/module tended to reach into each other. This change merges the client/runner and then separates the new "runner" from the module as much as possible.
2. Completely stop/discard the runner when rebootstrapping. The new logic carefully waits for all components to stop before moving on.
3. Simplify locking and make sure we take the locks where appropriate.
4. Merge bootstrap and re-configure logic. The dynamic manifest client no longer cares about _when_ a manifest should be applied, it simply gives it to the module (F3) and let's F3 us its normal bootstrap logic.

Finally, I've improved the tests to:

1. Always on exit (checking for errors).
2. Never fail from goroutines.
3. Correctly wait for manifest changes (previously, it would wait for at least one node to change manifests).

NOTE: This removes the ability to reconfig without rebootstrap, but preserves the ability to _pause_ without rebootstrap.


cmd/f3/manifest.go
```diff
@@ -171,15 +174,15 @@ var manifestServeCmd = cli.Command{
 		}
 
 		manifestPath := c.String("manifest")
-		loadManifestAndVersion := func() (manifest.Manifest, manifest.Version, error) {
+		loadManifestAndVersion := func() (*manifest.Manifest, manifest.Version, error) {
```
Stebalien, line 177: I changed manifests to by-reference because they can get a big large.

cmd/f3/manifest.go
```diff
@@ -200,46 +203,49 @@ var manifestServeCmd = cli.Command{
...
+		errgrp, ctx := errgroup.WithContext(c.Context)
+		errgrp.Go(func() error { return sender.Run(ctx) })
+		errgrp.Go(func() error {
+			checkTicker := time.NewTicker(checkInterval)
+			defer checkTicker.Stop()
+
+			for ctx.Err() == nil {
+				select {
+				case <-ctx.Done():
+					return nil
```
Stebalien, line 217: not an error

cmd/f3/manifest.go
```diff
@@ -200,46 +203,49 @@ var manifestServeCmd = cli.Command{
 		}
 		_, _ = fmt.Fprintf(c.App.Writer, "Started manifest sender with version: %s\n", manifestVersion)
 
-		go func() {
-			sender.Start(c.Context)
-		}()
```
Stebalien, line 205: Renamed to "run" and managed with an errorgroup.

cmd/f3/run.go
```diff
@@ -121,20 +121,18 @@ var runCmd = cli.Command{
...
-		mprovider.SetManifestChangeCallback(f3.ManifestChangeCallback(module))
 		go runMessageSubscription(ctx, module, gpbft.ActorID(id), signingBackend)
 
-		return module.Run(ctx)
+		if err := module.Start(ctx); err != nil {
+			return nil
+		}
+		<-ctx.Done()
+		return module.Stop(context.Background())
 	},
```
Stebalien, line 131: There may be better ways to do this?

f3.go
```diff
@@ -1,141 +1,70 @@
...
+	runner   *gpbftRunner
 }
 
 // New creates and setups f3 with libp2p
 // The context is used for initialization not runtime.
 // signingMarshaller can be nil for default SigningMarshaler
-func New(ctx context.Context, manifest manifest.ManifestProvider, ds datastore.Datastore, h host.Host,
+func New(_ctx context.Context, manifest manifest.ManifestProvider, ds datastore.Datastore, h host.Host,
 	ps *pubsub.PubSub, verif gpbft.Verifier, ec ec.Backend, log Logger, signingMarshaller gpbft.SigningMarshaler) (*F3, error) {
-	ds = namespace.Wrap(ds, manifest.Manifest().DatastorePrefix())
```
Stebalien, line 112: We do this per-manifest-version now.

f3.go
```diff
@@ -1,141 +1,70 @@
...
+	runningCtx context.Context
+	cancelCtx  context.CancelFunc
+	errgrp     *errgroup.Group
 
-func (mc *client) Logger() Logger {
-	return mc.logger
+	mu       sync.Mutex
+	cs       *certstore.Store
+	manifest *manifest.Manifest
+	runner   *gpbftRunner
```
Stebalien, line 44: I store these separately because:

1. The certificate-store remains even when the runner is stopped (paused).
2. The manifest can is set before bootstrap (if available).
3. The runner is unset whenever we pause.

host.go
```diff
@@ -53,81 +97,89 @@ func newRunner(m manifest.ManifestProvider, client *client) (*gpbftRunner, error
...
+				if err := h.participant.ReceiveAlarm(); err != nil {
+					return err
+				}
+			case msg, ok := <-messageQueue:
+				if !ok {
+					return xerrors.Errorf("incoming message queue closed")
+				}
+				if err := h.participant.ReceiveMessage(msg); err != nil {
+					// XXX: Log only?
+					h.log.Debugf("error when processing message: %+v", err)
```
Stebalien, line 173: Is this a hard error?

host.go
```diff
@@ -53,81 +97,89 @@ func newRunner(m manifest.ManifestProvider, client *client) (*gpbftRunner, error
...
+				if !ok {
+					finalityCertificates = make(chan *certs.FinalityCertificate, 4)
+					c, _ = h.certStore.SubscribeForNewCerts(finalityCertificates)
+				}
+				if err := h.receiveCertificate(c); err != nil {
+					return err
+				}
+			case <-h.alertTimer.C:
+				if err := h.participant.ReceiveAlarm(); err != nil {
+					return err
```
Stebalien, line 165: Is this a hard error?

host.go
Stebalien: I changed it to a log because it didn't look like it.

manifest/dynamic_manifest.go
```diff
@@ -31,169 +35,156 @@ type DynamicManifestProvider struct {
...
+	cpy.NetworkName = gpbft.NetworkName(newName)
+	return &cpy
 }
 
-// When a manifest configuration changes, a new network name
-// is set that depends on the manifest version of the previous version to avoid
-// overlapping previous configurations.
-func (m *DynamicManifestProvider) networkNameOnChange() gpbft.NetworkName {
-	return gpbft.NetworkName(string(m.manifest.NetworkName) + "/" + fmt.Sprintf("%d", m.manifest.Sequence))
+func (m ManifestUpdateMessage) Marshal() ([]byte, error) {
```
Stebalien, line 66: I had to switch to update messages to pull the version out of the manifest (where, IMO, it doesn't really belong). Honestly, I'm not sure if we should even _have_ an explicit version or if we should just use message sequence numbers and version by hashing...

manifest/dynamic_manifest.go
```diff
@@ -31,169 +35,156 @@ type DynamicManifestProvider struct {
...
+	initialManifest *Manifest
+	manifestChanges chan *Manifest
 }
 
-func (m *DynamicManifestProvider) Manifest() Manifest {
-	m.lk.RLock()
-	defer m.lk.RUnlock()
-	return m.manifest
+type ManifestUpdateMessage struct {
+	MessageSequence uint64
```
Stebalien, line 48: I added a separate message sequence number.

1. Message sequence numbers order messages.
2. Versions only change when reconfiguring/rebootstrapping.

manifest/dynamic_manifest.go
```diff
@@ -31,169 +35,156 @@ type DynamicManifestProvider struct {
...
-		select {
-		case <-ctx.Done():
-			break loop
+	manifestSub, err := manifestTopic.Subscribe()
+	if err != nil {
+		return xerrors.Errorf("subscribing to manifest pubsub topic: %w", err)
+	}
+
+	// XXX: load the initial manifest from disk!
+	// And save it!
```
Stebalien, line 123: We need to do this or nothing will work.

host.go
```diff
@@ -53,81 +97,89 @@ func newRunner(m manifest.ManifestProvider, client *client) (*gpbftRunner, error
...
+					finalityCertificates = make(chan *certs.FinalityCertificate, 4)
+					c, _ = h.certStore.SubscribeForNewCerts(finalityCertificates)
+				}
+				if err := h.receiveCertificate(c); err != nil {
+					return err
+				}
+				continue
+			case <-h.alertTimer.C:
+				if err := h.participant.ReceiveAlarm(); err != nil {
+					return err
```
Stebalien, line 147: ignore this error?

host.go
Stebalien: No, we definitely want to ignore this.

f3.go
```diff
@@ -1,141 +1,70 @@
...
-	}
-
-	m := F3{
-		ds:     ds,
-		host:   h,
-		pubsub: ps,
-		ec:     ec,
-		log:    log,
-
-		client: &client{
```
Stebalien, line 128: Removed.

f3.go
```diff
@@ -145,412 +78,197 @@ func New(ctx context.Context, manifest manifest.ManifestProvider, ds datastore.D
...
+				}
+
+				pendingManifest = update
+				if delay, err := m.computeBootstrapDelay(update); err != nil {
+					return err
+				} else if delay > 0 {
+					manifestChangeTimer.Reset(delay)
+					continue
+				}
+			case <-manifestChangeTimer.C:
```
Stebalien, line 213: We should have coverage for these... something is wrong.

host.go
Stebalien: I'm pretty sure this is not a hard error and it's something we should test.

manifest/dynamic_manifest.go
```diff
@@ -31,169 +35,156 @@ type DynamicManifestProvider struct {
...
-					log.Errorf("error fetching chain head: %+v", err)
-					m.lk.Unlock()
-					continue
-				}
+func NewDynamicManifestProvider(initialManifest *Manifest, pubsub *pubsub.PubSub, ec ec.Backend, manifestServerID peer.ID) ManifestProvider {
+	ctx, cancel := context.WithCancel(context.Background())
+	errgrp, ctx := errgroup.WithContext(ctx)
 
-				// if the upgrade epoch is reached or already passed.
-				if ts.Epoch() >= m.nextManifest.BootstrapEpoch {
```
Stebalien, line 101: We now handle this delay with the rest of the bootstrap logic in the `gpbftRunner`.

manifest/manifest.go
```diff
@@ -39,18 +39,15 @@ var (
 	}
 )
 
-type OnManifestChange func(ctx context.Context, prevManifest Manifest, errCh chan error)
+type OnManifestChange func(ctx context.Context, prevManifest Manifest) error
 
 type ManifestProvider interface {
-	// Run starts any background tasks required for the operation
-	// of the manifest provider.
-	Run(context.Context, chan error)
```
Stebalien, line 47: Renamed to start because it doesn't block.

Stebalien: I also removed the error channels here and everywhere else because there was no way to use them reliably. Specifically, there was no way to know if we had received all the errors so we couldn't receive an error then exit. And if we ever stopped reading on the channel, various components would block.

I guess one solution would have been to write to this error channel with a context... but I just got rid of it and switched to error groups instead.

manifest/manifest.go
```diff
@@ -39,18 +39,15 @@ var (
...
-	// Returns the list of gpbft options to be used for gpbft configuration
-	GpbftOptions() []gpbft.Option
```
Stebalien, line 49: Racy with manifest changes.

manifest/manifest.go
```diff
@@ -39,18 +39,15 @@ var (
...
-	// Set callback to trigger to apply new manifests from F3.
-	SetManifestChangeCallback(OnManifestChange)
-	// Manifest accessor
-	Manifest() Manifest
```
Stebalien, line 53: Not particularly useful because this manifest will likely disagree with the manifest being used by F3 at the moment.

manifest/manifest.go
```diff
@@ -39,18 +39,15 @@ var (
...
+type OnManifestChange func(ctx context.Context, prevManifest Manifest) error
 
 type ManifestProvider interface {
-	// Run starts any background tasks required for the operation
-	// of the manifest provider.
-	Run(context.Context, chan error)
-	// Returns the list of gpbft options to be used for gpbft configuration
-	GpbftOptions() []gpbft.Option
-	// Set callback to trigger to apply new manifests from F3.
-	SetManifestChangeCallback(OnManifestChange)
```
Stebalien, line 51: Switched to a channel.

manifest/manifest.go
```diff
@@ -74,21 +71,12 @@ type EcConfig struct {
 	CommiteeLookback         uint64
 }
 
-// Manifest identifies the specific configuration for
-// the F3 instance currently running.
+// Manifest identifies the specific configuration for the F3 instance currently running.
 type Manifest struct {
-	// Sequence number of the manifest.
-	// This is used to identify if a new config needs to be applied
-	Sequence uint64
```
Stebalien, line 82: Moved to a `ManifestUpdateMessage` struct.

manifest/manifest.go
```diff
@@ -74,21 +71,12 @@ type EcConfig struct {
...
 	// Initial instance to used for the f3 instance
 	InitialInstance uint64
 	// BootstrapEpoch from which the manifest should be applied
 	BootstrapEpoch int64
-	// Flag to determine if the peer should rebootstrap in this configuration
-	// change at BootstrapEpoch
-	ReBootstrap bool
-	// Specifies if the manifest should pause F3 until a new configuration arrives.
-	Pause bool
```
Stebalien, line 91: Pause is represented by a "nil" manifest in a manifest update struct.

manifest/manifest.go
```diff
@@ -74,21 +71,12 @@ type EcConfig struct {
...
-	// Sequence number of the manifest.
-	// This is used to identify if a new config needs to be applied
-	Sequence uint64
 	// Initial instance to used for the f3 instance
 	InitialInstance uint64
 	// BootstrapEpoch from which the manifest should be applied
 	BootstrapEpoch int64
-	// Flag to determine if the peer should rebootstrap in this configuration
-	// change at BootstrapEpoch
-	ReBootstrap bool
```
Stebalien, line 89: Removed because there was no safe way to synchronize non-rebootstrapping reconfigs.

test/f3_test.go
```diff
@@ -31,134 +32,89 @@ const (
 var log = logging.Logger("f3-testing")
 
 func TestSimpleF3(t *testing.T) {
-	ctx := context.Background()
 	env := newTestEnvironment(t, 2, false)
 
-	env.Connect(ctx)
-	env.run(ctx)
-	env.waitForInstanceNumber(ctx, 5, 10*time.Second, false)
-	env.stop()
```
Stebalien, line 40: stops automatically via `t.Cleanup`.

test/f3_test.go
```diff
@@ -31,134 +32,89 @@ const (
 var log = logging.Logger("f3-testing")
 
 func TestSimpleF3(t *testing.T) {
-	ctx := context.Background()
 	env := newTestEnvironment(t, 2, false)
 
-	env.Connect(ctx)
-	env.run(ctx)
```
Stebalien, line 38: renamed to `start` because it doesn't block.

test/f3_test.go
```diff
@@ -31,134 +32,89 @@ const (
 var log = logging.Logger("f3-testing")
 
 func TestSimpleF3(t *testing.T) {
-	ctx := context.Background()
 	env := newTestEnvironment(t, 2, false)
 
-	env.Connect(ctx)
```
Stebalien, line 37: renamed to `connectAll` to make it slightly more clear what this does.

test/f3_test.go
```diff
@@ -31,134 +32,89 @@ const (
...
-	env.run(ctx)
-	env.waitForInstanceNumber(ctx, 5, 10*time.Second, false)
```
Stebalien, line 39: The context is now internalized in `env` (and it gets canceled when the test ends).

test/f3_test.go
```diff
@@ -31,134 +32,89 @@ const (
...
-	require.True(t, env.nodes[0].CurrentGpbftInstance(t, ctx) < prevInstance)
-	env.waitForInstanceNumber(ctx, 3, 15*time.Second, false)
+	require.True(t, env.nodes[0].currentGpbftInstance() < prevInstance)
+	env.waitForInstanceNumber(3, 15*time.Second, false)
 	require.NotEqual(t, prev, env.nodes[0].f3.Manifest())
 	env.requireEqualManifests(false)
-	env.stop()
-}
-
-func TestDynamicManifest_WithoutRebootstrap(t *testing.T) {
```
Stebalien, line 84: Changing the manifest without rebootstrapping is no longer supported.

test/f3_test.go
```diff
@@ -253,54 +215,65 @@ func (e *testEnv) addPowerDeltaForParticipants(ctx context.Context, m *manifest.
...
 	}
 }
 
 // waits for all nodes to reach a specific instance number.
 // If the `strict` flag is enabled the check also applies to the non-running nodes
-func (e *testEnv) waitForInstanceNumber(ctx context.Context, instanceNumber uint64, timeout time.Duration, strict bool) {
+func (e *testEnv) waitForInstanceNumber(instanceNumber uint64, timeout time.Duration, strict bool) {
 	require.Eventually(e.t, func() bool {
-		reached := 0
-		for i := 0; i < len(e.nodes); i++ {
```
Stebalien, line 266: Refactored to be less complex.

test/f3_test.go
```diff
@@ -253,54 +215,65 @@ func (e *testEnv) addPowerDeltaForParticipants(ctx context.Context, m *manifest.
...
+		return true
 	}, timeout, e.manifest.ECPeriod)
 }
 
-func (e *testEnv) waitForManifestChange(prev manifest.Manifest, timeout time.Duration, nodes []*testNode) {
+func (e *testEnv) waitForManifestChange(prev *manifest.Manifest, timeout time.Duration) {
 	require.Eventually(e.t, func() bool {
-		reached := 0
-		for i := 0; i < len(e.nodes); i++ {
-			for i := 0; i < len(nodes); i++ {
```
Stebalien, line 287: This double-loop was incorrect. We'd count each node `len(e.nodes)` times.

f3.go
```diff
@@ -145,412 +78,197 @@ func New(ctx context.Context, manifest manifest.ManifestProvider, ds datastore.D
...
-		}
-		// Check context before starting the instance
-		select {
-		case <-ctx.Done():
-			// If the context is cancelled before starting, return immediately
-			errCh <- ctx.Err()
-			return
-		default:
-			// Proceed with running the instance
-			err := m.runner.Run(startInstance, ctx)
```
Stebalien, line 299: This was racy. Given a quick re-config (e.g., catching up with reconfigs), we could start with the wrong instance if two copies of this function ended up running at the same time.

f3.go
```diff
@@ -145,412 +78,197 @@ func New(ctx context.Context, manifest manifest.ManifestProvider, ds datastore.D
...
-	ts, err := m.ec.GetTipsetByEpoch(ctx, m.Manifest().BootstrapEpoch-m.Manifest().ECFinality)
-	if err != nil {
-		return xerrors.Errorf("getting initial power tipset: %w", err)
+	if cs == nil {
+		return nil, xerrors.Errorf("F3 is not running")
 	}
+	return cs.Get(ctx, instance)
+}
 
-	initialPowerTable, err := m.ec.GetPowerTable(ctx, ts.Key())
```
Stebalien, line 347: This needed to apply the power table delta (that's why I'm now wrapping the `ec.Backend`).

f3.go
```diff
@@ -145,412 +78,197 @@ func New(ctx context.Context, manifest manifest.ManifestProvider, ds datastore.D
...
 // Or if there is an error from the message handling routines.
-func (m *F3) Run(ctx context.Context) (_err error) {
-	ctx, m.cancelCtx = context.WithCancel(ctx)
-	defer m.cancelCtx()
-
-	runnerErrCh := make(chan error, 1)
-	manifestErrCh := make(chan error, 1)
-
-	// bootstrap runner for the initial manifest
-	go m.startGpbftRunner(ctx, runnerErrCh)
```
Stebalien, line 387: Racy. Could have been delayed and/or interleaved with re-config.

f3.go
```diff
@@ -145,412 +78,197 @@ func New(ctx context.Context, manifest manifest.ManifestProvider, ds datastore.D
...
-		_err = multierr.Append(_err, teardownErr)
-	}()
-
-	select {
-	case <-ctx.Done():
-		return nil
-	case err := <-runnerErrCh:
-		return err
-	case err := <-manifestErrCh:
+func (m *F3) Start(startCtx context.Context) (_err error) {
```
Stebalien, line 161: So:

- I like the start pattern because I can easily wait and return errors for setup issues.
- I don't like the start pattern because I get no notifications that something went wrong.

I'm thinking we should add a `Wait` to all of these services? Or something like that?

Stebalien: Basically:

```go
func (m *F3) WaitStop(ctx context.Context) error {
    select {
    case <-m.runningCtx.Done(): // should exit when everything is done.
    case <-ctx: // hm... return immediately? but that's a bit confusing.
    }
    return m.Close()
}
```

f3.go
```diff
@@ -145,412 +78,197 @@ func New(ctx context.Context, manifest manifest.ManifestProvider, ds datastore.D
...
-		m.runner.Stop()
+	initialDelay, err := m.computeBootstrapDelay(pendingManifest)
+	if err != nil {
+		return err
 	}
-	m.runner = nil
-}
 
-func (m *F3) handleIncomingMessages(ctx context.Context, queue chan gpbft.ValidatedMessage) {
-	ctx, m.client.incomingCancel = context.WithCancel(ctx)
```
Stebalien, line 429: In a goroutine and not in a lock, so setting `client.incomingCancel` is racy.

f3.go
```diff
@@ -145,412 +78,197 @@ func New(ctx context.Context, manifest manifest.ManifestProvider, ds datastore.D
...
-			// for now we just log the error and continue.
-			// This is not critical, but alternative approaches welcome.
-			m.log.Errorf("error stopping gpbft runner: %+v", err)
+	m.errgrp.Go(func() error {
+		manifestChangeTimer := time.NewTimer(initialDelay)
+		if pendingManifest == nil && !manifestChangeTimer.Stop() {
+			<-manifestChangeTimer.C
 		}
-		// empty message queue from outstanding messages
-		m.emptyMessageQueue()
```
Stebalien, line 471: I re-create the message queue each time. This was also a bit racy as we could ahve processed one more message after closing pubsub.

f3.go
```diff
@@ -145,412 +78,197 @@ func New(ctx context.Context, manifest manifest.ManifestProvider, ds datastore.D
...
-	m.mLk.Lock()
-	m.log.Infof("triggering manifest (seq=%d) change with rebootstrap", m.Manifest().Sequence)
-
-	// if runner still up
-	if m.runner != nil {
-		m.runner.Stop()
-	}
-
-	// clear the certstore
-	if err := m.certStore.DeleteAll(ctx); err != nil {
```
Stebalien, line 497: We now use a new namespace each time. I figure:

1. We'll clear the old datastore when we turn everything on for real.
2. We _should_ use a separate top-level `/f3` datastore to avoid polluting the other lotus datastores.

host.go
```diff
@@ -53,81 +93,91 @@ func newRunner(m manifest.ManifestProvider, client *client) (*gpbftRunner, error
...
+					// could easily receive these.
+					h.log.Debugf("error when processing message: %+v", err)
+				}
+			case <-h.runningCtx.Done():
+				return nil
 			}
-			if err := h.participant.ReceiveMessage(msg); err != nil {
-				return err
-			}
-		case start := <-h.client.manifestUpdate:
```
Stebalien, line 118: Manifest updates now always trigger a complete restart of the runner.

cmd/f3/run.go
Kubuxu: I think this is the best way to do it.

f3.go
```diff
@@ -1,141 +1,74 @@
...
+	cs       *certstore.Store
+	manifest *manifest.Manifest
+	runner   *gpbftRunner
 }
 
 // New creates and setups f3 with libp2p
 // The context is used for initialization not runtime.
 // signingMarshaller can be nil for default SigningMarshaler
-func New(ctx context.Context, manifest manifest.ManifestProvider, ds datastore.Datastore, h host.Host,
+func New(_ctx context.Context, manifest manifest.ManifestProvider, ds datastore.Datastore, h host.Host,
```
Kubuxu, line 50: I guess we want to keep this for metrics and such?

f3.go
```diff
@@ -145,412 +78,197 @@ func New(ctx context.Context, manifest manifest.ManifestProvider, ds datastore.D
...
-	m.msgSub, err = m.client.topic.Subscribe()
-	if err != nil {
-		m.psLk.Unlock()
-		return xerrors.Errorf("subscribing to topic: %w", err)
-	}
-	m.psLk.Unlock()
-	return nil
-}
+func (m *F3) GetLatestCert(ctx context.Context) (*certs.FinalityCertificate, error) {
+	m.mu.Lock()
```
Kubuxu, line 110: I think we can just use `defer` here, it is fast enough and non-blocking in any way.

Kubuxu: Although we might want to keep it like this for uniformity.

f3.go
```diff
@@ -145,412 +78,197 @@ func New(ctx context.Context, manifest manifest.ManifestProvider, ds datastore.D
...
 	}
+	return cs.Get(ctx, instance)
+}
 
-	initialPowerTable, err := m.ec.GetPowerTable(ctx, ts.Key())
-	if err != nil {
-		return xerrors.Errorf("getting initial power table: %w", err)
+// Returns the time at which the F3 instance specified by the passed manifest should be started, or
+// 0 if the passed manifest is nil.
+func (m *F3) computeBootstrapDelay(manifest *manifest.Manifest) (delay time.Duration, err error) {
```
Kubuxu, line 133: is the named return needed? The function name specifies that it is a delay.

f3.go
```diff
@@ -145,412 +78,197 @@ func New(ctx context.Context, manifest manifest.ManifestProvider, ds datastore.D
...
+	}
+
+	// We bootstrap now if we're at the correct time, whether or
+	// not we've hit the correct epoch. The important thing is
+	// that:
+	//
+	// 1. All nodes will use BootstrapEpoch - Finality to pick the base.
+	// 2. All nodes will bootstrap at the same time.
+	currentEpoch := ts.Epoch()
+	if currentEpoch < manifest.BootstrapEpoch {
```
Kubuxu, line 151: Maybe invert the if: return 0 `if currentEpoch >= manifest.BoostrapEpoch` then the more complex chunk is outside of if

f3.go
Kubuxu: yeha, I'm not sure, Run flow also has its issues

host.go
Kubuxu: I don't think we can, I think error here would suggest that gpbft is in bad state.

host.go
Kubuxu: I'm assuming you mean `don't want to ignore it`

host.go
Kubuxu: I think checking for PanicErr would be worth it

host.go
```diff
@@ -53,81 +93,91 @@ func newRunner(m manifest.ManifestProvider, client *client) (*gpbftRunner, error
...
+			case msg, ok := <-messageQueue:
+				if !ok {
+					return xerrors.Errorf("incoming message queue closed")
+				}
+				if err := h.participant.ReceiveMessage(msg); err != nil {
+					// We silently drop failed messages because GPBFT will
+					// return errors for, e.g., messages from old instances.
+					// Given the async nature of our pubsub message handling, we
+					// could easily receive these.
+					h.log.Debugf("error when processing message: %+v", err)
```
Kubuxu, line 171:
```suggestion
					h.log.Errorf("error when processing message (downgrade if too spammy): %+v", err)
```

host.go
Stebalien: Ah, wait, I meant to put that comment on the other case (receive message). Yeah, I don't _think_ we want to ignore this one. But I'm not actually sure... we should audit it later.

manifest/dynamic_manifest.go
Kubuxu: Add comment on the Sequence vs Version

Kubuxu (commented):
Directionally LGTM

f3.go
Stebalien: Even better, `func Done() <-chan struct{}`.

f3.go
Stebalien: Also if we need to do anything that might block in the future.

f3.go
Stebalien: Yeah, I mostly wanted to be consistent.

f3.go
Stebalien: No, I named it because the old function name was... not very clear.

Stebalien: (my old function name, I first named this "scheduleBootstrap", or something like that.

host.go
Stebalien: We'll do this in a followup.

Stebalien: https://github.com/filecoin-project/go-f3/issues/421

Kubuxu (approved):

Kubuxu (approved):

PR #419, opened 2024-07-06
Remove custom CI jobs in faviour of Unified CI

masih:
Remove test, test with race, generate checker and lint in favour of standard unified CI jobs to only contain fuzzing.


Kubuxu (approved):

masih:
OK i think we are missing [this](https://github.com/ipdxco/unified-github-workflows/pull/71)

masih:
I have manually enabled merge_group trigger until it is set at upstream workflows.

PR #424, opened 2024-07-07
Hash ticket before performing arithmetic

Kubuxu:
To ensure no arithmetic properties leak from BLS-G2 affine compressed
representation.


masih (approved):

anorth:
@Kubuxu could you please submit a PR to the FIP specifying this?

Kubuxu:
The way of going from ticket to ordering is not specified in the FIP, which is why I didn't submit it initially. I can submit it, though.

PR #427, opened 2024-07-08
Import go-log directly (except in GPBFT) and remove Logging

Stebalien:
It was causing circular import issues (which could have been fixed by moving it, I guess) but, more importantly, the abstraction really wasn't pulling its weight. We're keeping it out of GPBFT itself but our other packages already depend on things like go-datastore and go-libp2p, so depending on go-log doesn't really matter.

### DIFF
--- a/certexchange/polling/common.go
+++ b/certexchange/polling/common.go
@@ -5,5 +5,5 @@ import (
 	logging "github.com/ipfs/go-log/v2"
 )
 
-var log = logging.Logger("f3-certexchange")
+var log = logging.Logger("f3/certexchange")
 var clk clock.Clock = clock.New()
--- a/certexchange/polling/common.go
+++ b/certexchange/polling/common.go
@@ -2,6 +2,8 @@ package polling
 
 import (
 	"github.com/benbjohnson/clock"
+	logging "github.com/ipfs/go-log/v2"
 )
 
+var log = logging.Logger("f3-certexchange")
 var clk clock.Clock = clock.New()
--- a/certexchange/server.go
+++ b/certexchange/server.go
@@ -16,7 +16,7 @@ import (
 	"github.com/libp2p/go-libp2p/core/network"
 )
 
-var log = logging.Logger("f3-certexchange")
+var log = logging.Logger("f3/certexchange")
 
 const maxResponseLen = 256
 
--- a/certexchange/server.go
+++ b/certexchange/server.go
@@ -8,12 +8,15 @@ import (
 	"runtime/debug"
 	"time"
 
-	"github.com/filecoin-project/go-f3"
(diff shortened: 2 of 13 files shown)


certexchange/polling/common.go
```diff
@@ -2,6 +2,8 @@ package polling
 
 import (
 	"github.com/benbjohnson/clock"
+	logging "github.com/ipfs/go-log/v2"
 )
 
+var log = logging.Logger("f3-certexchange")
```
Kubuxu, line 8: same system name as main cert exchange, do we want to make it different?

Stebalien: Nah, it'll just use the same logger.

host.go
```diff
@@ -32,7 +31,7 @@ type gpbftRunner struct {
 	signingMarshaller gpbft.SigningMarshaler
 	verifier          gpbft.Verifier
 	broadcastCb       BroadcastMessage
-	log, logWithSkip  Logger
+	logWithSkip       Logger
```
Kubuxu, line 34: delete

logging.go
```diff
@@ -0,0 +1,17 @@
+package f3
+
+import (
+	"github.com/filecoin-project/go-f3/gpbft"
+	logging "github.com/ipfs/go-log/v2"
+)
+
+var log = logging.Logger("f3")
+var tracer gpbft.Tracer = (*gpbftTracer)(log)
```
Kubuxu, line 9: I think make a separate logger for this, also it should have a skip option set

certexchange/polling/common.go
Kubuxu, line 8:
```suggestion
var log = logging.Logger("f3/certexchange")
```

cmd/f3/run.go
```diff
@@ -21,9 +21,9 @@ import (
 	"golang.org/x/xerrors"
 )
 
-const DiscoveryTag = "f3-standalone"
+var log = logging.Logger("f3-cli")
```
Kubuxu, line 24:
```suggestion
var log = logging.Logger("f3/cli")
```

certexchange/server.go
```diff
@@ -8,13 +8,16 @@ import (
...
-	"github.com/filecoin-project/go-f3"
 	"github.com/filecoin-project/go-f3/certstore"
 	"github.com/filecoin-project/go-f3/gpbft"
+
+	logging "github.com/ipfs/go-log/v2"
 	"github.com/libp2p/go-libp2p/core/host"
 	"github.com/libp2p/go-libp2p/core/network"
 )
 
+var log = logging.Logger("f3-certexchange")
```
Kubuxu, line 19:
```suggestion
var log = logging.Logger("f3/certexchange")
```

Kubuxu (approved):

PR #430, opened 2024-07-08
Wait for cleanup on certexchange server shutdown

Stebalien:


Kubuxu (approved):

certexchange/server.go
```diff
@@ -111,9 +115,34 @@ func (s *Server) handleRequest(ctx context.Context, stream network.Stream) (_err
 
 // Start the server.
 func (s *Server) Start() error {
+	s.runningLk.Lock()
+	defer s.runningLk.Unlock()
```
Kubuxu, line 119: Can this race with StreamHanlder?

Kubuxu (approved):

certexchange/server.go
Stebalien: If you stop/restart, there's technically a race here and not much we can do about it.

I'll make it impossible to re-use the server. That'll at least prevent the worst-case where we receive a request on the wrong protocol.

Stebalien: Ah, wait. No, it's not racy.

1. The handler uses a context captured from `Start`, not stored on the `Server`.
2. We cancel the context before removing the stream handler.

When the stream handler takes the read lock, we're either:

1. After `Stop`: We'll see that the context has been canceled.
2. Before `Stop`: We'll hold the read lock and prevent `Stop` from returning.

Issue #160, opened 2024-04-15
Observablility and metrics


Kubuxu:
We should have metrics around round progression and the percentage of power voted (it might be tricky because we move on once a quorum is reached).
OpenCensus is the standard for it in the ecosystem.

- [x] signature validatiton time 
- [x] CPU & memory usage 
- [x] datastore size

jennijuju:
add monitoring for metadata size growth

masih:
@jennijuju Could you point me exactly to what we want to measure as "metadata size growth"?

My initial take is: size growth of every datastore namespace used by F3. Does that sound right?

Stebalien:
@masih honestly, we can just monitor the "metadata" datastore itself. IIRC, that datastore is usually pretty much empty so it should be mostly F3 stuff.

masih:
The only remaining item here is metadata size, for which I am not sure if Lotus metrics are exposed. I have posted something on fil-lotus-dev and pending response. 

For the rest the dashboard is updated to include them in addition to in-depth validation breakdown time, divergence, cert exchange, cert polling, gossipsub metrics.

PR #438, opened 2024-07-09
Integrate metrics for core GPBFT protocol

masih:
Introduce otel metrics to measure:
* number/rate of phase transitions
* histogram over the number of rounds per instance
* number of broadcasted messages
* number of re-broadcasted messages
* number of re-broadcast attempts

Part of #160


Kubuxu:
Will this work with the TODO context?
Do we want to add instance number labels?

masih:
It will work. In most cases the is no context to use and todo seems more meaningful than background. Though I have no strong opinions. 


Re labelling by instance id, my take is that we should not. That would cause explosive cardinality of metrics.

Kubuxu (approved):

Kubuxu:
Can you also add metric for: instance started, instance finished?

masih:
> Can you also add metric for: instance started, instance finished?

I had that initially but I think we can infer that using phase counter, because it is tagged by phase: increment of `INITIAL` and `TERMINATE` should indicate the instance start/termination. WDYT, does that make sense?

Kubuxu:
That will give metrics relative to node start. I'm looking for global `instance # started` so we could see outliers.

masih:
Sounds good; I'll add that 👍

Kubuxu:
Yeah, I expect them to be off by one in most cases but if one node falls out of line, the state (whether it is off by one or not) tells us where things failed (failed to start or failed to terminate).

Issue #421, opened 2024-07-07
Improve ReceiveMessage error handling


Stebalien:
Right now, `ReceiveMessage` returns both "fatal" errors (e.g., panics) and non-fatal errors (messages from the wrong instance). We should log non-errors internally and only return "real" errors.

This hasn't been an issue in testing because we're not using real pubsub with proper async message delivery. But it will be on mainnet.

PR #442, opened 2024-07-10
Skip messages from the wrong instance instead of returning an error

Stebalien:
Due to async message delivery, we'll frequently get messages from the wrong instance. Log (trace) and drop them.

Part of #421.

### DIFF
--- a/gpbft/participant.go
+++ b/gpbft/participant.go
@@ -161,8 +161,9 @@ func (p *Participant) ReceiveMessage(vmsg ValidatedMessage) (err error) {
 
 	// Drop messages for past instances.
 	if msg.Vote.Instance < p.currentInstance {
-		return fmt.Errorf("message %d, current instance %d: %w",
-			msg.Vote.Instance, p.currentInstance, ErrValidationTooOld)
+		p.tracer.Log("dropping message from old instance %d while received in instance %d",
+			msg.Vote.Instance, p.currentInstance)
+		return nil
 	}
 
 	// If the message is for the current instance, deliver immediately.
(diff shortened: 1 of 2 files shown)


gpbft/participant.go
```diff
@@ -161,8 +161,9 @@ func (p *Participant) ReceiveMessage(vmsg ValidatedMessage) (err error) {
 
 	// Drop messages for past instances.
 	if msg.Vote.Instance < p.currentInstance {
-		return fmt.Errorf("message %d, current instance %d: %w",
-			msg.Vote.Instance, p.currentInstance, ErrValidationTooOld)
+		p.tracer.Log("dropping message from old instance %d while received in instance %d",
```
masih, line 164: Do we want to signal this to upstream instead by non-critical error typing?

Stebalien: > Do we want to signal this to upstream instead by non-critical error typing?

We could add proper level-based logging, is that what you're suggesting?

masih: That too. What I was suggesting is introducing a `RetriableError` type which the caller can check if a returned error is a type of. So we continue to return error but type it based on whether it is retriable, i.e. non-critical, or not. The integrator then chooses to do whatever they wish with it.

masih: This is a non-blocker btw; we can add this later if upstream actually finds it useful.

Stebalien:
Ah, well, I need to fix the tests...

Kubuxu (approved):
LGTM but tests

PR #445, opened 2024-07-11
Broadcast messages via a blocking channel

### LINKED ISSUE #443: Rebroadcast overloads the broadcast bus
Rebroadcasting writes a bunch of messages to the broadcast bus all at once, which causes subscribers to be disconnected after they fall behind.

We probably shouldn't use this "bus" in this case but should instead directly write to a channel. Then, we can handle the "broadcast" part in the client.

Stebalien:
Otherwise rebroadcast will end up dropping the last and most important messages when the message bus blocks.

fixes #443

### DIFF
--- a/cmd/f3/run.go
+++ b/cmd/f3/run.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"math/big"
 	"os"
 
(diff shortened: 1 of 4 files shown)


cmd/f3/run.go
```diff
@@ -120,47 +121,46 @@ var runCmd = cli.Command{
...
 		}
-		<-ctx.Done()
+		select {
+		case err := <-errCh:
+			if err != nil {
+				log.Error(err)
+			}
+		case <-ctx.Done():
+		}
 		return module.Stop(context.Background())
```
Stebalien, line 137: drive-by improvement

test/f3_test.go
```diff
@@ -323,12 +324,17 @@ func (e *testEnv) waitForManifestChange(prev *manifest.Manifest, timeout time.Du
...
+	env := &testEnv{t: t, errgrp: grp, testCtx: ctx, net: mocknet.New()}
+
+	// Cleanup on exit.
 	env.t.Cleanup(func() {
 		cancel()
+		for _, n := range env.nodes {
+			require.NoError(env.t, n.f3.Stop(context.Background()))
+		}
 		require.NoError(env.t, env.errgrp.Wait())
 	})
```
Stebalien, line 339: This wasn't needed by the final version of this patch, but it was for some earlier iterations and, in general, is less likely to deadlock on exit. Basically, it ensures that we stop all F3 nodes:

1. Only once.
2. Before waiting.

test/f3_test.go
```diff
@@ -439,9 +446,6 @@ func (e *testEnv) resumeNode(i int) {
 func (e *testEnv) startNode(i int) {
 	n := e.nodes[i]
 	require.NoError(e.t, n.f3.Start(e.testCtx))
-	e.t.Cleanup(func() {
```
Stebalien, line 442: Moved.

Kubuxu:
Tests are failing

f3.go
```diff
@@ -68,15 +68,11 @@ func New(_ctx context.Context, manifest manifest.ManifestProvider, ds datastore.
...
-// After perparing inputs and signing over them, Broadcast should be called.
-//
-// If the passed channel is full at any point, it will be dropped from subscription and closed.
-// To stop subscribing, either the closer function can be used, or the channel can be abandoned.
-// Passing a channel multiple times to the Subscribe function will result in a panic.
-func (m *F3) SubscribeForMessagesToSign(ch chan<- *gpbft.MessageBuilder) (closer func()) {
-	_, closer = m.busBroadcast.Subscribe(ch)
-	return closer
+// MessageStoSign returns a channel of outbound messages that need to be signed by the client(s).
+// - The same channel is shared between all callers and will never be closed.
```
Kubuxu, line 72: Sharing this channel is bad (you don't know which messages you will get) but I don think we need to protect against it.

Kubuxu (approved):

f3.go
Stebalien: I think it's fine if it's a part of a documented part of the API.

PR #452, opened 2024-07-11
Improve test parallelism

Stebalien:
The main slowdown is due to the main "F3" tests, but we might as well run tests in parallel wherever possible. I excluded the certificate exchange tests as parallelizing them actually slowed everything down a bit.


Stebalien:
Down from 8m -> 5m30s, ish.

masih (approved):

adlrocha (approved):

PR #462, opened 2024-07-12
Gracefully handle EC diverging from F3

Kubuxu:
We will end up proposing base which will trigger backoff, giving us chance to re-boostrap,
Also added a metric for it.

### DIFF
--- a/ec/ec.go
+++ b/ec/ec.go
@@ -26,4 +26,7 @@ type TipSet interface {
 	Beacon() []byte
 	Epoch() int64
 	Timestamp() time.Time
+
+	// User readable name of the tipset
+	String() string
 }
--- a/ec/ec.go
+++ b/ec/ec.go
@@ -2,6 +2,7 @@ package ec
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/filecoin-project/go-f3/gpbft"
@@ -22,11 +23,10 @@ type Backend interface {
 }
 
 type TipSet interface {
+	fmt.Stringer
+
 	Key() gpbft.TipSetKey
 	Beacon() []byte
 	Epoch() int64
 	Timestamp() time.Time
-
-	// User readable name of the tipset
-	String() string
 }
--- a/ec/fake_ec.go
+++ b/ec/fake_ec.go
@@ -2,7 +2,6 @@ package ec
 
 import (
 	"context"
-	"encoding/base64"
 	"encoding/binary"
 	"fmt"
 	"sync"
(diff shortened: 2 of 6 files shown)


Kubuxu:
For some reason tests fail with what looks like OOM.

ec/ec.go
```diff
@@ -26,4 +26,7 @@ type TipSet interface {
 	Beacon() []byte
 	Epoch() int64
 	Timestamp() time.Time
+
+	// User readable name of the tipset
+	String() string
```
Stebalien, line 31: nit: put `fmt.Stringer` at the top.

ec/fake_ec.go
```diff
@@ -54,6 +55,14 @@ func (ts *Tipset) Timestamp() time.Time {
 	return ts.timestamp
 }
 
+func (ts *Tipset) String() string {
+	res := base64.RawStdEncoding.EncodeToString(ts.tsk[:gpbft.CidMaxLen])
```
Stebalien, line 59: I'd check the length before slicing.

Stebalien: Honestly, if we're going to stringify the tipset, I'd prefer to just stringify it as proper CIDs.

host.go
```diff
@@ -361,20 +361,20 @@ func (h *gpbftHost) collectChain(base ec.TipSet, head ec.TipSet) ([]ec.TipSet, e
...
-			// we reorged away from base
-			// scream and panic??
-			// TODO make sure this is correct, re-boostrap/manifest swap code has to be able to
-			// catch it
-			panic("reorg-ed away from base, dunno what to do, reboostrap is the answer")
+	current := head
+	for !bytes.Equal(current.Key(), base.Key()) {
+		if current.Epoch() < base.Epoch() {
+			metrics.headDiverged.Add(h.runningCtx, 1)
+			log.Errorw("reorg-ed away from base, reboostrap is the answer",
```
Stebalien, line 368: This is going to _majorly_ annoy users.

Kubuxu: It will happen only every minute or so, but agreed, I will downgrade to Warn and tone it down

ec/fake_ec.go
Kubuxu: If the length is wrong, then it means someone messed with it. The FakeEC will always create proper length, the test is there so if anyone changes it, it gets detected.

I'm importing multibase, at this point we might import go-cid but I'm not touching that right now.

Stebalien (approved):

PR #464, opened 2024-07-12
Implement additional message validation unit tests

masih:
Implement message validation unit tests to hit specific justification conditions.

Part of #9


Kubuxu (approved):
I don't love relying on string values of errors but I guess for tests it is fine.

masih:
> I don't love relying on string values of errors but I guess for tests it is fine.

I don't think there is any other way to assert validation error is hitting what it should (other than defining a ton of sentinel errors)

PR #467, opened 2024-07-12
Fix typo in manifest EC config

masih:
Fix minor typo in spelling of "committee".


Kubuxu (approved):
It will need to get fixed in the infra as well

masih:
> It will need to get fixed in the infra as well

Thanks for pointing that out Kuba, done.

PR #471, opened 2024-07-13
Pass context variables to runningCtx

Kubuxu:


Stebalien (changes requested):
NO!

Stebalien (approved):
Ok, fine. "without cancel".

OK!

PR #472, opened 2024-07-13
Remove manifest versions

### LINKED ISSUE #392: Write Ahead Log & Rebroadcast
We've implemented limited rebroadcasting to help lagging/restarting nodes catch up. However, if the 66%+ of the network crashes after starting an instance but before sending a single decide message, the network could decide on two different values for the same instance.

A simple solution here is write-ahead logging. That is:

1. Before sending any message (maybe limit it to commit messages?) log/sync the message to disk. To save space, we could just record a single message template.
2. On restart, re-load all (or maybe the last round? commits only?) messages from the last instance started but with no decision.
3. Rebroadcast those messages and resume from that point.

Of course, nothing will help if the actual _disks_ die. But this will at least help us recover in case someone finds a way to crash the entire network all at once.

The specific attack I'm worried about is as follows:

1. An attacker listens for commit messages and waits until they see a quorum (enough to reach a "decision").
2. The attacker checks to see if it knows of a _better_ tipset at the same height (more weight). E.g., the attacker may choose to withhold a block to make this happen.
3. The attacker then uses some previously unknown exploit to crash all lotus nodes.
4. The attacker submits a certificate for the "forgotten" decision to some bridge.
5. The network restarts/resumes.
6. The network agrees on a _different_ value.
7. The bridge is now borked.

### LINKED ISSUE #468: Manifest server does not handle sequence numbers and versions
Sequence numbers and versions were moved outside of the manifest, Manifest server should handle them.

Stebalien:
This also allows us to re-config without re-bootstrap (by restarting the local F3 instance). This is only safe for non-consensus parameters and will be horribly unsafe until we fix 392.

Also:

- Removes reliance on hashing json, relies on the network name and manual equality checks.
- Removes versions. We now expect the version to be explicitly specified in the network name.
- Starts message sequence numbers at the current time so we don't need to save them.
- Remove the EC from the dynamic manifest provider as it's unused.
- Tests.

fixes #468

### DIFF
--- a/cmd/f3/manifest.go
+++ b/cmd/f3/manifest.go
@@ -173,20 +173,7 @@ var manifestServeCmd = cli.Command{
 		}
 
 		manifestPath := c.String("manifest")
-		loadManifestAndVersion := func() (*manifest.Manifest, manifest.Version, error) {
-
-			m, err := loadManifest(manifestPath)
-			if err != nil {
-				return nil, "", fmt.Errorf("loading manifest: %w", err)
-			}
-			version, err := m.Version()
-			if err != nil {
-				return nil, "", fmt.Errorf("versioning manifest: %w", err)
-			}
-			return m, version, nil
-		}
-
-		initManifest, manifestVersion, err := loadManifestAndVersion()
+		currentManifest, err := loadManifest(manifestPath)
 		if err != nil {
 			return fmt.Errorf("loading initial manifest: %w", err)
 		}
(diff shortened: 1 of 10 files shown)


cmd/f3/manifest.go
```diff
@@ -215,12 +202,12 @@ var manifestServeCmd = cli.Command{
 				case <-ctx.Done():
 					return nil
 				case <-checkTicker.C:
-					if nextManifest, nextManifestVersion, err := loadManifestAndVersion(); err != nil {
+					if nextManifest, err := loadManifest(manifestPath); err != nil {
 						_, _ = fmt.Fprintf(c.App.ErrWriter, "Failed reload manifest: %v\n", err)
-					} else if manifestVersion != nextManifestVersion {
-						_, _ = fmt.Fprintf(c.App.Writer, "Loaded manifest with version: %s\n", nextManifestVersion)
+					} else if nextManifest.NetworkName != currentManifest.NetworkName {
```
Kubuxu, line 207: this sends out the update only when network name changes

Kubuxu (approved):

PR #473, opened 2024-07-13
Introduce a fake clock to F3

Kubuxu:


Kubuxu:
no idea why some tests  are still slow-ish. They seem to send a lot of messages

certexchange/polling/common_test.go
```diff
@@ -9,19 +9,14 @@ import (
...
 // The network name used in tests.
 const TestNetworkName gpbft.NetworkName = "testnet"
 
 // The clock used in tests. Time doesn't pass in tests unless you add time to this clock.
-var MockClock = clock.NewMock()
-
-func init() {
-	clk = MockClock
-}
+//var MockClock = clock.NewMock()
```
Stebalien, line 19: removeme

internal/clock/clock.go
```diff
@@ -0,0 +1,34 @@
...
+
+type clockKeyType struct{}
+
+var clockKey = clockKeyType{}
+
+func NewMock() *Mock {
+	return clock.NewMock()
+}
+
+func WithMockClock(ctx context.Context) (context.Context, *Mock) {
```
Stebalien, line 21: Docs would be nice.

internal/clock/clock.go
```diff
@@ -0,0 +1,34 @@
...
+
+type Clock = clock.Clock
+type Mock = clock.Mock
+type Timer = clock.Timer
+
+type clockKeyType struct{}
+
+var clockKey = clockKeyType{}
+
+func NewMock() *Mock {
```
Stebalien, line 17: Is this used?

manifest/manifest_sender.go
```diff
@@ -6,6 +6,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/filecoin-project/go-f3/internal/clock"
```
Stebalien, line 9: grouping

Stebalien (approved):
My one gripe is that we usually take the clock in the constructor, except when we don't have a constructor and we take it in `Start`. But I can live with that.

internal/clock/clock.go
Kubuxu: Yes in polling tests

Stebalien: Ah, got it.

Stebalien (approved):

Stebalien:
Welp. That looks like a mock clock bug. We're blocked writing to the channel in `Tick`. I'm _guessing_ we should be selecting on default here...

But this case _shouldn't_ happen if we're resetting correctly. It looks like we're re-starting a timer before draining it.

Stebalien:
I've forked to https://github.com/filecoin-project/go-clock/ (raul's version is behind anyways).

Stebalien:
https://github.com/filecoin-project/go-clock/pull/1

Stebalien:
> But this case shouldn't happen if we're resetting correctly. It looks like we're re-starting a timer before draining it.

That was due to a second bug. Fixed in the same PR.

PR #486, opened 2024-07-19
Manifest enhancements

Kubuxu:
- Save latest manifest to datastore
- Add protocol version to manifest

### DIFF
--- a/f3.go
+++ b/f3.go
@@ -233,7 +233,7 @@ func (m *F3) Stop(stopCtx context.Context) (_err error) {
 	)
 }
 
-func (m *F3) reconfigure(ctx context.Context, manifest *manifest.Manifest) (_err error) {
+func (m *F3) reconfigure(ctx context.Context, manif *manifest.Manifest) (_err error) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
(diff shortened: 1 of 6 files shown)


Kubuxu:
The power table is still WIP

f3.go
```diff
@@ -242,14 +242,19 @@ func (m *F3) reconfigure(ctx context.Context, manifest *manifest.Manifest) (_err
...
 	}
 
-	if manifest == nil {
+	// pause if new manifest is nil
+	if manif == nil {
+		return nil
+	}
+	// pause if we are not capable of supporting this version
+	if manif.ProtocolVersion > manifest.VersionCapability {
 		return nil
```
Stebalien, line 251: I'd log something (once).

manifest/dynamic_manifest.go
```diff
@@ -98,13 +105,27 @@ func (m *DynamicManifestProvider) Start(startCtx context.Context) (_err error) {
...
+		msgSeqNumber = update.MessageSequence
+		currentManifest = update.Manifest
+	}
 
-	currentManifest := m.initialManifest
-	m.manifestChanges <- m.initialManifest
+	m.manifestChanges <- currentManifest
 
-	m.errgrp.Go(func() error {
+	m.errgrp.Go(func() (_err error) {
```
Stebalien, line 128: Oops...

f3.go
Stebalien: Also, maybe we should have an equality here?

Stebalien (approved):

f3.go
Kubuxu: this IMO gives us more flexibility

Stebalien: Eh, up to you. I guess we're unlikely to really go "backwards" anyways.

PR #487, opened 2024-07-22
Implement the "ohshitstore"

### LINKED ISSUE #479: ohshitstore: Store EC-finalized power table deltas if F3 falls behind/fails to bootstrap
If F3 falls behind or fails to bootstrap, we need to store per-epoch power table deltas in a separate store so we can have a chance of catching up if/when the splitstore garbage collects old state. Once F3 has caught up with EC finality, we can clear this separate datastore.

I'm suggesting a _separate_ datastore from the finality certificate store because:

1. We're storing deltas per-epoch instead of per-instance (not quite 1:1).
2. It avoids adding a bunch of complexity to the certificate store, exchange protocol, etc.

Stebalien:
If we fall more than 1.5 EC finality behind, this will start storing power deltas up to EC finality (checking every 1/4 EC finalities).

NOTE: this records power deltas by-epoch. That should be safe enough for now but I'd like to switch to tipset keys before mainnet launch in case.

I implemented this as a completely separate service so the runner doesn't need to know that state can be garbage collected (and so we can keep it running even if we pause GPBFT). Although... we might not want to?

fixes #479

### DIFF
--- a/internal/powerstore/powerstore.go
+++ b/internal/powerstore/powerstore.go
@@ -1,0 +1,351 @@
+package powerstore
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"fmt"
+	"slices"
+
+	"github.com/filecoin-project/go-f3/certs"
+	"github.com/filecoin-project/go-f3/certstore"
+	"github.com/filecoin-project/go-f3/ec"
+	"github.com/filecoin-project/go-f3/gpbft"
+	"github.com/filecoin-project/go-f3/internal/clock"
+	"github.com/filecoin-project/go-f3/manifest"
+
+	"github.com/ipfs/go-datastore"
+	"github.com/ipfs/go-datastore/namespace"
+	"github.com/ipfs/go-datastore/query"
+	logging "github.com/ipfs/go-log/v2"
+	"go.uber.org/zap"
+	"golang.org/x/sync/errgroup"
+)
+
+var log = logging.Logger("f3/ohshitstore")
+
+var _ ec.Backend = (*Store)(nil)
+
+type Store struct {
+	ec.Backend
+
+	ds       datastore.Datastore
+	cs       *certstore.Store
+	manifest *manifest.Manifest
+
+	clock      clock.Clock
+	runningCtx context.Context
+	cancel     context.CancelFunc
+	errgrp     *errgroup.Group
+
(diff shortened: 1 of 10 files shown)


internal/powerstore/powerstore.go
```diff
@@ -0,0 +1,350 @@
...
+
+		if err := ps.ds.Delete(ctx, datastore.NewKey(r.Key)); err != nil {
+			log.Errorw("failed to delete power-table diffs to delete", "error", err)
+			return
+		}
+	}
+}
+
+func (ps *Store) run(ctx context.Context) error {
+	ticker := ps.clock.Ticker(ps.manifest.ECPeriod * time.Duration(max(ps.manifest.ECFinality/4, 1)))
```
Kubuxu, line 203: This might have weird interactions when a node is catching up. I suggest we do it more frequently; there is no harm in it, really.

internal/powerstore/powerstore.go
```diff
@@ -0,0 +1,350 @@
...
+	retryTimer := ps.clock.Timer(0)
+	defer ticker.Stop()
+	defer retryTimer.Stop()
+
+	startThreshold := max(ps.manifest.ECFinality*3/2, 1)
+	stopThreshold := max(ps.manifest.ECFinality/2, 1)
+
+	var initialized bool
```
Kubuxu, line 211: How does this get reset?

Kubuxu (commented):
Seems good, I will take one more look at it

internal/powerstore/powerstore.go
Stebalien: https://github.com/filecoin-project/go-f3/blob/ea515f8bb5e4e5065b6ad7c0250b8d4a5bc58904/internal/powerstore/powerstore.go#L233

internal/powerstore/powerstore.go
Stebalien: Fair enough. I can also make it faster when we're behind.

Stebalien: Eh, I just made it every 2-ec.

internal/powerstore/powerstore.go
Stebalien: Oh. Fixed.

Kubuxu (approved):

Kubuxu (approved):

PR #488, opened 2024-07-22
Try to fetch inital power table if specified

### LINKED ISSUE #477: Late-start with no state
@ZenGround0 ran into an issue where F3 failed to load state on start. @rvagg pointed out that this was likely because F3 reached back too far (after the splitstore deleted state).

This likely happened because calibnet failed to start F3 due to insufficient power. Restarting the node attempted to re-bootstrap F3 and failed because the power table had been garbage collected.

We need to solve this in a couple of ways:

1. For all nodes participating in F3, we cannot GC state until we have a finality certificate for that Epoch. Alternatively, we could pre-compute power tables and save them to the certificate store once we reach EC finality. This means we'll need to store power tables and power table deltas _before_ computing finality certificates, so it's a bit trickier.
2. We need to handle the late-bootstrap case. I.e., a node joins later with a snapshot. Eventually, we'll be able to solve this by shipping the latest finality certificate with that snapshot, but we can't do that till F3 actually starts working... it would be kind of nice to have a way to free-start from some epoch. I.e., take the current epoch, go back 900 epochs, ask someone for a finality certificate including that specific tipset, validate that the finality certificate has the correct power table, then start there.
3. Finally, we need to generally be robust against missing state.
```[tasklist]
### Tasks
- [ ] https://github.com/filecoin-project/go-f3/issues/479
- [ ] https://github.com/filecoin-project/go-f3/issues/485
```

Kubuxu:
Resolves #477

### DIFF
--- a/certexchange/client.go
+++ b/certexchange/client.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/filecoin-project/go-f3/certs"
 	"github.com/filecoin-project/go-f3/gpbft"
+	"github.com/filecoin-project/go-f3/internal/clock"
 
 	cid "github.com/ipfs/go-cid"
 	"github.com/libp2p/go-libp2p/core/host"
@@ -140,7 +141,9 @@ func (c *Client) Request(ctx context.Context, p peer.ID, req *Request) (_rh *Res
 	return &resp, ch, nil
 }
 
-func (c *Client) FindInitialPowerTable(ctx context.Context, powerTableCID cid.Cid) (gpbft.PowerEntries, error) {
+func FindInitialPowerTable(ctx context.Context, c Client, powerTableCID cid.Cid) (gpbft.PowerEntries, error) {
+	clk := clock.GetClock(ctx)
+
 	request := Request{
 		FirstInstance:     0,
 		Limit:             0,
@@ -151,6 +154,13 @@ func (c *Client) FindInitialPowerTable(ctx context.Context, powerTableCID cid.Ci
 		for _, p := range c.Host.Network().Peers() {
 			if ctx.Err() != nil {
 				return nil, ctx.Err()
+			}
+
+			targetProtocol := FetchProtocolName(c.NetworkName)
+			if proto, err := c.Host.Peerstore().FirstSupportedProtocol(p, targetProtocol); err != nil &&
+				proto != targetProtocol {
+
+				continue
 			}
 
 			rh, _, err := c.Request(ctx, p, &request)
@@ -165,8 +175,16 @@ func (c *Client) FindInitialPowerTable(ctx context.Context, powerTableCID cid.Ci
 			}
 			if !bytes.Equal(ptCID, powerTableCID.Bytes()) {
 				log.Infow("peer returned missmatching power table", "peer", p)
+				continue
 			}
 			return rh.PowerTable, nil
+		}
+
+		log.Infof("could not find anyone with initial power table, retrying after sleep")
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		case <-clk.After(1 * time.Second):
 		}
 	}
 }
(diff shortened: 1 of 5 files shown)


certexchange/client.go
```diff
@@ -136,3 +139,34 @@ func (c *Client) Request(ctx context.Context, p peer.ID, req *Request) (_rh *Res
...
+
+func (c *Client) FindInitialPowerTable(ctx context.Context, powerTableCID cid.Cid) (gpbft.PowerEntries, error) {
+	request := Request{
+		FirstInstance:     0,
+		Limit:             0,
+		IncludePowerTable: true,
+	}
+
+	for {
+		for _, p := range c.Host.Network().Peers() {
```
Stebalien, line 151: Check if the peer supports the protocol before asking. Otherwise, we'll block on multistream each time.

certexchange/client.go
```diff
@@ -136,3 +139,34 @@ func (c *Client) Request(ctx context.Context, p peer.ID, req *Request) (_rh *Res
...
 }
+
+func (c *Client) FindInitialPowerTable(ctx context.Context, powerTableCID cid.Cid) (gpbft.PowerEntries, error) {
+	request := Request{
+		FirstInstance:     0,
+		Limit:             0,
+		IncludePowerTable: true,
+	}
+
+	for {
```
Stebalien, line 150: Sleep.

Stebalien: Also, can we prefer peers we haven't asked yet somehow?

certexchange/client.go
```diff
@@ -136,3 +139,34 @@ func (c *Client) Request(ctx context.Context, p peer.ID, req *Request) (_rh *Res
...
+				log.Infow("requesting initial power table", "error", err, "peer", p)
+				continue
+			}
+			ptCID, err := certs.MakePowerTableCID(rh.PowerTable)
+			if err != nil {
+				log.Infow("computing iniital power table CID", "error", err)
+				continue
+			}
+			if !bytes.Equal(ptCID, powerTableCID.Bytes()) {
+				log.Infow("peer returned missmatching power table", "peer", p)
```
Stebalien, line 167: continue!

Stebalien: (I'd also test this case)

f3.go
```diff
@@ -254,8 +254,9 @@ func (m *F3) reconfigure(ctx context.Context, manif *manifest.Manifest) (_err er
 	// If we have a new manifest, reconfigure.
 	if m.manifest == nil || m.manifest.NetworkName != manif.NetworkName {
 		m.cs = nil
-		m.manifest = manif
 	}
+	// This allows for some possibly unsafe re-configuration
+	m.manifest = manif
```
Stebalien, line 259: maybe break this into a new PR? It's an important change.

store.go
```diff
@@ -25,13 +28,24 @@ func openCertstore(ctx context.Context, ec ec.Backend, ds datastore.Datastore, m
...
-		return nil, fmt.Errorf("getting initial power table: %w", err)
+	var initialPowerTable gpbft.PowerEntries
+
+	if err == nil {
+		initialPowerTable, err = ec.GetPowerTable(ctx, ts.Key())
+		if err != nil {
+			return nil, fmt.Errorf("getting initial power table: %w", err)
+		}
+	} else {
+		if m.InitialPowerTable != nil {
```
Stebalien, line 39:
```suggestion
	} else if m.InitialPowerTable != nil {
```

certexchange/client.go
```diff
@@ -136,3 +140,53 @@ func (c *Client) Request(ctx context.Context, p peer.ID, req *Request) (_rh *Res
...
+	defer ticker.Stop()
+
+	for {
+		for _, p := range c.Host.Network().Peers() {
+			if ctx.Err() != nil {
+				return nil, ctx.Err()
+			}
+
+			targetProtocol := FetchProtocolName(c.NetworkName)
+			if proto, err := c.Host.Peerstore().FirstSupportedProtocol(p, targetProtocol); err != nil &&
```
Stebalien, line 162:
```suggestion
			if proto, err := c.Host.Peerstore().FirstSupportedProtocol(p, targetProtocol); err != nil ||
```

Stebalien (approved):

PR #490, opened 2024-07-23
Engague the OhShitStore based on the f3Base, not the head

Stebalien:
The certstore stores the _base_ power table, not "future" power tables. So we need to keep EC power tables around until we've passed the instance lookback.

Also, fix an issue in the certstore where we were storing power tables one epoch after we should have.

### DIFF
--- a/certstore/certstore.go
+++ b/certstore/certstore.go
@@ -297,7 +297,7 @@ func (cs *Store) GetPowerTable(ctx context.Context, instance uint64) (gpbft.Powe
 
 	powerTable, err := cs.readPowerTable(ctx, startInstance)
 	if err != nil {
-		return nil, fmt.Errorf("failed fo find expected power table for instance %d: %w", startInstance, err)
+		return nil, fmt.Errorf("failed to find expected power table for instance %d: %w", startInstance, err)
 	}
 	if startInstance == instance {
 		return powerTable, nil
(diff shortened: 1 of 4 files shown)


Stebalien:
(consider merging by rebase)

certstore/certstore.go
```diff
@@ -389,8 +389,9 @@ func (cs *Store) Put(ctx context.Context, cert *certs.FinalityCertificate) error
 		return fmt.Errorf("putting the cert: %w", err)
 	}
 
-	if cert.GPBFTInstance%cs.powerTableFrequency == 0 {
-		if err := cs.putPowerTable(ctx, cert.GPBFTInstance, cs.latestPowerTable); err != nil {
+	// The new power table is the power table to validate the _next_ instance.
+	if (cert.GPBFTInstance+1)%cs.powerTableFrequency == 0 {
+		if err := cs.putPowerTable(ctx, cert.GPBFTInstance+1, newPowerTable); err != nil {
```
Kubuxu, line 394: The fact that this has been here for so long makes me slightly unsure of f3 with changing power table.

Kubuxu (approved):

certstore/certstore.go
Stebalien: So, the power table _diff_ was correct. The issue was that we were storing it at the wrong time. But yeah, the current tests _should_ catch this (I plan on looking into it today).

PR #497, opened 2024-07-23
Explicitly configure pubsub message IDs

Stebalien:
1. Explicitly use a hash of the message data for GPBFT messages to prevent an attacker from spamming the network by replaying valid messages. This is already what lotus does by default.
2. Explicitly use the default deduplication function (sender + nonce) for manifest updates as validation depends on the sender. This undoes the setting from Lotus.


Stebalien:
Hm. Ok, for the standalone manifest server... we likely want to copy the rest of the config from lotus as well.

Kubuxu (approved):

Kubuxu:
Looks good but it broke tests

Kubuxu (approved):

PR #498, opened 2024-07-23
Implement nil manifest unmarshalling

masih:
Move manifest unmarshalling to a package level function that can correctly unmarshal "null" JSON string as `nil` manifest pointer.

Add tests that assert expected behaviour.


Kubuxu (approved):

Stebalien (approved):
So this isn't for the network protocol, but for the local manifest? Honestly, I'd kind of prefer it if we had a way to pause locally without emptying the manifest.... but I guess this is fine.

PR #500, opened 2024-07-23
Reintroduce manifest pause

masih:
Add the manifest pause flag to explicitly stop the F3 participation of a node. This is a clearer signal compared to nil manifest.


Kubuxu (approved):

f3.go
```diff
@@ -160,7 +156,7 @@ func (m *F3) Start(startCtx context.Context) (_err error) {
 	}
 
 	// Try to get an initial manifest immediately if possible so uses can query it immediately.
-	var pendingManifest *manifest.Manifest
+	var pendingManifest manifest.Manifest
 	select {
 	case pendingManifest = <-m.manifestProvider.ManifestUpdates():
```
Stebalien, line 161: This will leave the manifest uninitialized.  We likely do need to keep the manifest nil in this case, or at least keep some kind of "have manifest" boolean.

f3.go
```diff
@@ -226,7 +214,6 @@ func (m *F3) Start(startCtx context.Context) (_err error) {
 				if err := m.reconfigure(m.runningCtx, pendingManifest); err != nil {
 					return fmt.Errorf("failed to reconfigure F3: %w", err)
 				}
-				pendingManifest = nil
```
Stebalien, line 229: We need to record the fact that we've applied the manifest somehow.

f3.go
```diff
@@ -126,11 +126,7 @@ func (m *F3) GetCert(ctx context.Context, instance uint64) (*certs.FinalityCerti
 
 // Returns the time at which the F3 instance specified by the passed manifest should be started, or
 // 0 if the passed manifest is nil.
-func (m *F3) computeBootstrapDelay(manifest *manifest.Manifest) (time.Duration, error) {
-	if manifest == nil {
-		return 0, nil
-	}
-
+func (m *F3) computeBootstrapDelay(manifest manifest.Manifest) (time.Duration, error) {
```
Stebalien, line 129: We need to return 0 if paused, at the very least

f3.go
```diff
@@ -253,21 +240,22 @@ func (m *F3) reconfigure(ctx context.Context, manif *manifest.Manifest) (_err er
...
 		return nil
 	}
+	// Pause if explicitly paused.
+	if manif.Pause {
+		return nil
+	}
 
 	// If we have a new manifest, reconfigure.
-	if m.manifest == nil || m.manifest.NetworkName != manif.NetworkName {
+	if m.manifest.NetworkName != manif.NetworkName {
```
Stebalien, line 256: I'd move this case up (unset the certificate store if the new manifest name disagrees with the old one.

Kubuxu (approved):

Stebalien (approved):

PR #501, opened 2024-07-23
Manifest: Separate sections per subsystem

Stebalien:
Don't embed them all in a single object, it makes it hard to read the encoded JSON.

### DIFF
--- a/manifest/manifest_test.go
+++ b/manifest/manifest_test.go
@@ -27,19 +27,19 @@ var base = manifest.Manifest{
 			PubKey: gpbft.PubKey{1},
 		},
 	},
-	GpbftConfig: manifest.GpbftConfig{
+	Gpbft: manifest.GpbftConfig{
 		Delta:                10,
 		DeltaBackOffExponent: 0.2,
 		MaxLookaheadRounds:   10,
 	},
-	EcConfig: manifest.EcConfig{
-		ECFinality:               900,
+	EC: manifest.EcConfig{
+		Finality:                 900,
 		CommitteeLookback:        5,
-		ECDelayMultiplier:        2.0,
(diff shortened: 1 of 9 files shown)


manifest/manifest_test.go
```diff
@@ -27,19 +27,19 @@ var base = manifest.Manifest{
...
+	Gpbft: manifest.GpbftConfig{
 		Delta:                10,
 		DeltaBackOffExponent: 0.2,
 		MaxLookaheadRounds:   10,
 	},
-	EcConfig: manifest.EcConfig{
-		ECFinality:               900,
+	EC: manifest.EcConfig{
+		Finality:                 900,
 		CommitteeLookback:        5,
```
Stebalien, line 37: Move this into the GPBFT section?

manifest/manifest_test.go
```diff
@@ -27,19 +27,19 @@ var base = manifest.Manifest{
 			PubKey: gpbft.PubKey{1},
 		},
 	},
-	GpbftConfig: manifest.GpbftConfig{
+	Gpbft: manifest.GpbftConfig{
 		Delta:                10,
```
Stebalien, line 31: It would be nice to have a better name.

manifest/manifest_test.go
Kubuxu: Yes

Kubuxu: No

Kubuxu: Hmm, gpbft section gets passed to GPBFT and GPBFT does not care about committee lookback, the lookback is external to it.

Stebalien: I moved it to the root.

Kubuxu (approved):

PR #506, opened 2024-07-23
Implement manifest validation

Stebalien:


masih (approved):

masih (approved):

manifest/manifest.go
```diff
@@ -148,6 +194,47 @@ func (m *Manifest) Equal(o *Manifest) bool {
 
 }
 
+func (m *Manifest) Validate() error {
+	switch {
+	case m == nil:
+		return fmt.Errorf("invalid manifest: manifest is nil!")
+	case m.NetworkName == "":
+		return fmt.Errorf("invalid manifest: network name must not be empty")
+	case m.BootstrapEpoch < m.EC.Finality:
```
Kubuxu, line 203:
```suggestion
	case m.BootstrapEpoch <= m.EC.Finality:
```

Kubuxu (approved):

masih (approved):

manifest/manifest.go
Stebalien: This _should_ work as-is, and I don't think it's important enough to change.

PR #518, opened 2024-07-25
Encode BigInts with a leading sign byte as we do everywhere else

### LINKED ISSUE #517: Correctly handle sign in big-int serialization
cbor-gen ignores the sign when encoding `big.Int` from go stdlib. Everywhere else in Filecoin uses the one from [go state types](https://github.com/filecoin-project/go-state-types/blob/9c9af5cce8781845379e7abfa9c32bd06f9769a1/big/int.go#L15-L17) which encodes it as a leading 0/1 byte.

So, the question is what to do for the normal power table? We don't actually care there, but I assume we want to be consistent?

Stebalien:
Previously, we dropped the sign. Unfortunately, the only ways to do this were to:

1. Replace the StoragePower type (what I did here).
2. Hack marshaling/unmarshaling for anything containing this type (which I really didn't want to do).

I mostly copied the big-int code from go-state-types, but I cleaned it up a bit and removed some unused parts. I also renamed/tweaked some functions because I _really_ hate the go big int interface.

fixes #517

### DIFF
--- a/big/int.go
+++ b/big/int.go
@@ -1,0 +1,270 @@
+package big
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"math/big"
+
+	cbg "github.com/whyrusleeping/cbor-gen"
+)
+
+// BigIntMaxSerializedLen is the max length of a byte slice representing a CBOR serialized big.
+const BigIntMaxSerializedLen = 128
+
+type Int big.Int
+
+func NewInt(i int64) *Int {
+	return (*Int)(big.NewInt(i))
+}
+
+func (i *Int) int() *big.Int {
+	return (*big.Int)(i)
+}
+
+func FromString(s string) (*Int, error) {
+	v, ok := new(big.Int).SetString(s, 10)
+	if !ok {
+		return nil, fmt.Errorf("failed to parse string as a big int")
+	}
+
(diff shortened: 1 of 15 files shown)


big/int.go
```diff
@@ -0,0 +1,270 @@
+package big
```
Stebalien, line 1: From https://github.com/filecoin-project/go-state-types with a few cleanups. Importantly, nil is no longer a half-way alternative to zero...

certs/cbor_gen.go
```diff
@@ -113,36 +100,22 @@ func (t *PowerTableDelta) UnmarshalCBOR(r io.Reader) (err error) {
...
-		buf := make([]byte, extra)
-		if _, err := io.ReadFull(cr, buf); err != nil {
+		b, err := cr.ReadByte()
+		if err != nil {
 			return err
 		}
-		t.PowerDelta = big.NewInt(0).SetBytes(buf)
-	} else {
-		t.PowerDelta = big.NewInt(0)
+		if b != cbg.CborNull[0] {
```
Stebalien, line 109: Ah, damn. We don't want this field to be nullable.

Stebalien:
Closing because this can end up with null ints.

PR #526, opened 2024-07-25
configure code cov threshold

Stebalien:
And only set a CI status on pull requests, not branches (nothing on main).

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,3 +1,10 @@
+coverage:
+  status:
+    project:
+      default:
+        target: auto
+        threshold: 5%
+        only_pulls: true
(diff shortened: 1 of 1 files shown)


codecov.yml
```diff
@@ -1,3 +1,10 @@
+coverage:
+  status:
+    project:
```
masih, line 3: Do we want to set patch threshold too? https://docs.codecov.com/docs/commit-status#patch-status

masih (approved):

codecov.yml
Stebalien: Hm. Probably.

PR #528, opened 2024-07-26
Measure time spent on validating messages tagged by result

masih:
Measure the time spent on validating messages and tag the metric by validation result.

### DIFF
--- a/host.go
+++ b/host.go
@@ -268,33 +268,34 @@ func (h *gpbftRunner) BroadcastMessage(msg *gpbft.GMessage) error {
 
 var _ pubsub.ValidatorEx = (*gpbftRunner)(nil).validatePubsubMessage
 
-func (h *gpbftRunner) validatePubsubMessage(ctx context.Context, pID peer.ID,
-	msg *pubsub.Message) pubsub.ValidationResult {
+func (h *gpbftRunner) validatePubsubMessage(ctx context.Context, _ peer.ID, msg *pubsub.Message) (_result pubsub.ValidationResult) {
+	start := time.Now()
+	defer func() {
+		recordValidationTime(ctx, start, _result)
+	}()
+
 	var gmsg gpbft.GMessage
-	err := gmsg.UnmarshalCBOR(bytes.NewReader(msg.Data))
-	if err != nil {
+	if err := gmsg.UnmarshalCBOR(bytes.NewReader(msg.Data)); err != nil {
 		return pubsub.ValidationReject
 	}
 
-	validatedMessage, err := h.participant.ValidateMessage(&gmsg)
-	if errors.Is(err, gpbft.ErrValidationInvalid) {
+	switch validatedMessage, err := h.participant.ValidateMessage(&gmsg); {
+	case errors.Is(err, gpbft.ErrValidationInvalid):
 		log.Debugf("validation error during validation: %+v", err)
 		return pubsub.ValidationReject
-	}
-	if errors.Is(err, gpbft.ErrValidationTooOld) {
+	case errors.Is(err, gpbft.ErrValidationTooOld):
 		// we got the message too late
 		return pubsub.ValidationIgnore
-	}
-	if errors.Is(err, gpbft.ErrValidationNoCommittee) {
+	case errors.Is(err, gpbft.ErrValidationNoCommittee):
 		log.Debugf("commitee error during validation: %+v", err)
 		return pubsub.ValidationIgnore
-	}
-	if err != nil {
+	case err != nil:
 		log.Infof("unknown error during validation: %+v", err)
 		return pubsub.ValidationIgnore
+	default:
+		msg.ValidatorData = validatedMessage
+		return pubsub.ValidationAccept
 	}
-	msg.ValidatorData = validatedMessage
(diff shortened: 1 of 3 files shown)


host.go
```diff
@@ -268,33 +268,36 @@ func (h *gpbftRunner) BroadcastMessage(msg *gpbft.GMessage) error {
...
+	defer func() {
+		recordValidationTime(start, _result)
+	}()
+
 	var gmsg gpbft.GMessage
-	err := gmsg.UnmarshalCBOR(bytes.NewReader(msg.Data))
-	if err != nil {
-		return pubsub.ValidationReject
+	if err := gmsg.UnmarshalCBOR(bytes.NewReader(msg.Data)); err != nil {
+		_result = pubsub.ValidationReject
```
Stebalien, line 279: Just return this. Named return values are always accessible from deferred functions (and bare returns are a language feature that should never have been implemented).

Stebalien: Or, per the go tour:

>  Naked return statements should be used only in short functions, as with the example shown here. They can harm readability in longer functions. 

Although I still think they should never have been added...

Stebalien: Also note: the reason we generally use `_name` for named return values instead of just `name` is to make it clear that they shouldn't be set. 

The concrete problem is that local variables can shadow named return values.

host.go
```diff
@@ -268,33 +268,36 @@ func (h *gpbftRunner) BroadcastMessage(msg *gpbft.GMessage) error {
 
 var _ pubsub.ValidatorEx = (*gpbftRunner)(nil).validatePubsubMessage
 
-func (h *gpbftRunner) validatePubsubMessage(ctx context.Context, pID peer.ID,
-	msg *pubsub.Message) pubsub.ValidationResult {
+func (h *gpbftRunner) validatePubsubMessage(_ context.Context, _ peer.ID, msg *pubsub.Message) (_result pubsub.ValidationResult) {
+	start := time.Now()
+	defer func() {
+		recordValidationTime(start, _result)
```
Stebalien, line 274: Any reason not to feed the context through?

masih: Because I thought OTEL will not record the metric if the context passed to metric is cancelled.

Based on a quick local test that is not true, and it measures things regardless of the state of context passed to a metric, which is slightly confusing for me.

Thanks Steven

metrics.go
```diff
@@ -1,19 +1,47 @@
...
-	reconfigured:       must(meter.Int64Counter("f3_reconfigured", metric.WithDescription("Number of times we reconfigured due to new manifest being delivered."))),
-	manifests_received: must(meter.Int64Counter("f3_manifests_received", metric.WithDescription("Number of manifests we have received"))),
+	headDiverged:      must(meter.Int64Counter("f3_head_diverged", metric.WithDescription("Number of times we encountered the head has diverged from base scenario."))),
+	reconfigured:      must(meter.Int64Counter("f3_reconfigured", metric.WithDescription("Number of times we reconfigured due to new manifest being delivered."))),
+	manifestsReceived: must(meter.Int64Counter("f3_manifests_received", metric.WithDescription("Number of manifests we have received"))),
+	validationTime: must(meter.Int64Histogram("f3_validation_time_ms",
+		metric.WithDescription("Histogram of time spent validating broadcasted in milliseconds"),
+		metric.WithExplicitBucketBoundaries(5.0, 10.0, 20.0, 30.0, 40.0, 50.0, 100.0, 200.0, 300.0, 400.0, 500.0, 1000.0),
```
Kubuxu, line 25: Does the Histogram give us the `_sum` metric as well?

Kubuxu: I would also add maybe two buckets of smaller sizes (I'm being optimistic).

masih: > Does the Histogram give us the `_sum` metric as well?

It does yes.

> I'm being optimistic

I like it!

Kubuxu (approved):

metrics.go
```diff
@@ -1,19 +1,48 @@
...
-	reconfigured:       must(meter.Int64Counter("f3_reconfigured", metric.WithDescription("Number of times we reconfigured due to new manifest being delivered."))),
-	manifests_received: must(meter.Int64Counter("f3_manifests_received", metric.WithDescription("Number of manifests we have received"))),
+	headDiverged:      must(meter.Int64Counter("f3_head_diverged", metric.WithDescription("Number of times we encountered the head has diverged from base scenario."))),
+	reconfigured:      must(meter.Int64Counter("f3_reconfigured", metric.WithDescription("Number of times we reconfigured due to new manifest being delivered."))),
+	manifestsReceived: must(meter.Int64Counter("f3_manifests_received", metric.WithDescription("Number of manifests we have received"))),
+	validationTime: must(meter.Int64Histogram("f3_validation_time_ms",
+		metric.WithDescription("Histogram of time spent validating broadcasted in milliseconds"),
+		metric.WithExplicitBucketBoundaries(1.0, 2.0, 3.0, 5.0, 10.0, 20.0, 30.0, 40.0, 50.0, 100.0, 200.0, 300.0, 400.0, 500.0, 1000.0),
+		metric.WithUnit("ms"),
```
masih, line 26: @Stebalien `millisecond` unit example.

PR #529, opened 2024-07-29
Add basic certificate exchange metrics

Stebalien:
This adds basic metrics so we can know how many requests we're making, how many peers we're talking to, how often we're polling, etc.

I still want to add some metrics around how often we're polling when we don't _need_ to, but I also want to rework some of that logic at the same time to reduce the number of unnecessary polls. So I'm punting that out of this round of work.

### DIFF
--- a/certexchange/client.go
+++ b/certexchange/client.go
@@ -16,8 +16,6 @@ import (
 	cid "github.com/ipfs/go-cid"
 	"github.com/libp2p/go-libp2p/core/host"
 	"github.com/libp2p/go-libp2p/core/peer"
-	"go.opentelemetry.io/otel"
-	"go.opentelemetry.io/otel/metric"
 )
 
 // We've estimated the max power table size to be less than 1MiB:
@@ -25,26 +23,6 @@ import (
 // 1. For 10k participants.
 // 2. <100 bytes per entry (key + id + power)
 const maxPowerTableSize = 1024 * 1024
-
-var clientMeter = otel.Meter("f3/certexchange/client")
-var clientMetrics = struct {
-	requests        metric.Int64Counter
-	dialFailures    metric.Int64Counter
-	requestFailures metric.Int64Counter
-}{
-	requests: must(clientMeter.Int64Counter(
-		"f3_certexchange_client_requests",
-		metric.WithDescription("The total number of requests made."),
-	)),
-	dialFailures: must(clientMeter.Int64Counter(
-		"f3_certexchange_client_failed_dials",
-		metric.WithDescription("The number of failed certexchange dials."),
-	)),
-	requestFailures: must(clientMeter.Int64Counter(
-		"f3_certexchange_client_failed_requests",
-		metric.WithDescription("The number of failed certexchange requests (total)."),
-	)),
-}
 
 // Client is a libp2p certificate exchange client for requesting finality certificates from specific
 // peers.
--- a/certexchange/client.go
+++ b/certexchange/client.go
@@ -16,6 +16,7 @@ import (
 	cid "github.com/ipfs/go-cid"
 	"github.com/libp2p/go-libp2p/core/host"
 	"github.com/libp2p/go-libp2p/core/peer"
+	"go.opentelemetry.io/otel/metric"
 )
 
 // We've estimated the max power table size to be less than 1MiB:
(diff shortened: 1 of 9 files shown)


Stebalien:
@masih I've never used open telemetry before so tell me if this looks correct. If not, don't waste a bunch of time pointing out every issue, just give me a general idea and then point me to some docs.

certexchange/client.go
```diff
@@ -24,6 +26,26 @@ import (
 // 2. <100 bytes per entry (key + id + power)
 const maxPowerTableSize = 1024 * 1024
 
+var clientMeter = otel.Meter("f3/certexchange/client")
```
Stebalien, line 29: I have two of these in this package. Is that ok?

certexchange/polling/peerTracker_test.go
```diff
@@ -120,20 +121,20 @@ func TestPeerTracker(t *testing.T) {
 	for _, n := range []int{0, 1, defaultRequests / 2, defaultRequests/2 - 1} {
 		discoverPeers(n)
 		pt.lastHitRound = pt.currentRound
-		suggested := pt.suggestPeers()
+		suggested := pt.suggestPeers(context.Background())
```
masih, line 124: Reuse context?

Kubuxu (approved):

certexchange/polling/subscriber.go
```diff
@@ -16,6 +18,58 @@ import (
...
+	)),
+	predictedPollingIntervalMS: must(meter.Int64Gauge(
+		"f3_certexchange_predicted_polling_interval_ms",
+		metric.WithDescription("The predicted certificate exchange polling interval (milliseconds)."),
+	)),
+	pollRTTMS: must(meter.Int64Histogram(
+		"f3_certexchange_poll_rtt_ms",
+		metric.WithDescription("The certificate exchange per-peer polling round-trip time (milliseconds)."),
+	)),
+	pollTotalMS: must(meter.Int64Histogram(
```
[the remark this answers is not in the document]
Stebalien: (I've spent too much time staring at metrics wondering what their units were)

certexchange/polling/subscriber.go
```diff
@@ -16,6 +18,58 @@ import (
 
 const maxRequestLength = 256
 
+var meter = otel.Meter("f3/certexchange")
```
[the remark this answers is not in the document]
Stebalien: Ah, as long as that's not going to be an issue, I agree.

certexchange/client.go
```diff
@@ -24,6 +26,26 @@ import (
...
+var clientMetrics = struct {
+	requests        metric.Int64Counter
+	dialFailures    metric.Int64Counter
+	requestFailures metric.Int64Counter
+}{
+	requests: must(clientMeter.Int64Counter(
```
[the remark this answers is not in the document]
Stebalien: Ah, ok. I'll use tagging. I'm currently tracking latency in the poller... but I should probably move that to the client.

certexchange/polling/subscriber.go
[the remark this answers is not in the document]
Stebalien: AH, we should just specify units. I see that there's an actual option for this.

Kubuxu:
In theory, the recommendation is to use base units but 🤷‍♂️.

Stebalien:
@masih @Kubuxu I refactored the metrics to use attributes, units, etc.

Stebalien:
Er, apparently I didn't finish this.

Stebalien:
Ok, actually done.

certexchange/client.go
```diff
@@ -56,11 +59,23 @@ func (c *Client) Request(ctx context.Context, p peer.ID, req *Request) (_rh *Res
 		}
 	}()
 
+	requestStart := time.Now()
+	var dialSuceeded bool
+	defer func() {
+		d := float64(time.Since(requestStart)) / float64(time.Second)
+		metrics.requestLatency.Record(ctx, d, metric.WithAttributes(
+			mhelper.Status(ctx, _err),
+			mhelper.AttrDialSucceeded.Bool(dialSuceeded),
```
Stebalien, line 68: Separate attribute because I want to track whether or not we failed in the dial and _why_ we failed.

certexchange/metrics.go
```diff
@@ -0,0 +1,38 @@
...
+var attrWithPowerTable = attribute.Key("with-power-table")
+
+var metrics = struct {
+	requestLatency     metric.Float64Histogram
+	totalResponseTime  metric.Float64Histogram
+	serveTime          metric.Float64Histogram
+	certificatesServed metric.Int64Histogram
+}{
+	requestLatency: must(meter.Float64Histogram(
+		"f3_certexchange_request_latency_s",
```
masih, line 19: Considering exporter adds unit to name we probably want to avoid adding _s?

certexchange/polling/metrics.go
```diff
@@ -0,0 +1,62 @@
...
+	activePeers              metric.Int64Gauge
+	backoffPeers             metric.Int64Gauge
+	predictedPollingInterval metric.Float64Gauge
+	pollDuration             metric.Float64Histogram
+	peersPolled              metric.Int64Histogram
+	peersRequiredPerPoll     metric.Int64Histogram
+	pollEfficiency           metric.Float64Histogram
+}{
+	activePeers: must(meter.Int64Gauge(
+		"f3_certexchange_active_peers",
```
masih, line 20: Include "polling" or "poll" in prefix to avoid possible naming conflict with metrics at parent package?

masih (approved):

PR #530, opened 2024-07-30
Offset the predicted CX interval if we align with the GPBFT interval

Stebalien:
If we end up polling for certificates at approximately same time we receive them from GPBFT, add a one-time delay to our timer. This won't affect our predicted interval, but will prevent us from racing with GPBFT for every certificate.

This will solve the issue @Kubuxu was seeing, but it'll still be a bit trigger happy if we're participating in GPBFT (e.g., due to instance duration variance). So, I'm posting this PR (it's still useful regardless) but I want to try something different: let the user configure how many instances they want to lag by. For clients participating in GPBFT, the answer is something less than the power lookback, where 2-5 is probably a good number. For clients not participating in GPBFT, the answer is 0.

### DIFF
--- a/certexchange/polling/poller.go
+++ b/certexchange/polling/poller.go
@@ -45,10 +45,15 @@ func NewPoller(ctx context.Context, client *certexchange.Client, store *certstor
 }
 
 type PollResult struct {
-	Status   PollStatus
-	Progress uint64
-	Latency  time.Duration
-	Error    error
+	Status  PollStatus
+	Latency time.Duration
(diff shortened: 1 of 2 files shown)


Kubuxu (approved):
LGTM, although it is hard for me to say that it will work precisely as we want it to.

Stebalien:
I'm still on the fence on whether or not I want to merge this. I'm going to start with the alternative approach described in the PR description and see where that gets.

PR #531, opened 2024-07-31
Measure time spent on validation as seconds in float64

masih:
For better accuracy and compliance with otel best practices measure the time spent on validating messages as float64 seconds.

While at it, use inline var struct instantiation instead of init to follow a consistent convention across the repo.

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -41,7 +41,7 @@ func recordValidationTime(ctx context.Context, start time.Time, result pubsub.Va
 	}
 	metrics.validationTime.Record(
 		ctx,
-		float64(time.Since(start))/float64(time.Second),
+		time.Since(start).Seconds(),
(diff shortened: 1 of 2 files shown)


metrics.go
```diff
@@ -41,7 +41,7 @@ func recordValidationTime(ctx context.Context, start time.Time, result pubsub.Va
 	}
 	metrics.validationTime.Record(
 		ctx,
-		time.Since(start).Milliseconds(),
+		float64(time.Since(start))/float64(time.Second),
```
Kubuxu, line 44:
```suggestion
		time.Since(start).Seconds(),
```

Kubuxu (approved):

metrics.go
masih: Why?

masih: Never mind... thanks Kuba

Issue #131, opened 2024-03-25
Cache results of validated justifications to short-circuit future message verification


anorth:
Each unique vote value in the protocol only needs to be justified once. After receiving a message with justified value V from one participant, a node need not check the justification received with other messages carrying the same V. This is a huge optimisation, probably necessary for feasible protocol execution at scale.

Kubuxu:
We might have to do this sooner than later.

Stebalien:
From F3 standup discussions: this isn't quite so simple.

1. We can't re-broadcast messages with invalid justifications, even if we've seen a valid justification.
2. We can't easily replace the message's justification given that pubsub messages are signed (and identified by the original sender/nonce).
3. We _can_ disable pubsub signatures and identify messages by their GPBFT payload, but, unfortunately, we use message IDs to deduplicate messages _before_ validating them. That means, if we receive a message with payload X and mark it as invalid (e.g., has an invalid justification) then receive the same payload X with a correct justification, pubsub will drop it before validation because we're identifying inbound messages by payload value.

The solution to all this is to give pubsub two different ways of identifying messages:

1. For pre-validation deduplication, identify messages by hash, always.
2. Only use the user-configurable message ID _after_ validation (to determine if we should rebroadcast the message, to identify messages in gossip, etc.).

However, we'd _additionally_ need the ability to _mutate_ pubsub messages: replace the possibly invalid justification (the one we didn't bother checking) with the known valid justification (from our cache).

masih:
Closing now that both #557 and #560  are merged. Please feel free to reopen if there is more here to do.

PR #534, opened 2024-08-02
Measure duplicate validated messages within a bounded size

masih:
As part of understanding duplicated effort and eventually designing a caching mechanism to reduce it, measure the number validations that are performed for messages, and separately for justifications.

The most expensive part of validation is signature validation, two of which exists at `GMessage` and `Justification`. Measuring the duplicate validations at the top level should help us understand the gain in caching already validated signatures.

The implementation introduces a new data structure to determine whether a given message/justification is seen before or not, identified by signatures. The implementation uses a total bounded size of 50K entries divided equally across counting duplicate messages and duplicate justifications. This number should be large enough to provide a reasonable signal about the amount of duplicate validations at runtime.

Relates to https://github.com/filecoin-project/go-f3/issues/131

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -2,8 +2,10 @@ package f3
 
 import (
 	"context"
+	"sync"
 	"time"
 
+	"github.com/filecoin-project/go-f3/gpbft"
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
@@ -11,13 +13,38 @@ import (
 )
 
 var meter = otel.Meter("f3")
+var sampleSets = struct {
+	messages       *sampleSet
+	justifications *sampleSet
+}{
+	// The the max size of 25K is based on network size of ~3K and 5 phases of GPBFT,
+	// where the vast majority of instances should complete in one round. That makes
+	// up ~15K messages per instance. We may observe messages from previous or future
+	// rounds, hence an additional capacity of 10K, making a total of 25K messages.
+	//
+	// Although, the number of observable messages with justification are expected to
+	// be less than the total number of observed messages we use the same 25K bounded
+	// size for justifications. The memory footprint of the additionally stored
+	// samples is negligible in return for a larger sample size and ultimately a more
+	// accurate measurement considering justification signature validation is more
+	// expensive than message signature validation.
+	//
+	// Since sampleSet caps the max size of cache keys to 96 bytes (the length
+	// of BLS signatures) the total memory footprint for 50K samples should be be
+	// below 10MB (5MB * 2 due to bookkeeping overhead).
+	messages:       newSampleSet(25_000),
+	justifications: newSampleSet(25_000),
+}
+
 var metrics = struct {
-	headDiverged       metric.Int64Counter
-	reconfigured       metric.Int64Counter
-	manifestsReceived  metric.Int64Counter
-	validationTime     metric.Float64Histogram
-	proposalFetchTime  metric.Float64Histogram
-	committeeFetchTime metric.Float64Histogram
+	headDiverged                      metric.Int64Counter
+	reconfigured                      metric.Int64Counter
+	manifestsReceived                 metric.Int64Counter
+	validationTime                    metric.Float64Histogram
+	proposalFetchTime                 metric.Float64Histogram
+	committeeFetchTime                metric.Float64Histogram
+	validationDuplicateMessages       metric.Int64Counter
+	validationDuplicateJustifications metric.Int64Counter
 }{
 	headDiverged:      must(meter.Int64Counter("f3_head_diverged", metric.WithDescription("Number of times we encountered the head has diverged from base scenario."))),
 	reconfigured:      must(meter.Int64Counter("f3_reconfigured", metric.WithDescription("Number of times we reconfigured due to new manifest being delivered."))),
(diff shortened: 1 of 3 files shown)


Kubuxu (approved):

metrics.go
```diff
@@ -1,23 +1,51 @@
...
+	// size for justifications. The memory footprint of the additionally stored
+	// samples is negligible in return for a larger sample size and ultimately a more
+	// accurate measurement considering justification signature validation is more
+	// expensive than message signature validation.
+	//
+	// Since sampleSet caps the max size of cache keys to 96 bytes (the length
+	// of BLS signatures) the total memory footprint for 50K samples should be be
+	// below 10MB (5MB * 2 due to bookkeeping overhead).
+	messages:       newSampleSet(25_000),
+	justifications: newSampleSet(25_000),
```
Stebalien, line 37: 1. I'd consider doubling this, just for better metrics.
2. It shouldn't be that hard to benchmark this. Fill up a set then call `runtime.ReadMemStats`.

metrics.go
```diff
@@ -76,3 +121,63 @@ func must[V any](v V, err error) V {
...
+
+	// maxSize defines the maximum number of samples to store.
+	maxSize int
+	// mu protects access to samples.
+	mu sync.Mutex
+	// samples stores the samples using a map for quick lookup.
+	samples map[string]struct{}
+	// order maintains the order of samples to evict by oldest entry irrespective of
+	// recency.
+	order *list.List
```
Stebalien, line 143: This works but, for this kind of thing, I generally prefer a live set and a decaying set. I.e.:

1. Fill up a live set until we reach a threshold.
2. At that point, rotate it out into the "decaying slot", never update it, discarding the previous decaying set.
3. Create a new live set.

It simplifies GC and removes the need for an ordered list, at the cost of potentially 2x memory.

Stebalien (approved):

metrics.go
masih: I like this; this approach should keep the memory footprint at 2X ish while potentially doubling the sample size 🍻

masih: @Stebalien I have implemented this suggestion. would you mind giving this PR another look before I merge please?

metrics.go
masih: We could bump this later if numbers come back high, unless you think 25K is too small to begin with?

metrics.go
masih: Actually hold off. I have missed something.

metrics.go
Stebalien: Nah, should be fine.

metrics.go
```diff
@@ -76,3 +120,87 @@ func must[V any](v V, err error) V {
...
+	// flop stores another set of samples until it reaches maxSizePerSet.
+	flop map[string]struct{}
+}
+
+// newSampleSet creates a new sampleSet with a specified max size.
+func newSampleSet(maxSize int) *sampleSet {
+	return &sampleSet{
+		maxSizePerSet: max(1, maxSize/2),
+		flip:          make(map[string]struct{}),
+		flop:          make(map[string]struct{}),
```
Stebalien, line 159: nit: might as well allocate the max size up-front.

metrics.go
```diff
@@ -76,3 +120,87 @@ func must[V any](v V, err error) V {
...
+		active = ss.flip
+	}
+	active[key] = struct{}{}
+
+	// Check if active map exceeds maxSize and switch if needed.
+	if len(active) >= ss.maxSizePerSet {
+		if isFlopActive {
+			ss.flip = make(map[string]struct{})
+		} else {
+			ss.flop = make(map[string]struct{})
```
Stebalien, line 201: Consider making use of go's new "clear" function to reduce GC pressure a bit once we reach a steady-state.

metrics.go
```diff
@@ -76,3 +120,87 @@ func must[V any](v V, err error) V {
...
+		}
+	}
```
Stebalien, line 203: You can make this a bit shorter by literally flipping:

```suggestion
	ss.flip[key] = struct{}{}

	// Check if flip exceeds maxSize and switch it with flop if needed.
	if len(ss.flip) >= ss.maxSizePerSet {
		clear(ss.flop)
		ss.flop, ss.flip = ss.flip, ss.flop
	}
```

metrics.go
```diff
@@ -76,3 +120,87 @@ func must[V any](v V, err error) V {
...
+// contains checks if the given sample v is contained within flip set.
+func (ss *sampleSet) contains(v []byte) bool {
+	// The number 96 comes from the length of BLS signatures, the kind of value we
+	// expect as the argument. Defensively re-slice it if it is larger at the price
+	// of losing accuracy.
+	//
+	// Alternatively we could hash the values but considering the memory footprint of
+	// these measurements (sub 10MB for a total of 50K samples) we choose larger
+	// memory consumption over CPU footprint.
+	key := string(v[:min(len(v), 96)])
```
Stebalien, line 172: I'd consider using `[96]byte` to reduce allocations. But that's getting _real_ nit-picky.

Stebalien (approved):
Down to the smallest of nits.

metrics.go
masih: Thanks to flippity flopping i think we effectively get uniqueness check over 2X the max size with the same (if not slightly less) memory footprint 🎉

metrics.go
masih: > I have missed something.

For reference, this was storing half the max size in each subset map, which is not right since it cannot guarantee uniqueness check across _at least_ max size number of samples. I changed it to use the max size directly as the size of each subset.

metrics.go
masih: I left this one out. I think we can do a bit more maybe in follow up PRs. 

P.S. loving the attention to detail 🙇

Kubuxu (approved):

PR #538, opened 2024-08-07
Naive observer node

Kubuxu:


cmd/f3/observer.go
```diff
@@ -0,0 +1,320 @@
...
+	jsonWriter := json.NewEncoder(msgLogFile)
+
+	err = pubSub.RegisterTopicValidator(manif.PubSubTopic(), func(_ context.Context, _ peer.ID, msg *pubsub.Message) pubsub.ValidationResult {
+		var gmsg gpbft.GMessage
+		if err := gmsg.UnmarshalCBOR(bytes.NewReader(msg.Data)); err != nil {
+			return pubsub.ValidationReject
+		}
+		msg.ValidatorData = gmsg
+
+		return pubsub.ValidationAccept
```
Stebalien, line 237: Could get us blacklisted, but probably fine for now.

cmd/f3/observer.go
```diff
@@ -0,0 +1,320 @@
...
+		return fmt.Errorf("joining topic: %w", err)
+	}
+
+	sub, err := topic.Subscribe()
+	if err != nil {
+		return fmt.Errorf("subscribing to topic: %w", err)
+	}
+
+	sigCh := make(chan os.Signal, 1)
+	signal.Notify(sigCh, syscall.SIGUSR1)
```
Stebalien, line 254: might as well handle interrupts as well? Or will the context do that?

Stebalien (approved):

cmd/f3/observer.go
Kubuxu: Context will do that

PR #543, opened 2024-08-08
Cache decompressed points

### LINKED ISSUE #542: Cache BLS point decompression
We spend 35% of our time decompressing public keys, but that operation is trivially cachable.

Stebalien:
This can save quite a bit of CPU time.

fixes #542

### DIFF
--- a/blssig/aggregation.go
+++ b/blssig/aggregation.go
@@ -74,7 +74,7 @@ func (v *Verifier) pubkeyToPoint(p gpbft.PubKey) (kyber.Point, error) {
 	// Initialize the cache, or re-initialize it if we've grown too big. We don't expect the
 	// latter to happen in practice (would need over 10k participants), but better be safe than
 	// sorry. We could, alternatively, use an LRU but... that's not worth the overhead for
-	// somethign that shouldn't happen.
+	// something that shouldn't happen.
 	if v.pointCache == nil || len(v.pointCache) >= maxPointCacheSize {
 		v.pointCache = make(map[string]kyber.Point)
 	}
--- a/blssig/aggregation.go
+++ b/blssig/aggregation.go
@@ -49,42 +49,6 @@ func (v *Verifier) VerifyAggregate(msg []byte, signature []byte, pubkeys []gpbft
 	return v.scheme.Verify(aggPubKey, msg, signature)
 }
 
-func (v *Verifier) pubkeyToPoint(p gpbft.PubKey) (kyber.Point, error) {
-	if len(p) > 96 {
-		return nil, fmt.Errorf("public key is too large: %d > 96", len(p))
-	}
-
-	v.mu.RLock()
-	point, ok := v.pointCache[string(p)]
-	v.mu.RUnlock()
-	if ok {
-		return point.Clone(), nil
-	}
-
-	point = v.keyGroup.Point()
-	err := point.UnmarshalBinary(p)
-	if err != nil {
-		return nil, fmt.Errorf("unarshalling pubkey: %w", err)
-	}
-	if point.Equal(v.keyGroup.Point().Null()) {
-		return nil, fmt.Errorf("public key is a null point")
-	}
-	v.mu.Lock()
-
-	// Initialize the cache, or re-initialize it if we've grown too big. We don't expect the
-	// latter to happen in practice (would need over 10k participants), but better be safe than
-	// sorry. We could, alternatively, use an LRU but... that's not worth the overhead for
-	// something that shouldn't happen.
-	if v.pointCache == nil || len(v.pointCache) >= maxPointCacheSize {
-		v.pointCache = make(map[string]kyber.Point)
-	}
-
-	v.pointCache[string(p)] = point
-	v.mu.Unlock()
-
-	return point.Clone(), nil
-}
-
 func (v *Verifier) pubkeysToMask(pubkeys []gpbft.PubKey) (*sign.Mask, error) {
 	kPubkeys := make([]kyber.Point, 0, len(pubkeys))
 	for i, p := range pubkeys {
--- a/blssig/aggregation.go
+++ b/blssig/aggregation.go
@@ -9,7 +9,10 @@ import (
 	"github.com/drand/kyber/sign"
 )
 
-func (v Verifier) Aggregate(pubkeys []gpbft.PubKey, signatures [][]byte) ([]byte, error) {
+// Max size of the point cache.
+const maxPointCacheSize = 10_000
+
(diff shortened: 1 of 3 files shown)


blssig/aggregation.go
```diff
@@ -46,19 +49,50 @@ func (v Verifier) VerifyAggregate(msg []byte, signature []byte, pubkeys []gpbft.
 	return v.scheme.Verify(aggPubKey, msg, signature)
 }
 
-func (v Verifier) pubkeysToMask(pubkeys []gpbft.PubKey) (*sign.Mask, error) {
+func (v *Verifier) pubkeyToPoint(p gpbft.PubKey) (kyber.Point, error) {
+	if len(p) > 96 {
+		return nil, fmt.Errorf("public key is too large: %d > 96", len(p))
+	}
+
+	v.mu.RLock()
```
Stebalien, line 57: rwmutex because, 99% of the time, I expect this to be in the cache.

blssig/aggregation.go
```diff
@@ -9,7 +9,10 @@ import (
 	"github.com/drand/kyber/sign"
 )
 
-func (v Verifier) Aggregate(pubkeys []gpbft.PubKey, signatures [][]byte) ([]byte, error) {
+// Max size of the point cache.
+const maxPointCacheSize = 10_000
```
Stebalien, line 13: bigger?

blssig/aggregation.go
```diff
@@ -46,19 +49,50 @@ func (v Verifier) VerifyAggregate(msg []byte, signature []byte, pubkeys []gpbft.
...
+	}
+	if point.Equal(v.keyGroup.Point().Null()) {
+		return nil, fmt.Errorf("public key is a null point")
+	}
+	v.mu.Lock()
+
+	// Initialize the cache, or re-initialize it if we've grown too big. We don't expect the
+	// latter to happen in practice (would need over 10k participants), but better be safe than
+	// sorry. We could, alternatively, use an LRU but... that's not worth the overhead for
+	// somethign that shouldn't happen.
```
masih, line 77:
```suggestion
	// something that shouldn't happen.
```

blssig/aggregation.go
masih: Seems big enough for now?

blssig/aggregation.go
```diff
@@ -46,19 +49,50 @@ func (v Verifier) VerifyAggregate(msg []byte, signature []byte, pubkeys []gpbft.
...
+	if v.pointCache == nil || len(v.pointCache) >= maxPointCacheSize {
+		v.pointCache = make(map[string]kyber.Point)
```
masih, line 79: How much churn do we get in participants? 

It doesn't matter much since the cache size is capped but I am curious if it is worth thinking about some eviction strategy to avoid sudden change in performance characteristics should we reach the max size.

blssig/aggregation.go
```diff
@@ -46,19 +49,50 @@ func (v Verifier) VerifyAggregate(msg []byte, signature []byte, pubkeys []gpbft.
...
+	}
+
+	v.pointCache[string(p)] = point
+	v.mu.Unlock()
+
+	return point.Clone(), nil
```
masih, line 85: Dumb question: `point.Clone` is safe for concurrent use; correct?

blssig/cache_test.go
```diff
@@ -0,0 +1,51 @@
...
+	bls12381 "github.com/drand/kyber-bls12381"
+	"github.com/drand/kyber/sign/bdn"
+	"github.com/stretchr/testify/require"
+
+	"github.com/filecoin-project/go-f3/gpbft"
+)
+
+const maxCacheMemory uint64 = 10 * 1023 * 1024 // 10MiB
+
+func TestCacheMemory(t *testing.T) {
```
masih, line 16: While at it assert max pub key size too?

blssig/cache_test.go
```diff
@@ -0,0 +1,51 @@
...
+	for i := range keys {
+		_, pub := scheme.NewKeyPair(rand)
+		pubKeyB, err := pub.MarshalBinary()
+		require.NoError(t, err)
+		keys[i] = pubKeyB
+	}
+
+	runtime.GC()
+	var beforeMemStats, afterMemStats runtime.MemStats
+	runtime.ReadMemStats(&beforeMemStats)
```
masih, line 31: Nice 🥇

blssig/cache_test.go
```diff
@@ -0,0 +1,51 @@
...
+	v := VerifierWithKeyOnG1()
+	for _, k := range keys[1:] {
+		_, err := v.pubkeyToPoint(k)
```
masih, line 34: Being picky: test `pubkeysToMask` too?

masih (approved):
Metrics for this would be awesome, either here or in a separate PR 🍻

blssig/cache_test.go
```diff
@@ -0,0 +1,51 @@
...
+	"testing"
+
+	bls12381 "github.com/drand/kyber-bls12381"
+	"github.com/drand/kyber/sign/bdn"
+	"github.com/stretchr/testify/require"
+
+	"github.com/filecoin-project/go-f3/gpbft"
+)
+
+const maxCacheMemory uint64 = 10 * 1023 * 1024 // 10MiB
```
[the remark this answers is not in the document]
Stebalien: Oh, lol. Yeah, I wrote it as `10<<20` at first but went with `1024 * 1024` (well, 1023 because typos) because I find that marginally clearer.

blssig/cache_test.go
Stebalien: That just calls `pubkeyToPoint` (all I'm testing here is that the cache doesn't grow too large). Everything else should already be tested by our signing test suite.

blssig/cache_test.go
Stebalien: Huh. The answer is actually 48. I think it's 96 for _signatures_?

Stebalien: Fixed and https://github.com/filecoin-project/go-f3/issues/547

blssig/aggregation.go
Stebalien: It's behind an interface but... if it's not, something is very wrong with kyber. Honestly, I'm not even sure if the clone is necessary, I'm just trying to be extra safe.

blssig/aggregation.go
Stebalien: Very little, that's why I stuck with a simple cache instead of trying to come up with a nice eviction strategy. The main source of churn will be SPs changing keys.

NOTE: if we hit >10k participants, _then_ this cache will become useless.

Stebalien:
Metrics: https://github.com/filecoin-project/go-f3/issues/548

Kubuxu (approved):

PR #551, opened 2024-08-08
Add signature verification/aggregation metrics and panic handlers

Stebalien:
I threw in the panic handlers just to be complete, I've seen too many panics in cryptography code.

### DIFF
--- a/blssig/metrics.go
+++ b/blssig/metrics.go
@@ -34,6 +34,6 @@ var metrics = struct {
 	)),
 	aggregate: measurements.Must(meter.Int64Histogram(
 		"f3_blssig_aggregate",
-		metric.WithDescription("Signatures aggregated."),
+		metric.WithDescription("Number of signatures aggregated."),
 	)),
 }
--- a/blssig/metrics.go
+++ b/blssig/metrics.go
@@ -1,0 +1,39 @@
+package blssig
+
+import (
+	logging "github.com/ipfs/go-log/v2"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
+
+	"github.com/filecoin-project/go-f3/internal/measurements"
+)
+
+var meter = otel.Meter("f3/blssig")
+var attrCached = attribute.Key("cached")
+
+var log = logging.Logger("f3/blssig")
+
+var metrics = struct {
+	decompressPoint metric.Int64Counter
+	verify          metric.Int64Counter
+	verifyAggregate metric.Int64Histogram
+	aggregate       metric.Int64Histogram
+}{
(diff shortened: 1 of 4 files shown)


blssig/metrics.go
```diff
@@ -0,0 +1,39 @@
...
+	)),
+	verifyAggregate: measurements.Must(meter.Int64Histogram(
+		"f3_blssig_verify_aggregate",
+		metric.WithDescription("Aggregate signatures verified."),
+	)),
+	aggregate: measurements.Must(meter.Int64Histogram(
+		"f3_blssig_aggregate",
+		metric.WithDescription("Signatures aggregated."),
+	)),
+}
```
Stebalien, line 39: do we need custom buckets here?

blssig/metrics.go
```diff
@@ -0,0 +1,39 @@
...
+		"f3_blssig_verify",
+		metric.WithDescription("Number of signatures verified."),
+	)),
+	verifyAggregate: measurements.Must(meter.Int64Histogram(
+		"f3_blssig_verify_aggregate",
+		metric.WithDescription("Aggregate signatures verified."),
+	)),
+	aggregate: measurements.Must(meter.Int64Histogram(
+		"f3_blssig_aggregate",
+		metric.WithDescription("Signatures aggregated."),
```
masih, line 37: Extra words for numpties like me?

```suggestion
		metric.WithDescription("Number of signatures aggregated."),
```

blssig/metrics.go
```diff
@@ -0,0 +1,39 @@
...
+		"f3_blssig_decompress_point",
+		metric.WithDescription("Number of times we decompress points."),
+	)),
+	verify: measurements.Must(meter.Int64Counter(
+		"f3_blssig_verify",
+		metric.WithDescription("Number of signatures verified."),
+	)),
+	verifyAggregate: measurements.Must(meter.Int64Histogram(
+		"f3_blssig_verify_aggregate",
+		metric.WithDescription("Aggregate signatures verified."),
```
masih, line 33:
```suggestion
		metric.WithDescription("Number of aggregate signatures verified."),
```

blssig/metrics.go
masih: Probably yes, adjusted based on expected network size and quorum? We can refine later if defaults don't make sense.

masih (approved):

blssig/metrics.go
Stebalien: I'll leave the defaults for now.

Kubuxu (approved):

Kubuxu:
I would prefer if the cache had explicit # cache queries and # cache hits to allow calculating the cache hit ratio.

Kubuxu:
Ahh, I see, you are doing it though a label, I guess it makes sense.

PR #552, opened 2024-08-08
Measure the number of validated messages with a finer grain tagging

masih:
Measure the total number of duplicate messages, along with duplicates, justified, and duplicate justifications.

This should provide a finer grain data for a more educated cache design.

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -37,14 +37,13 @@ var samples = struct {
 }
 
 var metrics = struct {
-	headDiverged                      metric.Int64Counter
-	reconfigured                      metric.Int64Counter
-	manifestsReceived                 metric.Int64Counter
-	validationTime                    metric.Float64Histogram
-	proposalFetchTime                 metric.Float64Histogram
-	committeeFetchTime                metric.Float64Histogram
-	validationDuplicateMessages       metric.Int64Counter
-	validationDuplicateJustifications metric.Int64Counter
+	headDiverged       metric.Int64Counter
+	reconfigured       metric.Int64Counter
+	manifestsReceived  metric.Int64Counter
+	validationTime     metric.Float64Histogram
+	proposalFetchTime  metric.Float64Histogram
+	committeeFetchTime metric.Float64Histogram
+	validatedMessages  metric.Int64Counter
 }{
 	headDiverged:      measurements.Must(meter.Int64Counter("f3_head_diverged", metric.WithDescription("Number of times we encountered the head has diverged from base scenario."))),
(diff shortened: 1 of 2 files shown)


metrics.go
```diff
@@ -64,23 +63,34 @@ var metrics = struct {
...
-func recordValidationDuplicates(ctx context.Context, msg gpbft.ValidatedMessage) {
+func recordValidatedMessage(ctx context.Context, msg gpbft.ValidatedMessage) {
 	// The given msg and its validated value should never be nil; but defensively
 	// check anyway.
 	if msg == nil || msg.Message() == nil {
 		return
 	}
+
+	// Count once for total number of validated messages.
+	metrics.validatedMessages.Add(ctx, 1)
```
Stebalien, line 78: remove this?

masih: Would we get total validated messages without it?

Stebalien: Yes? We update this at the end and can just sum disregarding labels.

masih: Sum alone would return sum of duplicates i would imagine. I'll test this tomorrow to be sure this line is needed. Thanks

metrics.go
```diff
@@ -64,23 +63,34 @@ var metrics = struct {
...
+
 	vmsg := msg.Message()
+	attrs := make([]attribute.KeyValue, 0, 3)
 	if samples.messages.Contains(vmsg.Signature) {
-		metrics.validationDuplicateMessages.Add(ctx, 1)
+		attrs = append(attrs, attribute.Bool("duplicate_message", true))
```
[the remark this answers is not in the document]
masih: What specifically would you like to query?

I think what happens here is that we get a metric per permutation of attributes, and only for those that have value record.

metrics.go
Stebalien: We can also sum on `duplicate_message`, etc. That would give us both the count of unique messages and the count of duplicates (separately). I'm flagging this because we're currently counting all messages twice, which is definitely wrong.

metrics.go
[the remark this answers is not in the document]
Stebalien: We can `sum (duplicate_message) (f3_validated_messages)`, that'll give us two timeseries: the duplicates and non-duplicates.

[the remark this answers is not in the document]
masih: I don't think it does @Stebalien unless I have missed something. I will explain in comment above. 

@Kubuxu please unresolve this conversation if it has a different concern from Steven's comment above.

metrics.go
masih: Thank you for your patience; I see now.

Kubuxu (approved):

PR #553, opened 2024-08-09
Align catch-up GPBFT start times with the expected instance interval

### LINKED ISSUE #550: Quantize instance start times when catching up
When catching up (behind by some significant amount of time), we have no great way of aligning start times. This can cause nodes to get "strung out" when catching up, which triggers exponential round backoff which makes catching up take a long time.

A simple solution is to try to align instance start times on some boundary. It won't be perfect but, if we set this boundary to be larger than the expected F3 instance time, we should be able to sync up and move forward quickly.

Stebalien:
When catching up, we expect many single-round instances consisting of 4 phases. So we try to align our start times with that to avoid staggering.

This should probably have a simulation test...

fixes #550

### DIFF
--- a/host.go
+++ b/host.go
@@ -211,31 +211,23 @@ func (h *gpbftRunner) computeNextInstanceStart(cert *certs.FinalityCertificate) 
 
 	base := ts.Timestamp()
 
-	defer func() {
-		now := h.clock.Now()
-		round0Duration := 4 * h.manifest.Gpbft.Delta
+	// Try to align instances while catching up, if configured.
+	if h.manifest.CatchUpAlignment > 0 {
+		defer func() {
+			now := h.clock.Now()
 
-		// If we were supposed to start this instance more than one GPBFT round ago, assume
-		// we're behind and align our start time to the next multiple of GPBFT round. This
-		// helps keep nodes in-sync when bootstrapping and catching up.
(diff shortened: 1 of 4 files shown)


host.go
```diff
@@ -209,15 +209,43 @@ func (h *gpbftRunner) computeNextInstanceStart(cert *certs.FinalityCertificate)
...
+		if _nextStart.Before(now.Add(-round0Duration)) {
+			// Given that we're catching up, we can assume that everyone is largely in
+			// agreement and we should decide everything in round 0. Therefore, we
+			// to finish in ~4x the maximum time it takes a message to propagate across
+			// the network.
+			//
+			// Aligning on this boundary means we can complete instances as fast as
+			// possible while still staying roughly in-sync.
+			//
+			// TODO: we should consider adding a buffer (e.g., a 2x multiplier?).
```
masih, line 230: Seems like something to add to manifest?

masih (approved):
I am thinking how we can test this on `ida` 🤔

host.go
Kubuxu: Agreed

Stebalien: Added to the manifest. It shouldn't be a breaking change (json ignores unknown fields in the manifest when unmarshaling, and we default to disabled).

PR #557, opened 2024-08-09
Introduce instance-aware validated message caching

masih:
Implement a grouped cache set, which allows grouping of a set of values, bounded growth and bulk eviction by group ID.

Once a message is validated by GPBFT, cache it to reduce the cost of re-validating duplicate messages at current or future instances.

Relates to #131


gpbft/participant.go
```diff
@@ -74,10 +83,11 @@ func NewParticipant(host Host, o ...Option) (*Participant, error) {
...
 	return &Participant{
-		options:    opts,
-		host:       host,
-		committees: make(map[uint64]*committee),
-		mqueue:     newMessageQueue(opts.maxLookaheadRounds),
+		options:                  opts,
+		host:                     host,
+		committees:               make(map[uint64]*committee),
+		mqueue:                   newMessageQueue(opts.maxLookaheadRounds),
+		alreadyValidatedMessages: caching.NewGroupedSet(10, 15_000),
```
masih, line 90: Thoughts on default values for these?

masih:
This should fix the failing fuzz test: https://github.com/filecoin-project/go-f3/pull/562

Kubuxu (approved):
LGTM, the only comment I would have is to add logs for group evictions and flops.

PR #560, opened 2024-08-10
Cache valid justifications to avoid duplicate justification validation

masih:
Use the same message cache to store valid justifications and refactor
validator to lookup cached values before validating a justification.

Bump the default set size from 15K to 25K since the same cache is used
for both messages and justifications.

Relates to #131


gpbft/metrics.go
```diff
@@ -35,8 +35,11 @@ var (
 	}
 	attrSkipToRound  = attribute.String("to", "round")
 	attrSkipToDecide = attribute.String("to", "decide")
-	attrCacheHit     = attribute.String("status", "hit")
-	attrCacheMiss    = attribute.String("status", "miss")
+
+	attrCacheHit               = attribute.String("status", "hit")
+	attrCacheMiss              = attribute.String("status", "miss")
```
Stebalien, line 40: nit: I'd use "cache" as the key here instead of status (e.g., in case we have other metrics that can have "status" in addition to cache hits/misses).

internal/caching/set.go
```diff
@@ -66,7 +76,25 @@ func (ss *Set) ContainsOrAdd(v []byte) bool {
...
+	if err != nil {
+		return [32]byte{}, err
+	}
+	_, err = hasher.Write(v)
+	if err != nil {
+		return [32]byte{}, err
+	}
+	keyB := hasher.Sum(nil)
+	var key [32]byte
+	copy(key[:], keyB)
```
Stebalien, line 93:
```suggestion
	var key [32]byte
	keyB := hasher.Sum(key[:0])
	copy(key[:], keyB)
```

If we do it this way, we can _in theory_ remove the final copy. But I'd just leave it there.

gpbft/gpbft.go
```diff
@@ -467,11 +467,23 @@ func (i *instance) tryCurrentPhase() error {
...
+// ValidateMessage checks message validity, including justification and
+// signatures. An invalid message can never become valid, so may be dropped. This
+// is a pure function and does not modify its arguments. It must be safe for
+// concurrent use.
+//
+// The caller may optionally specify CachedJustificationChecker to shortcut
+// validation of justifications. If the checker is non-nil, then the validation
+// will check if the justification is present and if so considers the
+// justification valid. Otherwise, it fully validates the justification.
+func ValidateMessage(powerTable *PowerTable, beacon []byte, host Host, msg *GMessage, isJustificationCached CachedJustificationChecker) error {
```
Stebalien, line 486: This is getting ugly (and was already ugly due to the number of arguments). Let's just inline it in `Participant.ValidateMessage`, that lets us keep all the caching in one place.

masih: it is indeed ugly; will do. 🙏

internal/caching/set.go
```diff
@@ -66,7 +76,25 @@ func (ss *Set) ContainsOrAdd(v []byte) bool {
...
+	_, err = hasher.Write(v)
+	if err != nil {
+		return [32]byte{}, err
+	}
+	var key [32]byte
+	keyB := hasher.Sum(key[:0])
+	copy(key[:], keyB)
+	// TODO: using blake's built-in keys means we lose the nice [32]byte return types
+	//       from Sum256. Consider pooling byte slices or other .Sum trickery to reduce
+	//       allocations.
```
Stebalien, line 96: Well, we are doing "trickery" here. But, `key` _does_ get moved to the stack, unfortunately.

gpbft/participant.go
```diff
@@ -155,33 +162,209 @@ func (p *Participant) ValidateMessage(msg *GMessage) (valid ValidatedMessage, er
...
+	//       caching, metrics etc.
+	//       See: https://github.com/filecoin-project/go-f3/issues/561
+
 	var buf bytes.Buffer
-	var marshalled bool
-	err = msg.MarshalCBOR(&buf)
-	if err != nil {
+	var cacheMessage bool
+	if err := msg.MarshalCBOR(&buf); err != nil {
 		log.Warnw("failed to marshal message for cache check", "err", err)
```
Stebalien, line 172: this is an error

Stebalien (approved):

internal/caching/set.go
```diff
@@ -66,7 +76,25 @@ func (ss *Set) ContainsOrAdd(v []byte) bool {
...
+}
+
+func (ss *Set) newKey(namespace, v []byte) ([32]byte, error) {
+	hasher, err := blake2b.New(blake2b.Size256, namespace)
+	if err != nil {
+		return [32]byte{}, err
+	}
+	_, err = hasher.Write(v)
+	if err != nil {
+		return [32]byte{}, err
```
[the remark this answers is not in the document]
masih: I am not sure; I'll have a think about this one 🙏

gpbft/participant.go
masih: As in error log level, or should return error?

If former, I used warn to avoid noise in logs, since caching is not a critical part of validation.
If for whatever reason this error occurs, the user would see about 15k of these per epoch.
Having said that I don't know how likely it is for this error to occur.

If latter, the rationale for proceeding with validation is that whether marshalling succeeds or not we can still validate the message. Just a bit slower and more resource intensive. 

WDYT? Does that make sense?

Stebalien: I'd just log with an Error. If this fails, something is really wrong.

internal/caching/set.go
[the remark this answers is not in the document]
Stebalien: Can leave it for now, none of my comments are critical.

masih:
The fuzz failures are due to tests taking too long to complete. We either need to reduce the iterations or use bigger runners. the latter is IMHO the way to go, along with long running fuzz tests.

PR #562, opened 2024-08-11
Restore default GOGC for CI fuzz tests

masih:
Remove the GOGC override for CI fuzz tests, because:
 * without it fuzz tests run faster,
 * we are close to the runner's memory limit, and we get false positive failures caused bo GitHub Actions SIGTERM if we go over the limit.


masih:
Compared to `main` fuzz tests run about 1m faster.

Kubuxu (approved):

PR #563, opened 2024-08-11
Use a fixed version of golangci-lint and fix lint issues

masih:
Although the CI uses staticcheck only, the team's preference is to continue and use golangci in `make lint`.

Fix the version of golangci-lint used for a more consistent checking, and fix the golangci lint issues not picked up by CI.

### DIFF
--- a/certexchange/client.go
+++ b/certexchange/client.go
@@ -185,7 +185,7 @@ func (c *Client) Request(ctx context.Context, p peer.ID, req *Request) (_rh *Res
 			}
 		}
 		return nil
-	}()
+	}() //nolint:errcheck
 	return &resp, ch, nil
 }
 
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 SHELL := /usr/bin/env bash
 
+GOLANGCILINT = go run github.com/golangci/golangci-lint/cmd/golangci-lint@v1.59.1
 
 all: generate test fuzz lint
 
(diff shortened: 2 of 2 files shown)


Kubuxu:
I would prefer to continue using golangci-lint and add it to CI. AFAIK it covers staticcheck and more.

Kubuxu (approved):

certexchange/client.go
```diff
@@ -137,54 +137,56 @@ func (c *Client) Request(ctx context.Context, p peer.ID, req *Request) (_rh *Res
 	// Copy/replace the cancel func so exiting the request doesn't cancel it.
 	cancelReq := cancel
 	cancel = nil
-	go func() (_err error) {
-		defer func() {
-			if perr := recover(); perr != nil {
-				_err = fmt.Errorf("panicked while receiving certificates from peer %s: %v\n%s", p, perr, string(debug.Stack()))
-				log.Error(err)
+	go func() {
+		_ = func() (_err error) {
```
Stebalien, line 141: Let's find a way to nolint this.

Stebalien:
Honestly, I'm with @masih here. I'd just drop golangci-lint as it's rarely helpful.

masih:
> Honestly, I'm with @masih here. I'd just drop golangci-lint as it's rarely helpful.

Maybe we would revisit this later. For now at least `make lint` uses a fixed version and does not return false negatives. Merging.

Issue #547, opened 2024-08-08
Introduce CBOR limits for public key sizes


Stebalien:
Makes it easier to ensure that we have the correct size.

Stebalien:
Alternatively, we can just add cbor-gen limits.

PR #564, opened 2024-08-12
Use fixed size byte array for public key

masih:
GPBFT used BLS public keys which are fixed to 48 bytes. Use fixed size array to represent it in code and reflect the changes across the repo.

Fixes https://github.com/filecoin-project/go-f3/issues/547

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,3 +38,7 @@ build: f3
 f3:
 	go build ./cmd/f3
 .PHONY: f3
+
+gen:
+	go generate ./...
+.PHONY: gen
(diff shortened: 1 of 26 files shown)


Stebalien:
Hm. So... looking at this, we should probably be passing keys around by-pointer (in function calls) to avoid copying 48 bytes each time.

masih:
As discussed with the team, closing in favour of cbor-gen limit: https://github.com/filecoin-project/go-f3/issues/547#issuecomment-2276021809

PR #568, opened 2024-08-13
Add GPBFT backoff to manifest

Stebalien:
This adds rebroadcast backoff config to the manifest. Unfortunately, this will require a network protocol bump if we actually want to change this.

### DIFF
--- a/manifest/manifest.go
+++ b/manifest/manifest.go
@@ -30,15 +30,12 @@ var (
 	}
 
 	DefaultGpbftConfig = GpbftConfig{
-		Delta:                3 * time.Second,
-		DeltaBackOffExponent: 2.0,
-		MaxLookaheadRounds:   5,
-	}
-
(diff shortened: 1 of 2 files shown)


manifest/manifest.go
```diff
@@ -288,15 +315,5 @@ func PubSubTopicFromNetworkName(nn gpbft.NetworkName) string {
 }
 
 func (m *Manifest) GpbftOptions() []gpbft.Option {
-	if m.Gpbft == (GpbftConfig{}) {
-		return DefaultGpbftOptions
-	}
-	var opts []gpbft.Option
-	if m.Gpbft.Delta != 0 {
-		opts = append(opts, gpbft.WithDelta(m.Gpbft.Delta))
-	}
```
Stebalien, line 297: removed because this wouldn't have passed validation anyways.

masih (approved):

PR #569, opened 2024-08-13
Improve logging around rebroadcast

masih:
Add logging to indicate when rebroadcast timeout is set and when rebroadcast is successful along with the number of messages rebroadcasted.


gpbft/gpbft.go
```diff
@@ -869,6 +869,7 @@ func (i *instance) tryRebroadcast() error {
 		}
 		i.rebroadcastTimeout = rebroadcastTimeoutOffset.Add(i.participant.rebroadcastAfter(0))
 		i.participant.host.SetAlarm(i.rebroadcastTimeout)
+		log.Debugw("Initial rebroadcast scheduled", "at", i.rebroadcastTimeout)
```
Stebalien, line 872: tracer?

Stebalien (approved):

PR #571, opened 2024-08-13
Release v0.1.0

Stebalien:


version.json
```diff
@@ -1,3 +1,3 @@
 {
-  "version": "v0.0.7"
+  "version": "v0.1.0"
```
masih, line 2: I'd be grateful if you could write a commit message about why the minor upgrade, i.e. Manifest protocol version bump, breaking changes, etc. etc. 🙏

masih (approved):
I am not sure if CI would make the tag on the right commit if this branch is not up to date with `main`; just something to watch out for.

PR #573, opened 2024-08-14
Add topic scoring & reduce pubsub spam

Stebalien:
1. Add topic scoring to both the GPBFT topic and the manifest topic.
2. Use hashing for message IDs in the manifest network and increase the rebroadcast time.
3. In the manifest validator, drop messages with old sequence numbers.


masih (approved):

Stebalien:
Bump the pubsub topic versions at the same time?

Kubuxu (approved):

PR #578, opened 2024-08-22
Fix ticket handling, switch to exponential ticket election

Kubuxu:


gpbft/gpbft.go
```diff
@@ -1304,10 +1299,27 @@ func (c *convergeState) Receive(sender ActorID, value ECChain, ticket Ticket, ju
 	c.senders[sender] = struct{}{}
 	key := value.Key()
 
-	// Keep only the first justification and ticket received for a value.
-	if _, found := c.values[key]; !found {
-		c.values[key] = ConvergeValue{Chain: value, Justification: justification}
-		c.tickets[key] = append(c.tickets[key], ConvergeTicket{Sender: sender, Ticket: ticket})
+	senderPower, _ := table.Get(sender)
+	// Keep only the first justification and best ticket
+	if v, found := c.values[key]; !found {
```
Stebalien, line 1304: Do we even need to keep multiple values? Instead, can we just keep the best ticket value (or, possibly, the best ticket for the best candidate)?

gpbft/gpbft.go
```diff
@@ -1320,26 +1332,19 @@ func (c *convergeState) FindMaxTicketProposal(table PowerTable) ConvergeValue {
...
-		for _, ticket := range c.tickets[key] {
-			senderPower, _ := table.Get(ticket.Sender)
-			ticketHash := blake2b.Sum256(ticket.Ticket)
-			ticketAsInt := new(big.Int).SetBytes(ticketHash[:])
-			weightedTicket := new(big.Int).Mul(ticketAsInt, big.NewInt(int64(senderPower)))
-			if maxTicket == nil || weightedTicket.Cmp(maxTicket) > 0 {
-				maxTicket = weightedTicket
-				maxValue = value
-			}
+	var bestValue ConvergeValue
```
Stebalien, line 1335: We can simplify this logic by initializing this value to our own value. Then we can skip the `c.HasSelfValue` part.

Stebalien (approved):
The float logic looks correct (although a bit complex).

gpbft/gpbft.go
Kubuxu: We could, but that will change if we do the audit recommendation.

gpbft/gpbft.go
Kubuxu: I removed the self-value and did some cleanup; please review the last commit.

gpbft/gpbft.go
Stebalien: Wasn't the audit recommendation to skip to the second-best ticket if the best ticket doesn't have a valid candidate? In that case, we can filter candidates up-front. I.e.:

1. Ignore all CONVERGE messages with values that aren't in our candidate set.
2. Take the value with the best ticket from the remaining values.

Kubuxu: yeah, I guess that could work

Stebalien (approved):

Stebalien:
(happy to iterate in a new PR)

gpbft/gpbft.go
masih: Quick sanity check: Has the audit recommendation been implemented in this PR?

Kubuxu: No it has not

PR #586, opened 2024-08-26
Reduce duplicate code in deny adversary

masih:
Introduce message matcher in deny adversary to reduce duplicate code across deny and deny-by-phase adversaries.

Note that the deny adversary symmetrically denies messages: message to or from targets will be denied if they are matched. This is a stronger adversary behaviour compared to `deny_phase` replaced by this commit, where in `deny_phase` only messages _to_ targeted actors where being denied.


Kubuxu:
We want weaker behaviour as the stronger adversary fails to reproduce the issue.

Kubuxu:
If you want a quick reproducer for the original issue:
in the `gpbft/ticket_quality.go` in `ComputeTicketQuality` return `-float64(power)`.

masih:
> If you want a quick reproducer for the original issue

Thanks Kuba, looking.

masih:
> We want the weaker behaviour as the stronger adversary fails to reproduce the issue.

OK I have parameterised the deny mode and tested every case. This is ready for another look whenever you get a chance Kuba. Thanks

Kubuxu (approved):

PR #589, opened 2024-08-26
Select next best converge value compatible with our candidate set

Kubuxu:
Based on recommendations of Least Authority

### DIFF
--- a/gpbft/gpbft.go
+++ b/gpbft/gpbft.go
@@ -544,33 +544,35 @@ func (i *instance) tryConverge() error {
 		return nil
 	}
 
-	winner := i.getRound(i.round).converged.FindMaxTicketProposal(i.powerTable, nil)
+	isValidConvergeValue := func(cv ConvergeValue) bool {
+		// If it is in candidate set
+		if i.isCandidate(cv.Chain) {
+			return true
+		}
+		// If it is not a candidate but it could possibly have been decided by another participant
+		// in the last round, consider it a candidate.
+		if cv.Justification.Vote.Step != PREPARE_PHASE {
+			return false
+		}
+		possibleDecisionLastRound := i.getRound(i.round-1).committed.CouldReachStrongQuorumFor(
+			cv.Chain.Key(), true)
+		return possibleDecisionLastRound
+	}
+
+	winner := i.getRound(i.round).converged.FindMaxTicketProposal(i.powerTable, isValidConvergeValue)
 	if !winner.IsValid() {
 		return fmt.Errorf("no values at CONVERGE")
 	}
-	possibleDecisionLastRound := i.getRound(i.round-1).committed.CouldReachStrongQuorumFor(
-		winner.Chain.Key(), true)
-	justification := winner.Justification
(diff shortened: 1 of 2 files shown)


gpbft/gpbft.go
```diff
@@ -544,35 +544,37 @@ func (i *instance) tryConverge() error {
...
+		// If it is in candidate set
+		if i.isCandidate(cv.Chain) {
+			return true
+		}
+		// If it is not a candidate but it could possibly have been decided by another participant
+		// in the last round, consider it a candidate.
+		if cv.Justification.Vote.Step != PREPARE_PHASE {
+			return false
+		}
+		possibleDecisionLastRound := i.getRound(i.round-1).committed.CouldReachStrongQuorumFor(
```
Stebalien, line 557: Let's call `i.getRound(i.round - 1)` outside the function.

gpbft/gpbft.go
```diff
@@ -544,35 +544,37 @@ func (i *instance) tryConverge() error {
...
+			cv.Chain.Key(), true)
+		return possibleDecisionLastRound
+	}
+
+	winner := i.getRound(i.round).converged.FindMaxTicketProposal(i.powerTable, isValidConvergeValue)
```
Stebalien, line 562: nit: rename to `FindBestTicketProposal`

gpbft/gpbft.go
```diff
@@ -1332,16 +1331,17 @@ func (c *convergeState) Receive(sender ActorID, table PowerTable, value ECChain,
 
 // FindMaxTicketProposal finds the value with the best ticket, weighted by
 // sender power. Returns an invalid (zero-value) ConvergeValue if no converge is found.
```
Stebalien, line 1333: document filter

gpbft/gpbft.go
```diff
@@ -1332,16 +1331,17 @@ func (c *convergeState) Receive(sender ActorID, table PowerTable, value ECChain,
...
-func (c *convergeState) FindMaxTicketProposal(table PowerTable) ConvergeValue {
+func (c *convergeState) FindMaxTicketProposal(table PowerTable, filter func(ConvergeValue) bool) ConvergeValue {
```
Stebalien, line 1334: I still think we should filter when receiving proposals instead of filtering here.

Stebalien (approved):

gpbft/gpbft.go
Kubuxu: I thought about it, but we should not because of how COMMIT works (the commit round is always active).

Stebalien: This is CONVERGE. We skip here: https://github.com/filecoin-project/go-f3/blob/71f7e0d13749f6f4368608640a7cd506ecf3ab3b/gpbft/gpbft.go#L358

Kubuxu: The thing that dictates sway behaviour is the state of previous commit round.

Stebalien: Kuba is right...

Stebalien (approved):

Issue #591, opened 2024-08-26
Handle QUALITY censorship


Stebalien:
https://github.com/filecoin-project/FIPs/discussions/809#discussioncomment-10424988
```[tasklist]
### Tasks
- [x] Include QUALITY messages in rebroadcast
- [ ] Accept quality messages in every round/phase
- [ ] Run a message-less quality phase every round after the first round
```

PR #597, opened 2024-08-27
Include QUALITY messages in rebroadcast

masih:
To improve censorship resistance and avoid lack of progress due to partially disseminated QUALITY messages include them in rebroadcast.

Part of #591

### DIFF
--- a/gpbft/gpbft.go
+++ b/gpbft/gpbft.go
@@ -1374,29 +1374,54 @@ func newBroadcastState() *broadcastState {
 // record stores messages that are required should rebroadcast becomes necessary.
 // The messages stored depend on the current progress of instance. If the
 // instance progresses to DECIDE then only the decide message will be recorded.
-// Otherwise, all messages except QUALITY from the current and previous rounds
+// Otherwise, all messages including QUALITY from the current and previous rounds
 // (i.e. the latest two rounds) are recorded.
 //
 // Note, the messages recorded are more than what FIL-00896 strictly requires at
 // the benefit of less reliance on rebroadcast and reduction in the possibility
 // of participants getting stuck.
+//
+// The rationale for including QUALITY messages as part of rebroadcast is to
+// improve censorship resistance at the face of unreliable message delivery,
+// which can ultimately result in lack of instance progress for extended periods
+// of time. See:
+//   - https://github.com/filecoin-project/FIPs/discussions/809#discussioncomment-10409902
+//   - https://github.com/filecoin-project/FIPs/discussions/809#discussioncomment-10424988
 func (bs *broadcastState) record(mb *MessageBuilder) {
 	switch mb.Payload.Step {
 	case DECIDE_PHASE:
 		// Clear all previous messages, as only DECIDE message need to be rebroadcasted.
 		// Note that DECIDE message is not associated to any round, and is always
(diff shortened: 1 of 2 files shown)


gpbft/gpbft.go
```diff
@@ -1374,29 +1374,54 @@ func newBroadcastState() *broadcastState {
...
 	switch mb.Payload.Step {
 	case DECIDE_PHASE:
 		// Clear all previous messages, as only DECIDE message need to be rebroadcasted.
 		// Note that DECIDE message is not associated to any round, and is always
 		// broadcasted using round zero.
-		bs.messagesByRound = make(map[uint64][]*MessageBuilder)
+		clear(bs.messagesByRound)
 		bs.messagesByRound[0] = []*MessageBuilder{mb}
-	case COMMIT_PHASE, PREPARE_PHASE, CONVERGE_PHASE:
+	case QUALITY_PHASE, CONVERGE_PHASE, PREPARE_PHASE, COMMIT_PHASE:
```
Stebalien, line 1398: would it be simpler to handle quality messages separately? Really, decide messages as well? Neither belong to any given round.

masih: I had a bunch of discussions in previous reviews on this. Originally I did implement it using dedicated fields. But it ended up being a map as that seemed simpler. 

I plan to change it back to that approach as part of making rebroadcasted messages more selective.

I am happy to reflect on your comment in this PR if it's a blocker. Otherwise I leave it to future PRs if the code here is correct. 

What is your preference?

Stebalien: It's not, I'll review as-is.

gpbft/gpbft.go
```diff
@@ -1374,29 +1374,54 @@ func newBroadcastState() *broadcastState {
...
+				var found bool
+				for i, mb := range bs.messagesByRound[0] {
+					if mb.Payload.Step == QUALITY_PHASE {
+						bs.messagesByRound[0] = bs.messagesByRound[0][i : i+1 : 1]
+						found = true
+						break
+					}
+				}
+				if !found {
+					log.Warn("No QUALITY message found for round 0 while trimming rebroadcast messages")
```
Stebalien, line 1414: This isn't necessarily an error right? Can't we skip quality if we join late?

Stebalien (approved):

gpbft/gpbft.go
masih: [begin instance](https://github.com/filecoin-project/go-f3/blob/f85100a1004339086dd5520a27fe5568707e36b1/gpbft/participant.go#L430) always [begins with quality](https://github.com/filecoin-project/go-f3/blob/f85100a1004339086dd5520a27fe5568707e36b1/gpbft/gpbft.go#L266). Even if a participant joins late, they would always [broadcast a QUALITY](https://github.com/filecoin-project/go-f3/blob/f85100a1004339086dd5520a27fe5568707e36b1/gpbft/gpbft.go#L477) first unless I have missed something.

So i think if we hit this line there is either a bug or the code has changed significantly enough for this condition to be revisited.

Does that make sense?

Stebalien: Yeah, you're right. We won't even receive any messages till we send our quality message.

PR #604, opened 2024-08-29
Use a mask for BLS aggregation and improve caching

### LINKED ISSUE #592: Change BLS signature/public key aggregation enable better caching
Right now we're not taking advantage of the bitfield mask when aggregating public keys. I.e., each time we aggregate public keys we use an all-ones bitfield and cannot re-use any of the results of the aggregation.

Instead, we should:

1. Use the full power table when aggregating public keys, using the bitfield to select the active keys. This will cost a bit more up-front, but we'll only pay that cost once per instance.
2. Cache the `coeff*pubkey+pubkey` for every public key for the rest of the instance (cache https://github.com/drand/kyber/blob/94dae51d79b4b0c2d2a9b9cc382b864cf3537783/sign/bdn/bdn.go#L183-L185).

This is the one N^2 cryptographic operation we do, so it should make F3 scale significantly better.

Stebalien:
Design:

1. Include all public keys present in the power table in the BDN mask.
2. Only select the relevant ones (via the bitmask) when aggregating signatures.
3. Pre-compute the terms used in public key aggregation, making it _much_ faster (it's a simple summation).
4. Partially pre-compute the coefficients used in signature aggregation. It's slightly faster (can avoid some hashing) but not that much faster because we still need to compute the terms (multiply the coefficients by the signatures being aggregated).

There are three commits here and this PR should be merged by rebase, not squash.

1. The first commit switches from drand's kyber fork back to dedis.
2. The second imports the BDN and GNARK modules, along with the following dedis PRs:
    1. https://github.com/dedis/kyber/pull/546
    2. https://github.com/dedis/kyber/pull/551
    3. https://github.com/dedis/kyber/pull/553
3. The final commit actually makes use of the caching mask.

The code imported in commit 2 should be (nearly) identical to the code in the `bdn` and `gnark` packages in https://github.com/Stebalien/kyber/tree/f3.

fixes #592


gpbft/api.go
```diff
@@ -100,15 +100,27 @@ type SigningMarshaler interface {
...
-	// Aggregates signatures from a participants.
-	Aggregate(pubKeys []PubKey, sigs [][]byte) ([]byte, error)
-	// VerifyAggregate verifies an aggregate signature.
+	// Return an Aggregate that can aggregate and verify aggregate signatures made by the given
+	// public keys.
+	//
 	// Implementations must be safe for concurrent use.
-	VerifyAggregate(payload, aggSig []byte, signers []PubKey) error
+	Aggregate(pubKeys []PubKey) (Aggregate, error)
 }
```
Stebalien, line 124: The interface changes. Basically:

1. You call `Verifier.Aggregate(allPublicKeys)`.
2. Then you call either `Aggregate` (name clash) or `VerifyAggregate` on the resulting `Aggregate` (naming?).

gpbft/participant.go
```diff
@@ -485,7 +485,16 @@ func (p *Participant) fetchCommittee(instance uint64) (*committee, error) {
 		if err := power.Validate(); err != nil {
 			return nil, fmt.Errorf("instance %d: %w: invalid power: %w", instance, ErrValidationNoCommittee, err)
 		}
-		comt = &committee{power: power, beacon: beacon}
+
+		// TODO: filter out participants with no effective power after rounding?
```
Stebalien, line 489: This could save us a little perf, but it won't matter much and it'll make things more complex. IMO, don't do it.

gpbft/participant.go
```diff
@@ -485,7 +485,16 @@ func (p *Participant) fetchCommittee(instance uint64) (*committee, error) {
...
+		// TODO: this is slow and under a lock, but we only want to do it once per
+		// instance... ideally we'd have a per-instance lock/once, but that probably isn't
+		// worth it.
```
Stebalien, line 492: IMO, punt till later (probably not much worse than fetching the committee under the lock).

Stebalien:
This is ready for review although still WIP until the upstream changes (https://github.com/drand/kyber/pull/61) land. Reviews there would also be helpful (only look at the last two commits, everything else is from a previous PR).

masih:
> although still WIP until the upstream changes

Woops; sorry for PR title change noise.

Kubuxu (commented):
Approach SGWM

Stebalien:
Ok, waiting on https://github.com/dedis/kyber/pull/546 now.

gpbft/participant.go
Stebalien: Nah.

internal/bls/bdn/mask.go
```diff
@@ -0,0 +1,221 @@
...
+}
+
+// CountEnabled returns the number of bit set to 1
+func (m *Mask) CountEnabled() int {
+	count := 0
+	for i := range m.publics {
+		byteIndex := i / 8
+		mask := byte(1) << uint(i&7)
+		if (m.mask[byteIndex] & mask) != 0 {
+			count++
```
Kubuxu, line 187: We could just popcount but doens't matter.

Stebalien: I'll make a PR upstream.

Stebalien: Hm. Later. I have too many stacked.

Kubuxu (approved):

Stebalien:
Merged by rebase to preserve the three separate commits.

PR #606, opened 2024-08-29
Handle late arriving QUALITY messages in every round

### LINKED ISSUE #591: Handle QUALITY censorship

masih:
Change the core GPBFT protocol to:
 1) accept QUALITY messages in ant phase or round.
 2) update candidate chains at any round or phase based on the initial proposal.

This would permit participants with late-arriving/partially delivered QUALITY messages to affect the candidates chains in CONVERGE phase to sway for the initial proposal or any of its prefixes should there be a strong quorum. No that the changes here do not update the current proposal based on late arriving QUALITY messages.

The changes above, in conjunction with rebroadcast of QUALITY messages introduced in #597 should significantly reduce the likelihood of lack of progress due to partially seen QUALITY messages. See:
 - https://github.com/filecoin-project/FIPs/discussions/809
 #discussioncomment-10409902

As part of the changes introduced by this commit, the data structure used by the instance to store candidate chains is changed to a map of chain keys for a faster candidate set update (O(1)), and a more concise way to assure uniqueness of the set of candidates.

Fixes #591

### DIFF
--- a/gpbft/gpbft.go
+++ b/gpbft/gpbft.go
@@ -190,11 +190,12 @@ type instance struct {
 	// The value to be transmitted at the next phase, which may be bottom.
 	// This value may change away from the proposal between phases.
 	value ECChain
-	// The set of values that are acceptable candidates to this instance.
-	// This includes the base chain, all prefixes of proposal that found a strong quorum
-	// of support in the QUALITY phase, and any chains that could possibly have been
-	// decided by another participant.
-	candidates []ECChain
+	// candidates contains a set of values that are acceptable candidates to this
+	// instance. This includes the base chain, all prefixes of proposal that found a
+	// strong quorum of support in the QUALITY phase or late arriving quality
+	// messages, including any chains that could possibly have been decided by
+	// another participant.
+	candidates map[ChainKey]struct{}
 	// The final termination value of the instance, for communication to the participant.
 	// This field is an alternative to plumbing an optional decision value out through
 	// all the method calls, or holding a callback handle to receive it here.
@@ -238,8 +239,10 @@ func newInstance(
 		proposal:         input,
 		broadcasted:      newBroadcastState(),
 		value:            ECChain{},
-		candidates:       []ECChain{input.BaseChain()},
-		quality:          newQuorumState(powerTable),
+		candidates: map[ChainKey]struct{}{
+			input.BaseChain().Key(): {},
+		},
+		quality: newQuorumState(powerTable),
 		rounds: map[uint64]*roundState{
 			0: newRoundState(powerTable),
 		},
@@ -351,11 +354,9 @@ func (i *instance) receiveOne(msg *GMessage) (bool, error) {
 	if i.phase == TERMINATED_PHASE {
 		return false, nil // No-op
 	}
-	// Ignore QUALITY messages after exiting the QUALITY phase.
 	// Ignore CONVERGE and PREPARE messages for prior rounds.
 	forPriorRound := msg.Vote.Round < i.round
-	if (msg.Vote.Step == QUALITY_PHASE && i.phase != QUALITY_PHASE) ||
-		(forPriorRound && msg.Vote.Step == CONVERGE_PHASE) ||
+	if (forPriorRound && msg.Vote.Step == CONVERGE_PHASE) ||
 		(forPriorRound && msg.Vote.Step == PREPARE_PHASE) {
 		return false, nil
 	}
@@ -373,8 +374,14 @@ func (i *instance) receiveOne(msg *GMessage) (bool, error) {
 	msgRound := i.getRound(msg.Vote.Round)
 	switch msg.Vote.Step {
 	case QUALITY_PHASE:
-		// Receive each prefix of the proposal independently.
+		// Receive each prefix of the proposal independently, which is accepted at any
+		// round/phase.
 		i.quality.ReceiveEachPrefix(msg.Sender, msg.Vote.Value)
+		// If the instance has surpassed QUALITY phase, update the candidates based
+		// on possible quorum of input prefixes.
+		if i.phase != QUALITY_PHASE {
+			return true, i.updateCandidatesFromQuality()
+		}
 	case CONVERGE_PHASE:
 		if err := msgRound.converged.Receive(msg.Sender, i.powerTable, msg.Vote.Value, msg.Ticket, msg.Justification); err != nil {
 			return false, fmt.Errorf("failed processing CONVERGE message: %w", err)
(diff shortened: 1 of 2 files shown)


gpbft/gpbft.go
```diff
@@ -544,6 +548,12 @@ func (i *instance) tryConverge() error {
 		return nil
 	}
 
+	// Check strong quorum values from late arriving QUALITY messages and update
+	// candidates accordingly.
+	if foundQualityQuorum := i.quality.HasStrongQuorumFor(i.initialProposal.Key()); foundQualityQuorum {
+		i.addCandidatePrefixes(i.initialProposal)
+	}
```
Stebalien, line 555: We need to walk all prefixes of `initialProposal` and call `addCandidatePrefixes` on the first prefix with strong quorum.

gpbft/gpbft.go
```diff
@@ -373,7 +381,8 @@ func (i *instance) receiveOne(msg *GMessage) (bool, error) {
 	msgRound := i.getRound(msg.Vote.Round)
 	switch msg.Vote.Step {
 	case QUALITY_PHASE:
-		// Receive each prefix of the proposal independently.
+		// Receive each prefix of the proposal independently, which is accepted at any
+		// round/phase.
 		i.quality.ReceiveEachPrefix(msg.Sender, msg.Vote.Value)
```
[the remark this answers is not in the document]
masih: I am not sure if we should, or at least there is no harm in proceeding, because if we are in CONVERGE phase then the processing of a late-arriving QUALITY message may expand candidates that [could result in adopting proposal after converge](https://github.com/filecoin-project/go-f3/blob/c0765caf1812cb7db0e1a7e4c0759f309dc53c51/gpbft/gpbft.go#L570). 

WDYT? does that make sense?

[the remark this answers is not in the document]
Kubuxu: makes sense to me

gpbft/gpbft.go
```diff
@@ -544,6 +548,17 @@ func (i *instance) tryConverge() error {
 		return nil
 	}
 
+	// Find the first initial proposal prefix that has strong quorum as a result of
+	// arriving QUALITY messages and update candidates for its consecutive prefixes.
```
Kubuxu, line 552: Will this select the smallest or the largest prefix? Which one would be the correct one in either case?

gpbft/gpbft.go
```diff
@@ -555,11 +570,11 @@ func (i *instance) tryConverge() error {
 	// in the last round, consider it a candidate.
 	if !i.isCandidate(winner.Chain) && winner.Justification.Vote.Step == PREPARE_PHASE && possibleDecisionLastRound {
 		i.log("⚠️ swaying from %s to %s by CONVERGE", &i.proposal, &winner.Chain)
-		i.candidates = append(i.candidates, winner.Chain)
+		i.addCandidate(winner.Chain)
```
Kubuxu, line 573: of note: The lack of coverage here means we have no converge sway test case.

Kubuxu (approved):

gpbft/gpbft.go
masih: yep working on it

gpbft/gpbft.go
masih: It should be the longest prefix with strong quorum IIUC. I have refactored this locally which I'll push shortly.

gpbft/gpbft.go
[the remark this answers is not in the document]
masih: I have gone back and forth on this one a few times. TBD in standup.

gpbft/gpbft.go
```diff
@@ -482,31 +489,35 @@ func (i *instance) beginQuality() error {
...
+		// expire.
+		foundQuorum := i.quality.HasStrongQuorumFor(i.input.Key())
+		timeoutExpired := atOrAfter(i.participant.host.Time(), i.phaseTimeout)
+
+		if foundQuorum {
+			// Keep current proposal.
+		} else if timeoutExpired {
+			// Change proposal to the longest prefix of input with strong quorum.
+			i.proposal = i.quality.FindStrongQuorumValueForLongestPrefixOf(i.input)
+		}
```
Kubuxu, line 504: Wouldn't this be the same, regardless of whether our proposal won or just its prefix?

Kubuxu: hmm, but we probably want to keep the behaviour of: move on when our proposal wins or timeout explores, instead of changing it to move on when any proposal wins

masih: Sorry, I don't follow. Same in what sense?

masih: The behaviour of the case where current phase is QUALITY is unchanged relative to `main` unless I have missed something.

gpbft/gpbft.go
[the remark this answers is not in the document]
masih: Discussed sync. Because of the timeout in CONVERGE it makes no difference. hence keeping the exit early as suggested by Steven

masih:
Tests for this requires making ticket quality pluggable, which Is a large enough change for me to need its own PR. Hence marking this PR as ready for review. 

@Kubuxu @Stebalien  Please let me know if you'd prefer the tests to be included in this PR.

Kubuxu (approved):

gpbft/gpbft.go
```diff
@@ -373,8 +374,14 @@ func (i *instance) receiveOne(msg *GMessage) (bool, error) {
...
 	switch msg.Vote.Step {
 	case QUALITY_PHASE:
-		// Receive each prefix of the proposal independently.
+		// Receive each prefix of the proposal independently, which is accepted at any
+		// round/phase.
 		i.quality.ReceiveEachPrefix(msg.Sender, msg.Vote.Value)
+		// If the instance has surpassed QUALITY phase, update the candidates based
+		// on possible quorum of input prefixes.
+		if i.phase != QUALITY_PHASE {
+			return true, i.tryQuality()
```
Stebalien, line 383: I'm not a fan:

1. Here, we check "if not quality, call try quality".
2. Then, in "try quality", we have a completely separate code path for "not quality".

That should just be an entirely different function.

masih: That makes sense

gpbft/gpbft.go
```diff
@@ -797,10 +807,23 @@ func (i *instance) skipToRound(round uint64, chain ECChain, justification *Justi
...
-		}
+	_, exists := i.candidates[c.Key()]
+	return exists
+}
+
+func (i *instance) addCandidatePrefixes(c ECChain) bool {
+	var addedAny bool
+	for l := range c {
+		addedAny = i.addCandidate(c.Prefix(l)) || addedAny
+	}
```
Stebalien, line 818: nit: I'd do this the other way and stop when we find the first prefix. I.e.:

```go
for i := len(c)-1; i > 0 && !addedAny; i-- {
    addedAny ||= i.addCandidate(c.Prefix(l))
}
```

Stebalien: (not sure if `||=` is a thing...)

gpbft/gpbft.go
```diff
@@ -1201,30 +1224,21 @@ func (q *quorumState) FindStrongQuorumFor(key ChainKey) (QuorumResult, bool) {
...
-			panic(fmt.Sprintf("multiple chains of length %d with strong quorum", prevLength))
+	longestPrefix := preferred.BaseChain()
+	for _, cp := range q.chainSupport {
+		if cp.hasStrongQuorum &&
+			preferred.HasPrefix(cp.chain) &&
+			len(cp.chain) > len(longestPrefix) {
+			longestPrefix = cp.chain
 		}
-		prevLength = len(v)
 	}
```
Stebalien, line 1240: Don't we add chain support for all prefixes? We should iterate over all prefixes of `preferred` starting from the end, right?

gpbft/gpbft.go
masih: Not in golang

gpbft/gpbft.go
```diff
@@ -798,10 +812,23 @@ func (i *instance) skipToRound(round uint64, chain ECChain, justification *Justi
...
-			return true
-		}
+	_, exists := i.candidates[c.Key()]
+	return exists
+}
+
+func (i *instance) addCandidatePrefixes(c ECChain) bool {
+	var addedAny bool
+	for l := len(c) - 1; l > 0 && !addedAny; l-- {
+		addedAny = i.addCandidate(c.Prefix(l)) || addedAny
```
Stebalien, line 822:
```suggestion
		addedAny = i.addCandidate(c.Prefix(l)) 
```

Stebalien: We can also just return early.

Stebalien (approved):

PR #607, opened 2024-08-29
Propagate DECIDE messages from the previous instance

### LINKED ISSUE #579: Avoid dropping decides from the previous instance
We currently drop all messages from the previous instance, but we should consider allowing decides through. Otherwise, some nodes may not be able to make progress until they can use the certificate exchange.

Stebalien:
fixes #579

### DIFF
--- a/gpbft/participant.go
+++ b/gpbft/participant.go
@@ -158,7 +158,7 @@ func (p *Participant) ValidateMessage(msg *GMessage) (valid ValidatedMessage, er
 		}
 	}()
 
-	comt, err := p.fetchCommittee(msg.Vote.Instance)
+	comt, err := p.fetchCommittee(msg.Vote.Instance, msg.Vote.Step)
 	if err != nil {
 		return nil, err
 	}
@@ -441,7 +441,7 @@ func (p *Participant) beginInstance() error {
 		return fmt.Errorf("invalid canonical chain: %w", err)
 	}
 
-	comt, err := p.fetchCommittee(p.currentInstance)
+	comt, err := p.fetchCommittee(p.currentInstance, INITIAL_PHASE)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


Stebalien:
The fuzzer now appears to be hanging pretty consistently. Not sure what's up with that but I'll try to debug it a bit.

masih:
> The fuzzer now appears to be hanging pretty consistently.

I suspect it's the runners. I'll switch to IPDX runners with extended fuzz time to investigate further.

gpbft/participant.go
```diff
@@ -542,6 +552,16 @@ func (p *Participant) Describe() string {
 	return p.gpbft.Describe()
 }
 
+func (p *Participant) tracing() bool {
+	return p.tracer != nil
+}
+
+func (p *Participant) trace(format string, args ...any) {
+	if p.tracer != nil {
```
masih, line 560:
```suggestion
	if p.tracing() {
```

gpbft/participant.go
```diff
@@ -542,6 +552,16 @@ func (p *Participant) Describe() string {
 	return p.gpbft.Describe()
 }
 
+func (p *Participant) tracing() bool {
```
masih, line 555: Subjective: `tracingEnabled`?

masih (commented):
IIUC, the sole purpose of this PR (and #579) is to **continue to propagate DECIDEs via pubsub**, because invalid messages will not be propagated at _pubsub layer_.  These changes make no difference to the operation of the core GPBFT protocol, because on `RecieveMessage` [all messages from previous instances are dropped](https://github.com/filecoin-project/go-f3/blob/50dd5c1801c0e575e53c1407ec825cdc0e6c022d/gpbft/participant.go#L386-L390) at _GPPBFT layer_.

Is my understanding correct @Stebalien? asking since #579 does't explicitly mention the scope of change is for pubsub layer only.

If so, this LGTM 👍

Kubuxu (approved):

Stebalien:
> These changes make no difference to the operation of the core GPBFT protocol, because on RecieveMessage [all messages from previous instances are dropped](https://github.com/filecoin-project/go-f3/blob/50dd5c1801c0e575e53c1407ec825cdc0e6c022d/gpbft/participant.go#L386-L390) at GPPBFT layer

Yes.

masih (approved):

Stebalien:
Fuzzing is failing on 'main' as well.

Stebalien:
At least on my machine...

PR #610, opened 2024-08-30
Use self-hosted runners for fuzzing

masih:
Testing IPDX self-hosted runners for fuzz tests.

### DIFF
--- a/.github/workflows/go-fuzz.yml
+++ b/.github/workflows/go-fuzz.yml
@@ -12,7 +12,7 @@ on:
 jobs:
   fuzz:
     name: go-fuzz
-    runs-on: ubuntu-latest
+    runs-on: [ self-hosted, linux, x64, 2xlarge ]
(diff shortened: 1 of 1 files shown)


Stebalien:
Also see https://github.com/filecoin-project/go-f3/pull/608/commits/3816904e9b7c537884402becab01af98b7c49304. I'm not sure why, but the specific failing fuzz test is repeatedly clearing this cache.

Stebalien:
Yeah, that commit makes things 2x faster...

masih:
> Yeah, that commit makes things 2x faster...

Interesting... Also https://github.com/filecoin-project/go-f3/pull/612

masih:
Using IPDX runners is looking hopeless, probably due to heavy demand? I might set up our own runners on K8S (with finer instance size selection). They should be cheap as they would get torn down when not in use

PR #611, opened 2024-08-30
Replace cache set maps instead of clearing them

Stebalien:
Clearing seems to be very slow, it's making
`FuzzHonestMultiInstance_SyncAgreement` take 2x longer. This is probably
because the keys are large and by-value? I'm not sure, but the cost of
re-allocating these maps shouldn't be huge anyways...


masih (approved):
But why?

I don't love this 🙈

Stebalien:
I don't love this either but... it shouldn't matter in practice and I have a feeling that there may be some lurking perf issues in the new `clear` function.

Kubuxu (approved):

Issue #614, opened 2024-08-30
Batch validate justification + message signature


Stebalien:
In general, batch validation can be tricky because any invalid signature will make the validation fail. However, we _can_ safely batch validate the message signature plus the justification signature, because either being incorrect means we treat the entire message as incorrect.

Stebalien:
Yeah, this looks worth it.

PR #615, opened 2024-08-30
[WIP] Demo batch verification function

Stebalien:
part of #614

### DIFF
--- a/blssig/verifier.go
+++ b/blssig/verifier.go
@@ -8,6 +8,9 @@ import (
 	"sync"
 
 	"go.dedis.ch/kyber/v4"
+	"go.dedis.ch/kyber/v4/pairing"
+	"go.dedis.ch/kyber/v4/sign"
+	"go.dedis.ch/kyber/v4/sign/bls"
 	"go.opentelemetry.io/otel/metric"
 
 	"github.com/filecoin-project/go-f3/gpbft"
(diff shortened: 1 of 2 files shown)


Stebalien:
cc @Kubuxu

blssig/verifier.go
```diff
@@ -103,5 +106,43 @@ func (v *Verifier) Verify(pubKey gpbft.PubKey, msg, sig []byte) (_err error) {
...
+
+	points := make([]kyber.Point, len(msgs))
+	for i, pubKey := range pubKeys {
+		var err error
+		points[i], err = v.pubkeyToPoint(pubKey)
+		if err != nil {
+			return fmt.Errorf("unarshalling public key: %w", err)
+		}
+	}
+	aggSig, err := v.blsScheme.AggregateSignatures(sigs...)
```
Kubuxu, line 142: I don't think this is generally safe. It might be secure in this case as the message and justification are guaranteed to be unique.
No, wait, they are not. Then, yeah, this is not safe. The BatchVerify method requires that the messages be unique. There is an alternative method for BatchVerifying that introduces randomness to both the message and signature, making them linearly independent. I could write that.
It does not affect the overall scheme and is just a verification time tweak.

Kubuxu: Hmm, I guess the BatchVerify will check that messages are unique. So it might be safe.

Stebalien:
At the moment this is only going from 4 pairings to 3, and adding some additional work into the mix for aggregation. In theory, we can go even faster but kyber currently doesn't allow for the optimizations we'd need in that case.

I'm going to close this for now. It's probably easiest to simply re-implement the batch verification here, but this PR isn't going to get us any closer anyways.

Issue #237, opened 2024-05-16
Test invalid converge cases


Stebalien:
It looks like we're not testing the following converge cases:

- [x] Swaying vote in converge round (we never hit `⚠️ swaying from %s to %s by CONVERGE`. I'm not actually sure what this means, but we should be hitting this log, right?
- [x] Try to converge to bottom with a bogus justification (not legal in the protocol, but we should test this case to make sure we correctly reject it).

Issue #616, opened 2024-09-02
Check for equivocations across messages received and justification


masih:
Messages with justification contain evidence of prior messages (either COMMIT or PREPARE) from current or previous round. Currently it is possible for nodes to equivocate across them, e.g. send PREPARE for one value and change it in the justification of consecutive COMMIT.

Fix:
* https://github.com/filecoin-project/go-f3/issues/342

masih:
Update: the fix for this is not exactly getting #342 done. It is actually a smaller subset of related changes. I will approach fixing this specific issue first before working on #342. Because, #342 would require revise of how justifications vs. individual message signatures are stored in the quorum state.

PR #617, opened 2024-09-02
Test sway at commit and converge for unseen candidates

masih:
Test that a participant sways for unseen candidates at `COMMIT` and `CONVERGE`.

Part of #237

### DIFF
--- a/emulator/driver.go
+++ b/emulator/driver.go
@@ -60,6 +60,12 @@ func (d *Driver) AddInstance(instance *Instance) {
 	d.require.NoError(d.host.addInstance(instance))
 }
 
+// PeekLastBroadcastRequest gets the last broadcast requested by the subject
+// participant without removing it from the pending broadcasts.
+func (d *Driver) PeekLastBroadcastRequest() *gpbft.GMessage {
+	return d.host.peekLastBroadcast()
+}
+
 func (d *Driver) DeliverAlarm() (bool, error) {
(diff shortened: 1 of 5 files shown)


Kubuxu:
The codecov claims that CONVERGE sway is not hit.

masih:
OK this is ready for another look; please let me know if you can think of a simpler way to test sway at CONVERGE. thanks

Kubuxu:
What is the source of non-determinism? I think it should be only the chain seed, which leads to beacons.

masih:
> What is the source of non-determinism?

Order of [chainsupport map](https://github.com/filecoin-project/go-f3/blob/33b0a89cde74412f58bc0bf7103438e8620d883c/gpbft/gpbft.go#L690) for committed state, and consequently candidate inclusion in `tryConverge` which determines whether we sway at CONVERGE or not.

Kubuxu:
> Order of [chainsupport map](https://github.com/filecoin-project/go-f3/blob/33b0a89cde74412f58bc0bf7103438e8620d883c/gpbft/gpbft.go#L690) for committed state

That seems counter to the comment there, as the comment states that there should be only one.

masih:
> That seems counter to the comment there, as the comment states that there should be only one.

Right. If i have not missed anything I think that would only hold if [this](https://github.com/filecoin-project/go-f3/issues/616) is fixed.

Stebalien (approved):

PR #620, opened 2024-09-02
On `tryCommit` only rebroadcast at current round / phase

### LINKED ISSUE #593: TryCommit rebroadcast behaviour is confusing COMMIT round with other round commit messages
We always accept commit messages, but TryCommit uses `i.phaseTimeout` to control some of its behaviour.
This has no side-effect apart from calling `tryRebroadcast`.

masih:
A `COMMIT` message may arrive at any round/phase. Regardless, it is handled by `tryCommit`. Refactor the logic of this function such that:
 * rebroadcast at `COMMIT` is only attempted if the instance is indeed at that phase and its corresponding round.
 * logic is easier to read with definitive switch cases, where each case is commented.

Fixes #593

### DIFF
--- a/gpbft/gpbft.go
+++ b/gpbft/gpbft.go
@@ -645,49 +645,55 @@ func (i *instance) beginCommit() {
 }
 
 func (i *instance) tryCommit(round uint64) error {
-	// Unlike all other phases, the COMMIT phase stays open to new messages even after an initial quorum is reached,
-	// and the algorithm moves on to the next round.
-	// A subsequent COMMIT message can cause the node to decide, so there is no check on the current phase.
+	// Unlike all other phases, the COMMIT phase stays open to new messages even
+	// after an initial quorum is reached, and the algorithm moves on to the next
+	// round. A subsequent COMMIT message can cause the node to decide, so there is
+	// no check on the current phase.
 	committed := i.getRound(round).committed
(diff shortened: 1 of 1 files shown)


gpbft/gpbft.go
```diff
@@ -645,49 +645,55 @@ func (i *instance) beginCommit() {
...
 	committed := i.getRound(round).committed
 	quorumValue, foundStrongQuorum := committed.FindStrongQuorumValue()
 	timedOut := atOrAfter(i.participant.host.Time(), i.phaseTimeout)
 	phaseComplete := timedOut && committed.ReceivedFromStrongQuorum()
 
-	if foundStrongQuorum && !quorumValue.IsZero() {
-		// A participant may be forced to decide a value that's not its preferred chain.
-		// The participant isn't influencing that decision against their interest, just accepting it.
+	switch {
+	case foundStrongQuorum && !quorumValue.IsZero():
```
Kubuxu, line 658: Just to affirm my understanding, a strong quorum of commits in whichever round, leads to decide and termination.

Kubuxu (approved):

gpbft/gpbft.go
masih: ...as long as it's not commit to bottom 👍

PR #627, opened 2024-09-04
Capture broadcasted messages for re-broadcasting even at panic

### LINKED ISSUE #236: Testing misbehaving signer
We should test what happens if we have a "faulty" client (lotus) that fails and/or refuses to sign messages (e.g., miner gets restarted, etc.).

masih:
Fix a bug where if `Host.RequestBroadcast` panics the requested message is never added to the re-broadcast state.

Fix an inconsistent behaviour where error at requested broadcasts via re-broadcast are returned instead of being silently logged.

Enhance emulator to allow pluggable signign logic with two additional signing implementation: erroneous and panic.

Implement tests that assert expected behaviour at signing error or panic.

Fixes #236

### DIFF
--- a/gpbft/gpbft.go
+++ b/gpbft/gpbft.go
@@ -872,11 +872,14 @@ func (i *instance) broadcast(round uint64, step Phase, value ECChain, createTick
 	if createTicket {
 		mb.BeaconForTicket = i.beacon
 	}
+
+	// Capture the broadcast and metrics first. Because, otherwise the instance will
+	// end up with partial re-broadcast messages if RequestBroadcast panics.
+	i.broadcasted.record(mb)
+	metrics.broadcastCounter.Add(context.TODO(), 1, metric.WithAttributes(attrPhase[p.Step]))
 	if err := i.participant.host.RequestBroadcast(mb); err != nil {
 		i.log("failed to request broadcast: %v", err)
 	}
-	i.broadcasted.record(mb)
-	metrics.broadcastCounter.Add(context.TODO(), 1, metric.WithAttributes(attrPhase[p.Step]))
 }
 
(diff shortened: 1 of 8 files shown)


gpbft/gpbft.go
```diff
@@ -872,11 +872,17 @@ func (i *instance) broadcast(round uint64, step Phase, value ECChain, createTick
 	if createTicket {
 		mb.BeaconForTicket = i.beacon
 	}
+
+	defer func() {
```
Stebalien, line 876: No need to defer this, we can just do this first.

gpbft/gpbft.go
```diff
@@ -957,10 +963,13 @@ func (i *instance) rebroadcast() error {
 		mbs := i.broadcasted.messagesByRound[round]
 		for _, mb := range mbs {
 			if err := i.participant.host.RequestBroadcast(mb); err != nil {
-				return err
+				// Silently log the error and proceed. This is consistent with the behaviour of
+				// instance for regular broadcasts.
+				i.log("failed to request rebroadcast %s at round %d: %v", mb.Payload.Step, mb.Payload.Round, err)
```
Stebalien, line 968: Hm... it would be nice to sleep and/or backoff a little. But I guess that's a bit difficult.

Stebalien (approved):
nits

gpbft/gpbft.go
masih: https://github.com/filecoin-project/go-f3/issues/632

masih: We might be able to use a simple circuit breaker at higher level (e.g. `f3`) to cover this.  Updated the issue description.

PR #633, opened 2024-09-05
Rename the term Ticket Quality to Ticket Rank for better readability

masih:
The Converge ticket selection mechanism previously used the term "quality" to determine which ticket is better where the lower the quality value the better the ticket.

Additionally, the term "quality" is already used in GPBFT for `QUALITY` phase, which is a totally different thing.

To avoid terminology confusion, and for a more intuitive notion in the context of ticket selection use the term "rank" instead, where instead one would say: whichever ticket ranks first is better, i.e. the smaller the value of rank the better the ticket.

While at it, update the docs on the compute function.

### DIFF
--- a/gpbft/gpbft.go
+++ b/gpbft/gpbft.go
@@ -1287,13 +1287,12 @@ type convergeState struct {
 type ConvergeValue struct {
 	Chain         ECChain
 	Justification *Justification
-
-	Quality float64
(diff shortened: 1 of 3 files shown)


Stebalien (approved):

Stebalien:
(I don't think this one will be controversial...)

PR #634, opened 2024-09-05
Consistently use `Phase` instead of `Step`

masih:
Use the term `Phase` consistently across the repo to refer to progress phases of GPBFT.

FIP update to follow.


cmd/f3/msgdump/envelope.go
```diff
@@ -39,7 +39,7 @@ type PJustification struct {
 type PPayload struct {
 	Instance         uint64
 	Round            uint64
-	Step             string `parquet:"Step,enum"`
+	Phase            string `parquet:"Phase,enum"`
```
masih, line 42: Note this is a breaking change if we ever want to analyse old parquet logs.

Stebalien (approved):

Stebalien:
LGTM, but I'm going to wait for @Kubuxu to sign off before merging.

cmd/f3/msgdump/envelope.go
Kubuxu: The analysis is done in duckdb, so the queries will have to be updated, but it is fine.

Issue #549, opened 2024-08-08
Handle rebroadcast & pubsub falling behind


Stebalien:
When we get a rebroadcast storm, pubsub can fall behind which can make the issue worse. We should:

- [ ] Ideally introduce some kind of staggering to rebroadcast.
- [ ] Possibly increase pubsub buffer sizes.
- [ ] Possibly rate-limit rebroadcast.
- [ ] Optimize _what_ we rebroadcast.

masih:
> Optimize what we rebroadcast.

We do broadcast more than what the FIP dictates: all messages except quality from the latest 2 rounds, unless we are in DECIDE phase. In which case, only DECIDE. 
What FIP demands depends on phase, and If I have not missed anything it _could_ be 2 messages less. Hence the simpler approach in implementation.

I suspect a more pressing issue is synchronised rebroadcast across nodes.

Stebalien:
> I suspect a more pressing issue is synchronised rebroadcast across nodes.

That will help us avoid this situation, but won't solve it if we do get into this situation.

rjan90:
@Stebalien What are the minimal things we need to implement here, so that in case of a severity where pubsub can falls behind, it does not cause any issues in the network?

Stebalien:
We should:

1. Optimize what we rebroadcast.
2. Stagger rebroadcast (i.e., have all nodes add a random delay to their rebroadcast timeout).

Stebalien:
We've:

1. Staggered rebroadcast.
2. Prioritized rebroadcasting newer messages.

I'm going to call this "done enough".

PR #635, opened 2024-09-05
Randomize rebroadcast backoff

Stebalien:
This fixes part of https://github.com/filecoin-project/go-f3/issues/549 to avoid having the entire network slam pubsub at the same time, although we should still optimize what we rebroadcast.

This also fixes the backoff calculator to return the actual backoff instead of always returning the max (oops).

### DIFF
--- a/gpbft/options.go
+++ b/gpbft/options.go
@@ -144,6 +144,6 @@ func exponentialBackoffer(exponent float64, base, maxBackoff time.Duration) func
 		if nextBackoff > float64(maxBackoff) {
 			return maxBackoff
 		}
-		return maxBackoff
+		return time.Duration(nextBackoff)
 	}
 }
--- a/gpbft/options.go
+++ b/gpbft/options.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"math/rand"
 	"time"
 )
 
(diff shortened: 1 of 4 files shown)


Stebalien:
The remaining part of #549 is to limit our rebroadcast to the last 2 rounds, IIRC?

Stebalien:
Nah, nevermind. We only broadcast from the last 2 rounds. It's not worth it to optimize more.

gpbft/options.go
```diff
@@ -117,33 +118,35 @@ func WithMaxCachedMessagesPerInstance(v int) Option {
...
 // which all messages in the current round are rebroadcast if the round cannot be
 // terminated, i.e. a strong quorum of senders has not yet been achieved.
 //
-// The backoff duration grows exponentially up to the configured max. Defaults to
-// exponent of 1.3 with 3s base backoff growing to a maximum of 30s.
-func WithRebroadcastBackoff(exponent float64, base, max time.Duration) Option {
+// The backoff duration grows exponentially up to the configured max and is randomized by "spread"
+// (a number between 0 and 1). Defaults to exponent of 1.3 with 3s base backoff growing to a maximum
+// of 30s, randomized by 10% (5% in either direction),
+func WithRebroadcastBackoff(exponent, spread float64, base, max time.Duration) Option {
```
masih, line 130: Thinking out loud (and maybe a bit of a language question since I am not a native speaker): would jitter be a better word compared to spread?

gpbft/options.go
```diff
@@ -117,33 +118,35 @@ func WithMaxCachedMessagesPerInstance(v int) Option {
...
+func exponentialBackoffer(exponent, spread float64, base, maxBackoff time.Duration) func(attempt int) time.Duration {
 	return func(attempt int) time.Duration {
-		nextBackoff := float64(base) * math.Pow(exponent, float64(attempt))
+		shift := 1.0 + (rand.Float64()-0.5)*spread
+		nextBackoff := float64(base) * math.Pow(exponent, float64(attempt)) * shift
 		if nextBackoff > float64(maxBackoff) {
 			return maxBackoff
 		}
-		return maxBackoff
+		return time.Duration(nextBackoff)
```
masih, line 150: Thank you for getting this sorted ❤️

gpbft/options.go
```diff
@@ -117,33 +118,35 @@ func WithMaxCachedMessagesPerInstance(v int) Option {
...
 // WithRebroadcastBackoff sets the duration after the gPBFT timeout has elapsed, at
 // which all messages in the current round are rebroadcast if the round cannot be
 // terminated, i.e. a strong quorum of senders has not yet been achieved.
 //
-// The backoff duration grows exponentially up to the configured max. Defaults to
-// exponent of 1.3 with 3s base backoff growing to a maximum of 30s.
-func WithRebroadcastBackoff(exponent float64, base, max time.Duration) Option {
+// The backoff duration grows exponentially up to the configured max and is randomized by "spread"
+// (a number between 0 and 1). Defaults to exponent of 1.3 with 3s base backoff growing to a maximum
+// of 30s, randomized by 10% (5% in either direction),
```
masih, line 129:
```suggestion
// of 30s, randomized by 10% (5% in either direction).
```

masih (approved):

masih:
> We only broadcast from the last 2 rounds

Right, it is already limited to the latest 2 rounds.

gpbft/options.go
Stebalien: Jitter is also correct. It was actually my first thought but I picked spread to make it clear that it's intentional.

But if jitter is also your first thought... it sounds like the word to use.

PR #636, opened 2024-09-05
refactor: simplify rebroadcast logic and prioritize important messages

Stebalien:
This keeps two message lists: one for the current round, one for the previous. With separate fields for quality/decide. That way we don't have to "garbage collect".

This also rebroadcasts newer messages first (except quality, which is always sent first). This _should_ help ensure peers get important messages even if some get dropped in the flood, but I'm not 100% sure?

This fixes the rest of #549, IMO. We may try to rebroadcast a bit more than we need to, but we'll rebroadcast the important messages first.

### DIFF
--- a/gpbft/gpbft.go
+++ b/gpbft/gpbft.go
@@ -952,7 +952,7 @@ func (i *instance) rebroadcast() error {
 	} else {
 		// We rebroadcast messages newest to oldest, except quality which we always
 		// rebroadcast first. That way, if we try to rebroadcast too many at once and some
-		// get dropped, we've already rebroadcast the most useful messages.
+		// get dropped, chances are we've already rebroadcast the most useful messages.
 		msgs = append(msgs, i.broadcasted.previousRound...)
 		msgs = append(msgs, i.broadcasted.currentRound...)
 		if i.broadcasted.quality != nil {
@@ -1393,11 +1393,19 @@ func (c *convergeState) FindProposalFor(chain ECChain) ConvergeValue {
 }
 
 type broadcastState struct {
-	quality       *MessageBuilder
-	decide        *MessageBuilder
+	// Our quality message, always rebroadcast until we reach a decision.
+	quality *MessageBuilder
+	// Our decide message. If set, this is the only message we rebroadcast (and all other fields
+	// should be nil).
+	decide *MessageBuilder
+	// The messages we sent during the previous round (`round - 1`), except quality/decide.
+	// Rebroadcast unless `decide` is non-nil.
 	previousRound []*MessageBuilder
-	currentRound  []*MessageBuilder
-	round         uint64
+	// The messages we sent this round (`round`), except quality/decide. Rebroadcast unless
+	// `decide` is non-nil.
+	currentRound []*MessageBuilder
+	// The latest round in which we broadcast a message.
+	round uint64
 }
 
 // record stores messages that are required should rebroadcast becomes necessary.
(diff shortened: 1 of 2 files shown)


gpbft/gpbft.go
```diff
@@ -940,33 +939,35 @@ func (i *instance) rebroadcastTimeoutElapsed() bool {
...
+	} else {
+		// We rebroadcast messages newest to oldest, except quality which we always
+		// rebroadcast first. That way, if we try to rebroadcast too many at once and some
+		// get dropped, we've already rebroadcast the most useful messages.
+		msgs = append(msgs, i.broadcasted.previousRound...)
+		msgs = append(msgs, i.broadcasted.currentRound...)
+		if i.broadcasted.quality != nil {
+			msgs = append(msgs, i.broadcasted.quality)
+		}
+		slices.Reverse(msgs)
```
Stebalien, line 961: @masih does this prioritization make sense?

gpbft/gpbft.go
```diff
@@ -1418,36 +1417,35 @@ func newBroadcastState() *broadcastState {
...
-				delete(bs.messagesByRound, redundantRound)
-			}
+		// Note that DECIDE message is not associated to any round.
+		bs.currentRound = nil
+		bs.previousRound = nil
+		bs.quality = nil
+		bs.decide = mb
+	case CONVERGE_PHASE, PREPARE_PHASE, COMMIT_PHASE:
+		if mb.Payload.Round < bs.round {
+			log.Warnw("Unexpected broadcast of a message for a prior round")
```
masih, line 1435:
```suggestion
			log.Warnw("Unexpected broadcast of a message for a prior round", "latestRecordedRound", round, "messageRound", mb.Payload.Round)
```

Or use `log.Warn` if you think the suggested log line is an overkill.

gpbft/gpbft.go
```diff
@@ -1418,36 +1417,35 @@ func newBroadcastState() *broadcastState {
...
+		switch mb.Payload.Round {
+		case bs.round:
+			bs.currentRound = append(bs.currentRound, mb)
+		case bs.round + 1:
+			bs.previousRound = bs.currentRound
+			bs.currentRound = []*MessageBuilder{mb}
+		default:
+			bs.previousRound = nil
+			bs.currentRound = []*MessageBuilder{mb}
 		}
```
masih, line 1447: Subjective: I think it would be easier to read if `if` statement is combined with `switch` and `switch` is used without a typed variable even thought it would make the case conditions a bit more verbose.

masih (approved):
Only subjective comments on my part; LGTM 🍻

gpbft/gpbft.go
Stebalien: > It does, though the pedantic in me would say gossipsub makes no sequential delivery guarantees.

It doesn't but... it drops messages once queues start filling up, so earlier messages are more likely to be delivered (just by how it's implemented).

gpbft/gpbft.go
```diff
@@ -940,33 +939,35 @@ func (i *instance) rebroadcastTimeoutElapsed() bool {
...
+	var msgs []*MessageBuilder
+	if i.broadcasted.decide != nil {
+		msgs = append(msgs, i.broadcasted.decide)
+	} else {
+		// We rebroadcast messages newest to oldest, except quality which we always
+		// rebroadcast first. That way, if we try to rebroadcast too many at once and some
+		// get dropped, we've already rebroadcast the most useful messages.
+		msgs = append(msgs, i.broadcasted.previousRound...)
+		msgs = append(msgs, i.broadcasted.currentRound...)
+		if i.broadcasted.quality != nil {
```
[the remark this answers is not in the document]
Stebalien: I'm reversing the order of the messages within rounds as well.

PR #640, opened 2024-09-12
Implement WriteAheadLog

Kubuxu:
Pushing the first commit with complete implementation of WAL for peliminary review.
Integration in progress.

### DIFF
--- a/internal/writeaheadlog/wal.go
+++ b/internal/writeaheadlog/wal.go
@@ -1,6 +1,7 @@
 package writeaheadlog
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"io"
@@ -34,22 +35,17 @@ type WriteAheadLog[T any, PT interface {
 
 	path string
 
-	logFiles []logContent[T, PT]
+	loaded   bool
+	logFiles []logStat
 	active   struct {
-		file       *os.File
-		cborWriter *cbg.CborWriter
-		logContent[T, PT]
+		file *os.File
+		logStat
 	}
 }
 
-type logContent[T any, PT interface {
-	*T
-	Entry
-}] struct {
+type logStat struct {
 	logname  string
-	content  []T
 	maxEpoch uint64
-	complete bool
 }
 
 func Open[T any, PT interface {
@@ -60,7 +56,7 @@ func Open[T any, PT interface {
 		path: directory,
 	}
 
-	err := wal.readInAll()
+	err := wal.hydrate()
 	if err != nil {
 		return nil, fmt.Errorf("reading the WAL: %v", err)
 	}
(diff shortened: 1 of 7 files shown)


Kubuxu:
@masih @Stebalien I would appriciate a light review of the WAL

internal/writeaheadlog/wal.go
```diff
@@ -0,0 +1,295 @@
...
+	Entry
+}] struct {
+	// pointer type trickery above
+	lk sync.Mutex
+
+	path string
+
+	logFiles []logContent[T, PT]
+	active   struct {
+		file       *os.File
```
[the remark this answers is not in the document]
Kubuxu: I don't think it matters here.

internal/writeaheadlog/wal.go
```diff
@@ -0,0 +1,295 @@
...
+
+	if err := wal.maybeRotate(); err != nil {
+		return fmt.Errorf("attemting to rotate: %w", err)
+	}
+
+	err := PT(&value).MarshalCBOR(wal.active.cborWriter)
+	if err != nil {
+		return fmt.Errorf("saving value to WAL: %w", err)
+	}
+	err = wal.active.file.Sync()
```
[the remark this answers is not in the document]
Kubuxu: Yes, it does.

internal/writeaheadlog/wal.go
```diff
@@ -0,0 +1,295 @@
+package writeaheadlog
```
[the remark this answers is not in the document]
Kubuxu: The issue is package name aliasing, and calling it the instance just `wal` is too convenient.

internal/writeaheadlog/wal.go
```diff
@@ -0,0 +1,295 @@
...
+type Entry interface {
+	WALEpoch() uint64
+	cbg.CBORMarshaler
+	cbg.CBORUnmarshaler
+}
+
+type WriteAheadLog[T any, PT interface {
+	*T
+	Entry
+}] struct {
```
Stebalien, line 31: I'm going put my rust police badge aside and ask: does this _need_ to be generic?

internal/writeaheadlog/wal.go
```diff
@@ -0,0 +1,295 @@
...
+	var err error
+	wal.active.file, err = os.OpenFile(
+		filepath.Join(wal.path, wal.active.logname),
+		os.O_CREATE|os.O_WRONLY|os.O_EXCL, 0666)
+
+	if err != nil {
+		wal.active.file = nil
+		return fmt.Errorf("opening new log file %q: %w", wal.active.logname, err)
+	}
+	wal.active.cborWriter = cbg.NewCborWriter(wal.active.file)
```
Stebalien, line 197: Consider buffering.

internal/writeaheadlog/wal.go
```diff
@@ -0,0 +1,295 @@
...
+
+	wal.active.logContent.complete = true
+	wal.logFiles = append(wal.logFiles, wal.active.logContent)
+	wal.active.logContent = logContent[T, PT]{}
+	return nil
+}
+
+func (wal *WriteAheadLog[T, PT]) readInAll() error {
+	_, err := os.Stat(wal.path)
+	if errors.Is(err, os.ErrNotExist) {
```
Stebalien, line 225: nit: just directly call `os.ReadDir`? If the directory does not exist, there's nothing to read in anyways.

internal/writeaheadlog/wal.go
```diff
@@ -0,0 +1,295 @@
...
+// it eats parsing errors and logs in on purpose
+func (wal *WriteAheadLog[T, PT]) readLogFile(logname string) (logContent[T, PT], error) {
+	var res logContent[T, PT]
+	logFile, err := os.Open(filepath.Join(wal.path, logname))
+	if err != nil {
+		return res, fmt.Errorf("opening logfile %q: %w", logname, err)
+	}
+	defer logFile.Close()
+
+	logFileReader := cbg.NewCborReader(logFile)
```
Stebalien, line 266: nit: buffered reader.

internal/writeaheadlog/wal.go
```diff
@@ -0,0 +1,295 @@
...
+
+	res = logContent[T, PT]{
+		logname:  logname,
+		content:  content,
+		maxEpoch: maxEpoch,
+		complete: true,
+	}
+
+	switch {
+	case err == nil:
```
Stebalien, line 286: this case isn't even possible, right?

internal/writeaheadlog/wal.go
Stebalien: I guess it doesn't really matter. I'm just wondering if it could be a bit simpler. If this makes testing simpler, that's fine.

internal/writeaheadlog/wal.go
```diff
@@ -0,0 +1,295 @@
...
+	for _, c := range wal.logFiles {
+		res = append(res, c.content...)
+	}
+	res = append(res, wal.active.content...)
+	return res
+}
+
+/*
+// TODO: version file
+// or mabye not todo, IDK
```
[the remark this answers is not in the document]
Stebalien: Eh, we can do it later. "no version" can mean "v0".

internal/writeaheadlog/wal.go
```diff
@@ -0,0 +1,295 @@
...
+
+const roatateAt = 1 << 20 // 1MiB
+
+func (wal *WriteAheadLog[T, PT]) maybeRotate() error {
+	if wal.active.file == nil {
+		return wal.rotate()
+	}
+	stats, err := wal.active.file.Stat()
+	if err != nil {
+		return fmt.Errorf("collecting stats for the file: %w", err)
```
Stebalien, line 159: IIRC, the "right" way to do this is to seek to the end of the file and look at the return value. We want to be at the end of the file anyways...

Alternatively, we can get our current position by calling `Seek(0, 1)`.

internal/writeaheadlog/wal.go
```diff
@@ -0,0 +1,295 @@
...
+
+	err := wal.readInAll()
+	if err != nil {
+		return nil, fmt.Errorf("reading the WAL: %v", err)
+	}
+
+	return &wal, nil
+}
+
+func (wal *WriteAheadLog[T, PT]) All() []T {
```
Stebalien, line 71: Any real reason to keep all these in memory? IMO, we'll want a more specialized datastructure for keeping track of and indexing sent messages.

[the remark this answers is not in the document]
Stebalien: `All` is convention. But, now that we have iterators, it would be kind of nice to use them here: https://medium.com/eureka-engineering/a-look-at-iterators-in-go-f8e86062937c (especially if we stop storing these in memory).

internal/writeaheadlog/wal.go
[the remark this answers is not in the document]
Kubuxu: This was my thinking

internal/writeaheadlog/wal.go
Kubuxu: Yeah, makes testing simpler, I should also probably revert it and accept pointers.

internal/writeaheadlog/wal.go
Kubuxu: I don't think it is, I expect us to hit EOF or UnexpectedEOF.

internal/writeaheadlog/wal.go
Kubuxu: I agree

internal/writeaheadlog/wal.go
```diff
@@ -0,0 +1,295 @@
...
+		content:  content,
+		maxEpoch: maxEpoch,
+		complete: true,
+	}
+
+	switch {
+	case err == nil:
+		return res, nil
+	case errors.Is(err, io.EOF):
+		return res, nil
```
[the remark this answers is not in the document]
Kubuxu: The err == nil case a bit unexpected so I left a commend

masih (approved):

Issue #603, opened 2024-08-28
Make F3 consensus final


Stebalien:
We need to:

1. Set a checkpoint (in the client) whenever we reach an F3 decision.
2. If we get a chain with an invalid base from lotus, we should wait a bit and try again rather than complaining loudly. I believe this should currently "work", but we'll likely have some noisy log messages.

masih:
> If we get a chain with an invalid base from lotus, we should wait a bit and try again rather than complaining loudly. I believe this should currently "work", but we'll likely have some noisy log messages.

@Stebalien is this not already covered by:
1. [collectChain](https://github.com/filecoin-project/go-f3/blob/3f40360201135dbe92035148e90ee4a1e65d2b89/host.go#L417) returning nil if there is a divergence, which will cause [base with empty suffix](https://github.com/filecoin-project/go-f3/blob/3f40360201135dbe92035148e90ee4a1e65d2b89/host.go#L517) to be proposed as the next instance chain, plus
2. backoff at [computeNextInstanceStart](https://github.com/filecoin-project/go-f3/blob/3f40360201135dbe92035148e90ee4a1e65d2b89/host.go#L245), which will increase on base finalisation

?

Stebalien:
Yeah, it _sounds_ like that part works.

masih:
This is done when the Lotus PR gets merged. Currently in review.

PR #642, opened 2024-09-13
Add finalize API to EC backend

masih:
Expand the F3 EC backend API to mark a given tipset as final. This translates to checkpointing the tipset in Lotus beyond which no forks are allowed.

Part of #603


ec/ec.go
```diff
@@ -14,12 +14,21 @@ type Backend interface {
 	GetTipsetByEpoch(ctx context.Context, epoch int64) (TipSet, error)
 	// GetTipset returns the tipset with the given key.
 	GetTipset(context.Context, gpbft.TipSetKey) (TipSet, error)
-	// GetHead returns the current head tipset of the chain
+	// GetHead returns the current head tipset of the chain, which is guaranteed to
+	// be a descendant of the latest finalized tipset.
+	//
+	// See FinalizeTipset.
```
Stebalien, line 20:
```suggestion
	// See Finalize.
```

Stebalien: I guess we can make this guarantee because _we_ call `Finalize`? Seems ok.

host.go
```diff
@@ -650,6 +649,11 @@ func (h *gpbftHost) ReceiveDecision(decision *gpbft.Justification) (time.Time, e
 		log.Error(err)
 		return time.Time{}, err
 	}
+	if err := h.ec.Finalize(h.runningCtx, cert.ECChain.Head().Key); err != nil {
```
Stebalien, line 652: I assume this can block if we're forced to sync to a chain we don't have? We should leave a comment if that's possible.

Stebalien (approved):

host.go
masih: How long would the sync take in practice? I wonder if it's worth optimising for some non-blocking sync then checkopint, and what the likelihood of the need for sync is to begin with.

Stebalien: We often won't need to sync, but it can take an arbitrary amount of time (e.g., if we're really behind but we get a recent F3 decision).

ec/ec.go
```diff
@@ -14,12 +14,21 @@ type Backend interface {
 	GetTipsetByEpoch(ctx context.Context, epoch int64) (TipSet, error)
 	// GetTipset returns the tipset with the given key.
 	GetTipset(context.Context, gpbft.TipSetKey) (TipSet, error)
-	// GetHead returns the current head tipset of the chain
+	// GetHead returns the current head tipset of the chain, which is guaranteed to
```
Stebalien, line 17:
```suggestion
	// GetHead returns the current head tipset of the chain, which must
```

host.go
```diff
@@ -190,6 +189,19 @@ func (h *gpbftRunner) Start(ctx context.Context) (_err error) {
...
 	nextInstance := c.GPBFTInstance + 1
 	currentInstance := h.participant.CurrentInstance()
+
+	// Call to Finalize checkpoints the tipset in Lotus chain store. This may cause a
+	// sync in a case where the target tipset is not already stored by the chain
+	// store, which is a blocking operation.
+	//
+	// Triggering the checkpointing here means that certstore remains the sole source
+	// of truth in terms of tipset has been finalised.
+	if err := h.ec.Finalize(h.runningCtx, c.ECChain.Head().Key); err != nil {
```
Stebalien, line 199: nit: move the two lines above this below to keep the code together.

Stebalien (approved):

Stebalien (approved):

Issue #643, opened 2024-09-16
Accessing potentially empty slice.


ruseinov:
Should there be `is_empty` check? e.g. `if c.isZero() { return nil } `? Otherwise this will panic:
https://github.com/filecoin-project/go-f3/blob/2c6f7bf80f19398f97318599ae90c97e5291c47e/gpbft/chain.go#L143

The same goes for https://github.com/filecoin-project/go-f3/blob/bc35325b79c328949d4a7c22e8133a7d1cc5ee64/gpbft/chain.go#L176 and perhaps some other occurrences.

Stebalien:
I guess we could return `nil` in these cases, but that would just move the panic elsewhere. Better documentation is likely in-order, these functions should only be called on non-zero chains.

ruseinov:
> I guess we could return `nil` in these cases, but that would just move the panic elsewhere. Better documentation is likely in-order, these functions should only be called on non-zero chains.

Right, I'm combing through a lot of this code now as I'm working on the Rust implementation and trying to organize the API so that it's panic-free, propagating errors all the way to wherever they need to be handled or finally turned into a panic. I can appreciate that it might not be as ergonomic to do here.

PR #645, opened 2024-09-17
Remove panic cases from TipSet methods and simplify nil handling

Stebalien:
Panicing is perfectly fine in these cases, but not necessary and a little inconsistent (e.g., `Suffix` didn't panic).

1. Avoid panicing when TipSet is nil in Validate.
2. Use nil tipsets as zero tipsets, no need to look at the key. Empty tipset keys will fail `Validate` anyways. Remove `TipSet.Zero`.
3. Return a nil base when the chain has no base instead of panicing.
4. Return a nil head when the chain is empty.
5. Return a nil base chain when the chain has no base.
6. Improve some documentation.

fixes https://github.com/filecoin-project/go-f3/issues/643

### DIFF
--- a/gpbft/chain.go
+++ b/gpbft/chain.go
@@ -59,6 +59,9 @@ type TipSet struct {
 // Validates a tipset.
 // Note the zero value is invalid.
 func (ts *TipSet) Validate() error {
+	if ts == nil {
+		return errors.New("tipset must not be nil")
+	}
 	if len(ts.Key) == 0 {
 		return errors.New("tipset key must not be empty")
 	}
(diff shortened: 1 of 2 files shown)


Stebalien:
cc @ruseinov who brought this up.

ruseinov:
Looks good, thank you!

gpbft/chain.go
```diff
@@ -74,11 +77,10 @@ func (ts *TipSet) Validate() error {
 	return nil
 }
 
-func (ts *TipSet) IsZero() bool {
-	return len(ts.Key) == 0
-}
-
 func (ts *TipSet) Equal(b *TipSet) bool {
+	if ts == nil || b == nil {
```
Kubuxu, line 81: We can simplify this to `ts == b`, can't we?

Kubuxu (approved):

PR #649, opened 2024-09-20
Drop messages that are no longer useful for GPBFT progression

### LINKED ISSUE #583: Drop all messages we wouldn't rebroadcast (pubsub related)
This will help us quiet down pubsub between instances and should help us save some bandwidth. This also means we can punt fixing #576 because the only messages that "loop" are messages we want to rebroadcast anyways.

masih:
As a GPBFT instance progresses some messages become irrelevant, in that they do not effectively aid the progress of the instance for participants. Instead, GPBFT offers other built-in mechanisms to aid progress of lagging participants such as selective rebroadcast, propagation of DECIDE messages from the previous instance and finality certificate exchange.

The changes here introduce a dedicated error type returned as part of message validation to signal that although a message is valid it is no longer relevant. This error type is then checked by pubsub to avoid further propagation of those messages.

This reduces the redundant pubsub traffic for the network participants.

Fixes #583

### DIFF
--- a/gpbft/validator.go
+++ b/gpbft/validator.go
@@ -77,15 +77,36 @@ func (v *cachingValidator) ValidateMessage(msg *GMessage) (valid ValidatedMessag
 	}
 
 	// Infer whether to proceed validating the message relative to the current instance.
-	switch currentInstance := v.progress.Load().id; {
-	case msg.Vote.Instance >= currentInstance+v.committeeLookback:
+	switch current := v.progress.Load(); {
+	case msg.Vote.Instance >= current.id+v.committeeLookback:
 		// Message is beyond current + committee lookback.
 		return nil, ErrValidationNoCommittee
-	case msg.Vote.Instance >= currentInstance,
-		msg.Vote.Instance == currentInstance-1 && msg.Vote.Phase == DECIDE_PHASE:
+	case msg.Vote.Instance > current.id,
+		msg.Vote.Instance+1 == current.id && msg.Vote.Phase == DECIDE_PHASE:
 		// Only proceed to validate the message if it:
 		//  * belongs to an instance within the range of current to current + committee lookback, or
 		//  * is a DECIDE message belonging to previous instance.
+	case msg.Vote.Instance == current.id:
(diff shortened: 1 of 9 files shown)


gpbft/validator.go
```diff
@@ -77,15 +77,33 @@ func (v *cachingValidator) ValidateMessage(msg *GMessage) (valid ValidatedMessag
...
+		//   * When current instance is at DECIDE phase only validate DECIDE messages.
+		//   * Otherwise, only validate messages that would be rebroadcasted, i.e. QUALITY,
+		//     DECIDE, messages from previous round, and messages from current round.
+		// Anything else is not relevant.
+		switch {
+		case current.phase == DECIDE_PHASE && msg.Vote.Phase != DECIDE_PHASE:
+			return nil, ErrValidationNotRelevant
+		case msg.Vote.Phase == QUALITY_PHASE,
+			msg.Vote.Phase == DECIDE_PHASE,
+			msg.Vote.Round == current.round-1, // Use explicit case for previous round to avoid unt64 overflow.
```
Stebalien, line 101: But this will wrap?

gpbft/validator.go
```diff
@@ -77,15 +77,33 @@ func (v *cachingValidator) ValidateMessage(msg *GMessage) (valid ValidatedMessag
...
+		// i.e.:
+		//   * When current instance is at DECIDE phase only validate DECIDE messages.
+		//   * Otherwise, only validate messages that would be rebroadcasted, i.e. QUALITY,
+		//     DECIDE, messages from previous round, and messages from current round.
+		// Anything else is not relevant.
+		switch {
+		case current.phase == DECIDE_PHASE && msg.Vote.Phase != DECIDE_PHASE:
+			return nil, ErrValidationNotRelevant
+		case msg.Vote.Phase == QUALITY_PHASE,
+			msg.Vote.Phase == DECIDE_PHASE,
```
Stebalien, line 100: vote == quality and decide? I assume one of these is supposed to be our phase, right?

Stebalien: nvm.

Stebalien (approved):

Issue #561, opened 2024-08-10
Refactor validation logic for a more pluggable functionality


masih:
The current validation logic is encapsulated at a package-level function that takes a number of arguments to validate a given message. This is useful for users whom are just interested in validating a given gpbft message but does require `Host` to be passed in. `Host` is a heavy struct; but in validation it is only used to marshal payload.

Separately, it is desirable to cache the computationally expensive parts of validation specially if the rate of duplicates in the live network is high. The two computationally expensive parts are: message signature validation and justification aggregated signature validation. The work in #557 and #560 introduces caching for validation purposes. But the API can be refined further to make future extensions easier while making modifications less error-prone since message validation is highly critical.

Consider:
* separating validator into its own struct that is then plugged into `Participant`
* ideally making `Participant` oblivious of caching in validator in favour of a swappable validation cache mechanism.

masih:
Also see: https://github.com/filecoin-project/go-f3/pull/560#discussion_r1714173707

PR #651, opened 2024-09-22
Refactor `Chain` interface into separate do-er interfaces

masih:
Refactor the `Chain` interface into two separate smaller interfaces for a more pluggable committee caching.

Refactor committee cache out of participant.

This work is part of a larger refactor of validation mechanism and is separated into its own commit for easier review.

Part of #561

### DIFF
--- a/gpbft/committee.go
+++ b/gpbft/committee.go
@@ -1,0 +1,52 @@
+package gpbft
+
+import (
+	"fmt"
+	"sync"
+)
+
+var _ CommitteeProvider = (*cachedCommitteeProvider)(nil)
+
+type cachedCommitteeProvider struct {
+	delegate CommitteeProvider
+
+	// mu guards access to committees.
+	mu         sync.Mutex
+	committees map[uint64]*Committee
+}
+
+func newCachedCommitteeProvider(delegate CommitteeProvider) *cachedCommitteeProvider {
+	return &cachedCommitteeProvider{
+		delegate:   delegate,
+		committees: make(map[uint64]*Committee),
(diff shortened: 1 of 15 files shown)


gpbft/committee.go
```diff
@@ -0,0 +1,56 @@
...
+		delegate:   delegate,
+		committees: make(map[uint64]*Committee),
+	}
+}
+
+func (c *cachedCommitteeProvider) GetCommittee(instance uint64) (*Committee, error) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	committee, found := c.committees[instance]
+	if !found {
```
Stebalien, line 29: nit: invert this and return early if found.

gpbft/committee.go
```diff
@@ -0,0 +1,56 @@
...
+		var err error
+		switch committee, err = c.delegate.GetCommittee(instance); {
+		case err != nil:
+			return nil, fmt.Errorf("instance %d: %w: %w", instance, ErrValidationNoCommittee, err)
+		case committee == nil:
+			return nil, fmt.Errorf("unexpected nil committee for instance %d", instance)
+		default:
+			if err := committee.PowerTable.Validate(); err != nil {
+				return nil, fmt.Errorf("instance %d: %w: invalid power: %w", instance, ErrValidationNoCommittee, err)
+			}
```
Stebalien, line 39: This is fine but... it's a bit odd that we're calling this at the caching layer. I'd move this to the host.

Stebalien (approved):

PR #662, opened 2024-09-24
Load the initial power table from the network even if we have the tipset

Stebalien:
We generally have all tipsets, just not all states. We need to load the power table from the network in case either call fails.


Kubuxu (approved):

f3_test.go
```diff
@@ -31,9 +31,10 @@ func init() {
 	// Hash-based deduplication breaks fast rebroadcast, even if we set the time window to be
 	// really short because gossipsub has a minimum 1m cache scan interval.
 	psutil.GPBFTMessageIdFn = pubsub.DefaultMsgIdFn
+	psutil.ManifestMessageIdFn = pubsub.DefaultMsgIdFn
 }
 
-var ManifestSenderTimeout = 30 * time.Second
+var manifestSenderTimeout = 10 * time.Second
```
Stebalien, line 37: This ensures we don't "miss" the manifest update and flake the test.

f3_test.go
```diff
@@ -108,13 +109,10 @@ func TestF3PauseResumeCatchup(t *testing.T) {
...
 	// Resuming node 1 should continue agreeing on instances.
 	env.resumeNode(1)
-	// make sure we're ahead of node 1.
-	resumeInstance := env.nodes[1].currentGpbftInstance() + 1
-	env.waitForInstanceNumber(resumeInstance, 30*time.Second, false)
 
 	// Wait until we're far enough that pure GPBFT catchup should be impossible.
-	targetInstance := resumeInstance + env.manifest.CommitteeLookback + 1
-	env.waitForInstanceNumber(targetInstance, 30*time.Second, false)
+	targetInstance := env.nodes[1].currentGpbftInstance() + env.manifest.CommitteeLookback + 1
```
Stebalien, line 114: Simplification.

f3_test.go
```diff
@@ -108,13 +109,10 @@ func TestF3PauseResumeCatchup(t *testing.T) {
...
+	env.waitForInstanceNumber(targetInstance, 60*time.Second, false)
```
Stebalien, line 115: This usually doesn't take 60s, but I'm trying to make these less flaky.

f3_test.go
```diff
@@ -484,10 +547,10 @@ func (e *testEnv) waitFor(f func(n *testNode) bool, timeout time.Duration) {
 }
 
 func (e *testEnv) waitForNodesInitialization() {
-	f := func(n *testNode) bool {
+	e.t.Helper()
+	e.waitFor(func(n *testNode) bool {
 		return n.f3.IsRunning()
-	}
-	e.waitFor(f, 5*time.Second)
+	}, 30*time.Second)
```
Stebalien, line 553: This probably isn't necessary anymore (I was trying to fix an issue where the race-detector slowed things down too much) but... 5s was unnecessarily short anyways.

Stebalien:
Finally! It passes!

masih (approved):

masih:
> Finally! It passes!

`TestF3LateBootstrap` strikes again 😭

Issue #599, opened 2024-08-28
Change Lotus Miner Participation API to avoid equivocations


Kubuxu:
The current API does not prevent equivocations if requests are mirrored

Kubuxu:
Discussed: change the paticipation API to be per-instance, and both lotus and miner code should remember that leases were givinen out.
We need to give out future instance as well.

PR #663, opened 2024-09-26
Simplify participant locking and expose current GPBFT progress

### LINKED ISSUE #658: Simplify mutex protected current instance in host and participant using progress observer
Review and simplify the mutex protected reference to current instance in `Participant` and `Host` implementations by reusing the progress update mechanism introduced as part of #657. 

Ideally, we should be able to get rid of `instanceMutex` entirely, and as a bonus  provide a progress API to Lotus which we would probably need in the future.

[Original context](https://github.com/filecoin-project/go-f3/pull/657#discussion_r1771370895).

masih:
The host implementation needs to know the progress of the current instance for certificate storage and catch up purposes. As a result the current instance must be safe to read from multiple goroutines. This was achieved by a dedicated mutex that synchronised all access to the current instanceID.

Separately, validation logic requires to know how far the current instance has progressed in order to effectively validate incoming messages.

The changes here simplify the locking by unifying the logic for checking the progress of an instance: both validator and host can now get the latest progress from the participant that is safe for concurrent use. This is achieved by moving the progress observer mechanism out of validator and into the participant, while introducing the concept of `gpbft.Progress` a function that returns the current instance, round and phase.

As part of this change, Lotus integration can now get the current round and phase as well as current instance, which is useful for self equivocation checking as well as debugging purposes.

Fixes #658
Relates to #599

### DIFF
--- a/gpbft/participant.go
+++ b/gpbft/participant.go
@@ -9,6 +9,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/filecoin-project/go-f3/internal/caching"
 	logging "github.com/ipfs/go-log/v2"
 	"go.opentelemetry.io/otel/metric"
 )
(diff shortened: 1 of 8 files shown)


gpbft/participant.go
```diff
@@ -48,7 +42,19 @@ type Participant struct {
...
+	// instance exists.
+	progression *atomicProgression
+	// messageCache maintains a cache of unique identifiers for valid messages or
+	// justifications that this Participant has received since the previous instance.
+	// It ensures that only relevant messages or justifications are retained by
+	// automatically pruning entries from older instances as the participant
+	// progresses to the next instance.
+	//
+	// See Participant.finishCurrentInstance, Participant.validator.
+	messageCache *caching.GroupedSet
```
masih, line 56: I ended up pulling this back into participant, so that on finish instance we can call it to evict older messages.

I think having this in Participant might come handy for implementing optimisations or equivocation protections that require cross-round or cross-instance state of received messages.

Stebalien (approved):

PR #667, opened 2024-09-27
Reduce GOGC in fuzz tests

### LINKED ISSUE #666: Investigate sporadic fuzz failures with exit code 143
The exit code 143 signals that the process was killed by `SIGTERM`. This happens when the process oversteps its maximum granted resources, most commonly memory.

We are observing this kind of failure more now. We can either reduce GO_GC  to have fuzz running a bit slower or use self-hosted runners with higher resource limit per CI job.

masih:
Investigate false negative fuzz test failures due to hitting memory limit by reverting GOGC back to default 100.

Fixes #666

### DIFF
--- a/.github/workflows/go-fuzz.yml
+++ b/.github/workflows/go-fuzz.yml
@@ -21,6 +21,7 @@ jobs:
       - name: Fuzz
         env:
           FUZZTIME: 30s
+          GOGC: '100'
         run: make fuzz
       - name: Upload fuzz failure seed corpus as run artifact
         if: failure()


masih:
Repeated fuzzing [3 times](https://github.com/filecoin-project/go-f3/actions/runs/11067775266/job/30752747703) with the changes here and the time it take for the fuzzing to run is more or less the same as the time it takes on `main`. 

All three attempts succeeded.

Stebalien (approved):

PR #670, opened 2024-09-27
Refactor F3 tests

Stebalien:
This refactors the F3 tests to make them work a bit more reliably. Specifically:

1. It ensures that the initial manfest given to the F3 nodes and the initial manifest given to the dynamic manifest server are the same to avoid a spurious update.
2. It fixes the network names.
3. It switches to a builder pattern (ish) for defining test setups. This will help me  in a test I'm currently writing.

### DIFF
--- a/f3_test.go
+++ b/f3_test.go
@@ -37,21 +37,16 @@ var ManifestSenderTimeout = 30 * time.Second
 
 func TestF3Simple(t *testing.T) {
 	t.Parallel()
-	env := newTestEnvironment(t, 2, false)
-
-	env.connectAll()
-	env.start()
+	env := newTestEnvironment(t).withNodes(2).start()
 	env.waitForInstanceNumber(5, 10*time.Second, false)
 }
 
 func TestF3WithLookback(t *testing.T) {
 	t.Parallel()
-	env := newTestEnvironment(t, 2, true)
-	env.manifest.EC.HeadLookback = 20
-	env.updateManifest()
+	env := newTestEnvironment(t).withNodes(2).withManifest(func(m *manifest.Manifest) {
+		m.EC.HeadLookback = 20
+	}).start()
 
-	env.connectAll()
-	env.start()
 	env.waitForInstanceNumber(5, 10*time.Second, false)
 
 	// Wait a second to let everything settle.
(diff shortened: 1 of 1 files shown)


f3_test.go
```diff
@@ -37,21 +37,16 @@ var ManifestSenderTimeout = 30 * time.Second
...
 	env.waitForInstanceNumber(5, 10*time.Second, false)
 }
 
 func TestF3WithLookback(t *testing.T) {
 	t.Parallel()
-	env := newTestEnvironment(t, 2, true)
-	env.manifest.EC.HeadLookback = 20
-	env.updateManifest()
+	env := newTestEnvironment(t).withNodes(2).withManifest(func(m *manifest.Manifest) {
+		m.EC.HeadLookback = 20
```
Stebalien, line 47: It's now possible to mutate the manifest before constructing the f3 nodes, avoiding the need for a dynamic manifest.

f3_test.go
```diff
@@ -74,12 +69,12 @@ func TestF3WithLookback(t *testing.T) {
...
 
-	// Now make sure we've advanced by significantly less than 100 instances.
-	// We want to make sure we're not racing.
+	// Now make sure we've advanced by less than 100 instances. We want to make sure we're not
+	// racing.
 	cert, err = env.nodes[0].f3.GetLatestCert(env.testCtx)
 	require.NoError(t, err)
 	require.NotNil(t, cert)
-	require.Less(t, cert.GPBFTInstance, uint64(80))
+	require.Less(t, cert.GPBFTInstance, uint64(90))
```
Stebalien, line 77: This was constantly flaky.

f3_test.go
```diff
@@ -385,87 +422,59 @@ func (e *testEnv) advance() {
...
-				return false
-			}
-		}
-		return true
-	}, timeout)
+	return e
 }
 
-func newTestEnvironment(t *testing.T, n int, dynamicManifest bool) *testEnv {
+func newTestEnvironment(t *testing.T) *testEnv {
```
Stebalien, line 433: This no longer constructs any nodes.

f3_test.go
```diff
@@ -385,87 +422,59 @@ func (e *testEnv) advance() {
...
-	for i := 0; i < n; i++ {
-		env.initNode(i, manifestServer)
-	}
-
 	return env
 }
 
-func (e *testEnv) initNode(i int, manifestServer peer.ID) {
-	n, err := e.newF3Instance(i, manifestServer)
+func (e *testEnv) addNode() *testNode {
```
Stebalien, line 465: This creates a test node but doesn't actually construct the F3 instance. That happens when `init` and/or `start` is called, giving us a chance to mess with the manifest/ec/etc.

f3_test.go
```diff
@@ -496,10 +505,49 @@ func (e *testEnv) waitForNodesStopped(timeout time.Duration) {
 	e.waitFor(f, timeout)
 }
 
-func (e *testEnv) start() {
+// Connect & link all nodes. This operation is idempotent.
+func (e *testEnv) connectAll() *testEnv {
```
Stebalien, line 509: Equivalent to the old code, just simpler.

masih (approved):

PR #673, opened 2024-09-30
Expose access to progress on f3 runner

masih:
The root F3 instance holds reference to the `gpbftRunner`, not `gpbftHost`. To allow access to current progress from Lotus, move the `Progress` implementation to runner and expose at root `F3` type.

Part of #599


Stebalien (approved):
LGTM but needs a rebase.

PR #677, opened 2024-10-01
Avoid holding the F3 lock while starting it

Stebalien:
This was causing some testing issues because, e.g., calls to `IsRunning` would block while the startup logic was fetching the initial power table.

### DIFF
--- a/f3_test.go
+++ b/f3_test.go
@@ -108,17 +108,13 @@ func TestF3PauseResumeCatchup(t *testing.T) {
 
 	// Resuming node 1 should continue agreeing on instances.
 	env.resumeNode(1)
-	// make sure we're ahead of node 2, which may be ahead of node 0.
-	resumeInstance := env.nodes[2].currentGpbftInstance() + 1
+	// make sure we're ahead of node 1.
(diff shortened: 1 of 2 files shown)


f3_test.go
```diff
@@ -108,26 +108,22 @@ func TestF3PauseResumeCatchup(t *testing.T) {
 
 	// Resuming node 1 should continue agreeing on instances.
 	env.resumeNode(1)
-	// make sure we're ahead of node 2, which may be ahead of node 0.
-	resumeInstance := env.nodes[2].currentGpbftInstance() + 1
+	// make sure we're ahead of node 1.
+	resumeInstance := env.nodes[1].currentGpbftInstance() + 1
 	env.waitForInstanceNumber(resumeInstance, 30*time.Second, false)
```
Stebalien, line 113: We no longer have access to the cert-store while paused but... there wasn't much point in that anyways. This test is now a little less accurate (check then pause, instead of pause then check), but it should be good enough and shouldn't be flaky.

Stebalien:
I'm not a huge fan of adding another layer, but we were starting to have a _lot_ of random adjacent services and this lets us atomically swap them out.

masih (approved):

PR #678, opened 2024-10-01
Bump the protocol version compatibility

Stebalien:
I can't recall if we've made any major breaking changes, but I want to bump this one more time before the final release just to be safe.


Stebalien:
Don't merge this till we're ready for a final release.

masih (approved):

PR #680, opened 2024-10-02
Fix receive cert error handling

masih:
The runner seemed to swallow errors in `receiveCertificate`. Return them and log at caller.

There are already a bunch of TODOs to determine if fatal errors should be separated. Hence, the logging only.


Stebalien:
Ok, I'm confused. I _changed_ this. `StartInstanceAt` shouldn't return an error.

Stebalien:
Oh! I never actually removed the error return and can't because we catch panics.

Stebalien:
I pushed another commit that reverts the rest of #674. I had thought I could remove errors from `StartInstanceAt`, and I'm not entirely sure how that patch ended up half-finished.

PR #687, opened 2024-10-03
Implement new "fusing" and "no" manifest providers

Stebalien:
1. `FusingManifestProvider` switches from dynamic manifests to static manifests when we get within one finality of the passed static manifest.
2. `NoManifestProvider` is a no-op provider that never returns a manifest. Useful for disabling F3 while still constructing it (without providing some kind of no-op paused manifest).

Also, improve error handling and manifest validation. It may be a bit over the top, but I really want to make sure we fail early if we have an issue with the manifest.


manifest/fusing_manifest_provider.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/filecoin-project/go-f3/ec"
+	"go.uber.org/multierr"
+	"golang.org/x/sync/errgroup"
+)
+
+// HeadGetter is the minimal subset of github.com/filecoin-project/go-f3/ec.Backend required by the
+// MixedManifestProvider.
+type HeadGetter interface {
+	GetHead(context.Context) (ec.TipSet, error)
+}
```
Stebalien, line 17: I tried to find a way to avoid this, but couldn't. I would need to have sent some kind of "ranked list" of possible manifests, letting the F3 module decide between them based on the current epoch, or something like that. It was looking pretty nasty.

manifest/fusing_manifest_provider.go
```diff
@@ -0,0 +1,109 @@
...
+				m.updateManifest(update)
+			}
+		}
+	})
+
+	return nil
+}
+
+func (m *FusingManifestProvider) updateManifest(update *Manifest) {
+	drain(m.manifestCh)
```
Stebalien, line 102: I'm _pretty_ sure this mechanism is safe?

Stebalien:
Tests still in progress, but I wanted to get this up for review.

manifest/fusing_manifest_provider.go
```diff
@@ -0,0 +1,118 @@
...
+type HeadGetter interface {
+	GetHead(context.Context) (ec.TipSet, error)
+}
+
+var _ ManifestProvider = (*FusingManifestProvider)(nil)
+
+// FusingManifestProvider is a ManifestProvider that starts by providing dynamic manifest updates
+// then switches to a static manifest when we get within finality of said manifest's bootstrap
+// epoch.
+type FusingManifestProvider struct {
```
masih, line 25: Tests for this please? Either here or in a follow up PR that uses it. Up to you.

manifest/fusing_manifest_provider.go
```diff
@@ -0,0 +1,118 @@
...
+		clock:      clk,
+		manifestCh: make(chan *Manifest, 1),
+	}, nil
+}
+
+func (m *FusingManifestProvider) ManifestUpdates() <-chan *Manifest {
+	return m.manifestCh
+}
+
+func (m *FusingManifestProvider) Start(ctx context.Context) error {
```
masih, line 63: I would add a bunch of logs here. I think we want to know what's happening in this function for debugging purposes, specifically around switch over epoch, when it actually happens and so on.

manifest/fusing_manifest_provider.go
masih: See comment on `drain` implementation.

manifest/no_manifest_provider.go
```diff
@@ -0,0 +1,13 @@
...
+	"context"
+)
+
+var _ ManifestProvider = (*NoManifestProvider)(nil)
+
+type NoManifestProvider struct{}
+
+func (m NoManifestProvider) Start(context.Context) error       { return nil }
+func (m NoManifestProvider) Stop(context.Context) error        { return nil }
+func (m NoManifestProvider) ManifestUpdates() <-chan *Manifest { return nil }
```
masih, line 13:
```suggestion
func (NoManifestProvider) Start(context.Context) error       { return nil }
func (NoManifestProvider) Stop(context.Context) error        { return nil }
func (NoManifestProvider) ManifestUpdates() <-chan *Manifest { return nil }
```

manifest/no_manifest_provider.go
```diff
@@ -0,0 +1,13 @@
+package manifest
+
+import (
+	"context"
+)
+
+var _ ManifestProvider = (*NoManifestProvider)(nil)
+
+type NoManifestProvider struct{}
```
masih, line 9: Subjective: 
* `Noop` instad of No.
* While at it rename the files containing provider implementations in this package to either have `_mainfest_provider.go` suffix, or not.  I would probably use `_provider.go`.

masih (approved):

manifest/fusing_manifest_provider.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func (m *FusingManifestProvider) updateManifest(update *Manifest) {
+	drain(m.manifestCh)
+	m.manifestCh <- update
+}
+
+func (m *FusingManifestProvider) Stop(ctx context.Context) error {
+	m.cancel()
+	return m.errgrp.Wait()
```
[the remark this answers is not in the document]
Stebalien: Yeah, this is an issue everywhere.

manifest/util.go
```diff
@@ -0,0 +1,11 @@
+package manifest
+
+func drain[C any](ch <-chan C) {
+	for {
+		select {
+		case <-ch:
+		default:
+			return
+		}
+	}
```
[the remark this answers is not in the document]
Stebalien: The channel isn't closed, I just want to drain until it's empty so I can write a new value. Technically the loop itself is necessary, but it makes it work no matter the size of the channel (assuming `drain` is never called concurrently with writing to it).

manifest/fusing_manifest_provider.go
Stebalien: Yep, that's next.

manifest/no_manifest_provider.go
Stebalien: Yeah, I like the suggestion.

Kubuxu (approved):
SGWM bar tests

manifest/fusing_manifest_provider.go
Stebalien: I added a couple of logs.

Stebalien:
1. Added a few logs.
2. Added tests.
3. Avoided aborting everything if stopping the dynamic manifest provider fails.

manifest/util.go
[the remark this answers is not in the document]
masih: ah brain fart on my part re channel closure 👍 
Does select statement in go guarantee that channel is read from first if `default` is present?

manifest/static.go
```diff
@@ -12,10 +12,13 @@ type StaticManifestProvider struct {
 	ch <-chan *Manifest
 }
 
-func NewStaticManifestProvider(m *Manifest) ManifestProvider {
+func NewStaticManifestProvider(m *Manifest) (*StaticManifestProvider, error) {
```
masih, line 15: Rename this file to `static_provider.go` while at it?

masih (approved):

PR #688, opened 2024-10-03
Specify whether or not to finalize tipsets in the manifest itself

Stebalien:
This'll make it easier to wire up the bootstrap logic.

### DIFF
--- a/host.go
+++ b/host.go
@@ -216,8 +216,14 @@ func (h *gpbftRunner) Start(ctx context.Context) (_err error) {
 						// There is not much we can do here other than logging. The next instance start
 						// will effectively retry checkpointing the latest finalized tipset. This error
 						// will not impact the selection of next instance chain.
-						log.Error(fmt.Errorf("error while finalizing decision at EC: %w", err))
+						log.Errorf("error while finalizing decision at EC: %+v", err)
 					}
+				} else {
+					ts := cert.ECChain.Head()
+					log.Infow("not finalizing a new head because Finalize the manifest specifies that tipsets should not be finalized",
+						"tsk", ts.Key,
+						"epoch", ts.Epoch,
+					)
 				}
 				const keepInstancesInWAL = 5
 				if cert.GPBFTInstance > keepInstancesInWAL {
--- a/host.go
+++ b/host.go
@@ -210,12 +210,14 @@ func (h *gpbftRunner) Start(ctx context.Context) (_err error) {
 					// does, error loudly since the chances are the cause is a programmer error.
 					return errors.New("cert store subscription to finalize tipsets was closed unexpectedly")
 				}
-				key := cert.ECChain.Head().Key
-				if err := h.ec.Finalize(h.runningCtx, key); err != nil {
-					// There is not much we can do here other than logging. The next instance start
-					// will effectively retry checkpointing the latest finalized tipset. This error
-					// will not impact the selection of next instance chain.
-					log.Error(fmt.Errorf("error while finalizing decision at EC: %w", err))
+				if h.manifest.EC.Finalize {
+					key := cert.ECChain.Head().Key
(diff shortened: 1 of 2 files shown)


Stebalien:
We should probably _also_ add an option to the dynamic manifest provider allowing us to filter/reject manifests we don't want. Specifically, mainnet nodes should:

1. Reject dynamic manifests where the network name is equal to the "static" network name (the final network name).
2. Reject dynamic manifests that finalize tipsets (except when testing).

manifest/manifest.go
```diff
@@ -27,6 +27,7 @@ var (
 		// MaxBackoff is 15min given default params
 		BaseDecisionBackoffTable: []float64{1.3, 1.69, 2.2, 2.86, 3.71, 4.83, 6.27, 7.5},
 		HeadLookback:             0,
+		Finalize:                 true,
```
masih, line 30: Would be nice to make the default be zero-value of `bool`

host.go
```diff
@@ -210,12 +210,14 @@ func (h *gpbftRunner) Start(ctx context.Context) (_err error) {
...
-					// will effectively retry checkpointing the latest finalized tipset. This error
-					// will not impact the selection of next instance chain.
-					log.Error(fmt.Errorf("error while finalizing decision at EC: %w", err))
+				if h.manifest.EC.Finalize {
+					key := cert.ECChain.Head().Key
+					if err := h.ec.Finalize(h.runningCtx, key); err != nil {
+						// There is not much we can do here other than logging. The next instance start
+						// will effectively retry checkpointing the latest finalized tipset. This error
+						// will not impact the selection of next instance chain.
+						log.Error(fmt.Errorf("error while finalizing decision at EC: %w", err))
```
masih, line 219: I fat-fingered this line, apologies. Below should fix it.
```suggestion
						log.Errorf("error while finalizing decision at EC: %w", err)
```

host.go
```diff
@@ -210,12 +210,14 @@ func (h *gpbftRunner) Start(ctx context.Context) (_err error) {
 					// does, error loudly since the chances are the cause is a programmer error.
 					return errors.New("cert store subscription to finalize tipsets was closed unexpectedly")
 				}
-				key := cert.ECChain.Head().Key
-				if err := h.ec.Finalize(h.runningCtx, key); err != nil {
-					// There is not much we can do here other than logging. The next instance start
-					// will effectively retry checkpointing the latest finalized tipset. This error
-					// will not impact the selection of next instance chain.
-					log.Error(fmt.Errorf("error while finalizing decision at EC: %w", err))
+				if h.manifest.EC.Finalize {
```
masih, line 213: For the benefit of user, do we want to log somewhere at start to indicate tipsets will not be marked as finalized even if Lotus's `F3Consensus` build param is true?

I would, because I suspect this might surprise Lotus operators.

masih (approved):

masih:
> We should probably _also_ add an option to the dynamic manifest provider allowing us to filter/reject manifests we don't want.

Would manifest protocol version alone not be enough? Why we want a dynamic filter over manifests within the same version?

Stebalien:
> Would manifest protocol version alone not be enough? Why we want a dynamic filter over manifests within the same version?

Security. To make sure the dynamic manifest provider can't make nodes do things.

manifest/manifest.go
Stebalien: I generally agree but... I think this way is actually safer in this case. But I could be swayed either way.

host.go
Stebalien: Whether or not to finalize will be hard-coded, but I do agree that an info log is in order.

host.go
Stebalien: Needs to be `%+v`, committing locally.

PR #689, opened 2024-10-03
Support filtering received dynamic manifest

Stebalien:
That way we can, e.g., prevent the manifest server from sending manifests that apply to the mainnet network name, actually finalize tipsets, etc.

See #688.

### DIFF
--- a/cmd/f3/observer.go
+++ b/cmd/f3/observer.go
@@ -166,7 +166,10 @@ var observerCmd = cli.Command{
 		if err != nil {
 			return fmt.Errorf("decoding manifestServerID: %w", err)
 		}
-		manifestProvider, err := manifest.NewDynamicManifestProvider(nil, ds, pubSub, manifestServerID)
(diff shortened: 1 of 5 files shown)


cmd/f3/observer.go
```diff
@@ -166,7 +166,10 @@ var observerCmd = cli.Command{
 		if err != nil {
 			return fmt.Errorf("decoding manifestServerID: %w", err)
 		}
-		manifestProvider, err := manifest.NewDynamicManifestProvider(nil, ds, pubSub, manifestServerID)
+		manifestProvider, err := manifest.NewDynamicManifestProvider(
+			pubSub, manifestServerID,
+			manifest.DynamicManifestProviderWithDatastore(ds),
```
Stebalien, line 171: Gets a bit wordy, but makes it easier to add more params later.

masih (approved):

Kubuxu (approved):

PR #691, opened 2024-10-03
Rebroadcast already signed messages from WAL

### LINKED ISSUE #474: Record messages for rebroadcast after signing
We currently record them _before_ signing. Unfortunately, this isn't safe it a participant changes from one client to the other. To be clear, changing from one client to another isn't safe in general but... this is _extra_ unsafe.

This will also let us directly use the rebroadcast queue to fix #392.

masih:
Previously, rebroadcast functioned by keeping track of message templates and calling `RequestBroadcast` again to re-broadcast them. This approach can cause erroneous behaviour when messages are re-signed in conjunction with participant churn.

To avoid this category of errors treat broadcast as final: Every broadcasted message is recorded in WAL and once recorded the same message is used for re-broadcast; no re-signing. A dedicated re-broadcast API is introduced to take instance, round and phase and use it to then look up broadcast messages.

Note that as a result of the changes above, failures are also sicky. In that, if signing of a message fails at the time of original broadcast, future requests to rebroadcast it will have no effect.

The changes here also include a series of WAL functionality refactors to make it easier to traverse write-ahead log entries.

Fixes #474


emulator/host.go
```diff
@@ -48,6 +50,17 @@ func (h *driverHost) RequestBroadcast(mb *gpbft.MessageBuilder) error {
...
+	require.True(h.t, h.messages.PutIfAbsent(msg))
+	return nil
+}
+
+func (h *driverHost) RequestRebroadcast(instance, round uint64, phase gpbft.Phase) error {
+	message, found := h.messages.Get(instance, round, phase)
+	if found {
+		h.receivedBroadcasts = append(h.receivedBroadcasts, message)
+	} else {
+		fmt.Printf("asdasd")
```
Stebalien, line 62: removeme?

gpbft/api.go
```diff
@@ -94,6 +94,9 @@ type Network interface {
 	NetworkName() NetworkName
 	// Requests that the message is signed and broadcasted, it should also be delivered locally
 	RequestBroadcast(mb *MessageBuilder) error
+	// RequestRebroadcast requests that a message at given instance, round and phase
+	// previously broadcasted via RequestBroadcast be rebroadcasted.
+	RequestRebroadcast(instance, round uint64, phase Phase) error
```
Stebalien, line 99: nit: maybe it's time to create some form of `gpbft.Instant` struct specifying an instance/round/phase?

gpbft/gpbft.go
```diff
@@ -959,27 +959,27 @@ func (i *instance) rebroadcast() error {
...
 			// Silently log the error and proceed. This is consistent with the behaviour of
 			// instance for regular broadcasts.
-			i.log("failed to request rebroadcast %s at round %d: %v", mb.Payload.Phase, mb.Payload.Round, err)
+			i.log("failed to request rebroadcast %s at round %d: %v", request.phase, request.round, err)
 		} else {
-			i.log("rebroadcasting %s at round %d for value %s", mb.Payload.Phase.String(), mb.Payload.Round, mb.Payload.Value)
+			i.log("rebroadcasting %s at round %d", request.phase, request.round)
 			metrics.reBroadcastCounter.Add(context.TODO(), 1)
 		}
 	}
```
Stebalien, line 985: This works, but can't we also just not record anything? That is:

1. If we're in the decide phase, blindly request rebroadcast for decide.
2. If we're in any other phase, blindly request rebroadcast for each phase in the previous/current rounds, up to the current phase?

It doesn't matter if we actually sent any messages at those points in time.

Stebalien (commented):
couple of early comments, still reviewing

host.go
```diff
@@ -522,7 +519,7 @@ func (h *gpbftRunner) Progress() (instance, round uint64, phase gpbft.Phase) {
 // These will be used as input to a subsequent instance of the protocol.
 // The chain should be a suffix of the last chain notified to the host via
 // ReceiveDecision (or known to be final via some other channel).
-func (h *gpbftHost) GetProposal(instance uint64) (_ *gpbft.SupplementalData, _ gpbft.ECChain, _err error) {
+func (h *gpbftRunner) GetProposal(instance uint64) (_ *gpbft.SupplementalData, _ gpbft.ECChain, _err error) {
```
Stebalien, line 522: any reason to change this?

internal/writeaheadlog/message_wal.go
```diff
@@ -0,0 +1,54 @@
...
+	// partial delivery anyway.
+	return matches, mwal.ForEach(
+		func(subject *gpbft.GMessage) bool {
+			if subject.Vote.Instance == instance && subject.Vote.Round == round && subject.Vote.Phase == phase {
+				matches = append(matches, subject)
+			}
+			return true
+		},
+	)
+}
```
Stebalien, line 50: This will re-read from disk every time we rebroadcast? That's not great.

Can we not create an index on start?

internal/writeaheadlog/wal.go
```diff
@@ -231,61 +250,50 @@ func (wal *WriteAheadLog[T, PT]) hydrate() error {
...
-		var single T
-		err = PT(&single).UnmarshalCBOR(logFileReader)
-		if err == nil {
-			if keepVaues {
-				content = append(content, single)
+	defer func() {
+		if err := target.Close(); err != nil {
+			log.Errorw("failed to close log while reading", "name", logName, "err", err)
+		}
+	}()
```
Stebalien, line 284: return this error? I guess it isn't critical...

emulator/host.go
masih: oops

host.go
masih: Because it was a nightmare navigating between runner and host, when they are essentially the same. Without this change one has to cast the runner to broadcast etc. etc. 

I am willing to be convinced this type separation is necessary.

internal/writeaheadlog/message_wal.go
masih: Are we OK keeping the entire WAL in memory?

gpbft/api.go
masih: That's fine as long as `Instant` and `instance` are not confusing to the reader.

masih: I will do this in a separate PR if that's not a blocker, since i'd like to work it into Progress too.

Stebalien: Yep, not critical. Just something that keeps coming up.

host.go
Stebalien: No preference. The only reason we had the split was to make it clear which methods were gpbft-facing and which were client-facing.

internal/writeaheadlog/message_wal.go
Stebalien: At least for a fixed number of instances, yes. As @Kubuxu pointed out last time we talked, given the exponential backoff nature of rounds, we'll never have to remember that many messages.

Stebalien: (I'd prefer to just remember one instance, but whatever is easiest).

internal/writeaheadlog/message_entry.go
```diff
@@ -0,0 +1,26 @@
...
+}
+
+func (we *MessageEntry) MarshalCBOR(w io.Writer) error {
+	return we.Message.MarshalCBOR(w)
+}
+
+func (we *MessageEntry) UnmarshalCBOR(r io.Reader) error {
+	we.Message = &gpbft.GMessage{}
+	return we.Message.UnmarshalCBOR(r)
+}
```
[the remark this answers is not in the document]
masih: This would mean a bunch of nil checks for `Message` field, which is more hassle than it's worth?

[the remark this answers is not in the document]
Stebalien: How? Anyways, this doesn't really matter. No point in spending any time on it.

internal/writeaheadlog/message_wal.go
masih: The limit of 5 instances in WAL is externally enforced, via runner start loop. If for whatever reason logs are partially removed the hydration will load all the logs.

I am going to move that cap into WAL itself for safety.

internal/writeaheadlog/message_entry.go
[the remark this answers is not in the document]
masih: Because of pointer trickery in WAL, e.g. `PT(&entry)`. I honestly think we should remove the generics. But that's a discussion for another day.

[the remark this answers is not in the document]
Stebalien: Oh. Got it.

host.go
masih: > to make it clear which methods were gpbft-facing and which were client-facing

For what it's worth I as a maintainer find it confusing and too expensive at the price of casting.

Stebalien: Two casts? I don't think it added much anyways (the two types was a replacement for two entirely separate structs), but this kind of newtyping really isn't that uncommon.

internal/writeaheadlog/message_wal.go
Stebalien: This is going to lead to a whole bunch of unnecessary code churn, IMO. From what I can tell, all the WAL changes were introduced primarily to support `ForEach`, which isn't necessary in this case.

masih: OK
Just so I get it right: instead of `forEach`, you'd like to see call to `All`, then filter. Correct?

masih: Discussed on slack.

host.go
masih: Since this is two votes against one, i will revert this and cast.

masih: > this kind of newtyping really isn't that uncommon.

It is my opinion that this practice is awful.

Stebalien: Eh, I've gotten used to it at this point.

host.go
```diff
@@ -63,31 +70,52 @@ func newRunner(
...
 		return nil, fmt.Errorf("reading WAL: %w", err)
 	}
+	var maxInstance uint64
 	for _, v := range walEntries {
 		runner.equivFilter.ProcessBroadcast(v.Message)
+		instance := v.Message.Vote.Instance
+		if runner.selfMessages[instance] == nil {
+			runner.selfMessages[instance] = make(map[roundPhase]*gpbft.GMessage)
+		}
+		// WAL is dump. To avoid relying on it returning sorted entries or making it
```
Stebalien, line 105:
```suggestion
		// WAL is dumb. To avoid relying on it returning sorted entries or making it
```

Stebalien: Also an option to just sort by instance with `slices.SortFunc`.

host.go
```diff
@@ -232,6 +260,11 @@ func (h *gpbftRunner) Start(ctx context.Context) (_err error) {
 						log.Errorw("failed to purge messages from WAL", "error", err)
 					}
 				}
+				for instance := range h.selfMessages {
+					if instance < cert.GPBFTInstance {
+						delete(h.selfMessages, instance)
```
Stebalien, line 265: locks?

host.go
```diff
@@ -333,6 +366,14 @@ func (h *gpbftRunner) BroadcastMessage(msg *gpbft.GMessage) error {
 		log.Errorw("appending to WAL", "error", err)
 	}
 
+	if h.selfMessages[msg.Vote.Instance] == nil {
```
Stebalien, line 369: Also probably need a lock.

host.go
```diff
@@ -474,6 +515,16 @@ var (
 // gpbftHost is a newtype of gpbftRunner exposing APIs required by the gpbft.Participant
 type gpbftHost gpbftRunner
 
+func (h *gpbftHost) RequestRebroadcast(instance, round uint64, phase gpbft.Phase) error {
+	if roundPhaseMessages, found := h.selfMessages[instance]; found {
```
Stebalien, line 519: lock

host.go
Stebalien: But note: I'm not going complain if you want to re-do this change in a new PR.

host.go
masih: > Also an option to just sort by instance with `slices.SortFunc`.

Even then I need to iterate to find the right index. we are already traversing all the messages for the filter. At this point returns are diminishing fast on trying to polish WAL interaction.

host.go
```diff
@@ -474,6 +523,18 @@ var (
 // gpbftHost is a newtype of gpbftRunner exposing APIs required by the gpbft.Participant
 type gpbftHost gpbftRunner
 
+func (h *gpbftHost) RequestRebroadcast(instance, round uint64, phase gpbft.Phase) error {
+	h.msgsMutex.Lock()
+	defer h.msgsMutex.Unlock()
+	if roundPhaseMessages, found := h.selfMessages[instance]; found {
+		if msg, found := roundPhaseMessages[roundPhase{round: round, phase: phase}]; found {
+			obfuscatedHost := (*gpbftRunner)(h)
+			return obfuscatedHost.BroadcastMessage(msg)
```
Stebalien, line 532: We probably don't want to hold the lock while calling this.

Stebalien (approved):
LGTM with one change.

host.go
Stebalien: Looks like that's definitely (test deadlock).

Stebalien (approved):

host.go
```diff
@@ -46,6 +47,15 @@ type gpbftRunner struct {
 	runningCtx context.Context
 	errgrp     *errgroup.Group
 	ctxCancel  context.CancelFunc
+
+	// msgsMutex guards access to selfMessages
+	msgsMutex    sync.Mutex
+	selfMessages map[uint64]map[roundPhase]*gpbft.GMessage
```
Kubuxu, line 53: There can be multiple messages for a single (instance, round, phase) as we can have participating IDs.

host.go
```diff
@@ -333,6 +373,17 @@ func (h *gpbftRunner) BroadcastMessage(msg *gpbft.GMessage) error {
...
 
+	h.msgsMutex.Lock()
+	if h.selfMessages[msg.Vote.Instance] == nil {
+		h.selfMessages[msg.Vote.Instance] = make(map[roundPhase][]*gpbft.GMessage)
+	}
+	key := roundPhase{
+		round: msg.Vote.Round,
+		phase: msg.Vote.Phase,
+	}
+	h.selfMessages[msg.Vote.Instance][key] = append(h.selfMessages[msg.Vote.Instance][key], msg)
```
Kubuxu, line 384: This is called by re-broadcast, which will exponentially increase the number of messages.

Kubuxu: We are also WALing the re-boradcasted messages, which is probably not what we want to do. I suggest to have a different path for sending out re-broadcast messages that only checks the equivocation filter.

host.go
```diff
@@ -474,6 +543,26 @@ var (
 // gpbftHost is a newtype of gpbftRunner exposing APIs required by the gpbft.Participant
 type gpbftHost gpbftRunner
 
+func (h *gpbftHost) RequestRebroadcast(instance, round uint64, phase gpbft.Phase) error {
+	var rebroadcasts []*gpbft.GMessage
+	h.msgsMutex.Lock()
+	if roundPhaseMessages, found := h.selfMessages[instance]; found {
+		if messages, found := roundPhaseMessages[roundPhase{round: round, phase: phase}]; found {
+			rebroadcasts = make([]*gpbft.GMessage, 0, len(messages))
+			rebroadcasts = append(rebroadcasts, messages...)
```
Kubuxu, line 552:
```suggestion
			rebroadcasts = slices.Clone(messages)
```

Kubuxu (approved):

PR #694, opened 2024-10-03
Resume instance by sending messages from self

### LINKED ISSUE #679: Handle resumption/rebroadcast via WAL
Now that there is WAL, ideally a participant should be able to resume progress within a GPBFT instance if restarted. To do so the participant only requires to rebroadcast messages that it would have done prior to restart (i.e. `broadcastState`). If all participants do this then it should be enough to resume participation without any further state-keeping or bespoke steps after restart.

masih:
WAL keeps track of all the messages and we cache messages from the latest instance for rebroadcast purposes.

Before starting a new instance, if there are any messages from self for that instance feed them to the participant. That should speed up catch up at that round and effectively aid resume progress for the participant.

Fixes #679

### DIFF
--- a/host.go
+++ b/host.go
@@ -152,7 +152,7 @@ func (h *gpbftRunner) Start(ctx context.Context) (_err error) {
 			log.Errorf("error when receiving certificate: %+v", err)
 		}
 	default:
-		if err := h.participant.StartInstanceAt(h.manifest.InitialInstance, h.clock.Now()); err != nil {
+		if err := h.startInstanceAt(h.manifest.InitialInstance, h.clock.Now()); err != nil {
 			log.Errorf("error when starting instance %d: %+v", h.manifest.InitialInstance, err)
 		}
 	}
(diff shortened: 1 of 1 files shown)


host.go
```diff
@@ -289,6 +289,28 @@ func (h *gpbftRunner) receiveCertificate(c *certs.FinalityCertificate) error {
 	log.Infow("skipping forwards based on cert", "from", currentInstance, "to", nextInstance)
 
 	nextInstanceStart := h.computeNextInstanceStart(c)
+
+	// Look for any existing messages in WAL for the next instance, and if there is
```
Stebalien, line 293: Hm. I guess technically we'll see this on start. Although we'll never see it for instance 0.

Any reason not to replay everything immediately in `gpbftRunner.Start`?

Stebalien: Or we can look at the current instance at that point.

masih: That would take more gymnastics to get the current instance 🤷 

What I have done in the latest changes is to refactor calls to participant's `StartInstanceAt`, and replay message for the instance being started every time. 

Please let know what you think.

Stebalien: Oh, that's much nicer.

Stebalien (approved):

host.go
```diff
@@ -289,7 +289,32 @@ func (h *gpbftRunner) receiveCertificate(c *certs.FinalityCertificate) error {
...
+		for _, message := range messages {
+			replay = append(replay, message...)
+		}
+	}
+	h.msgsMutex.Unlock()
+	for _, message := range replay {
+		if validated, err := h.participant.ValidateMessage(message); err != nil {
+			log.Warnw("invalid self message", "message", message, "err", err)
+		} else if err := h.participant.ReceiveMessage(validated); err != nil {
+			log.Warnw("failed to send resumption message", "message", message, "err", err)
```
Kubuxu, line 314: The only thing I'm worried is the ordering of messages but it is probably fine.

Kubuxu (approved):

host.go
masih: It should just work, but no harm in sorting them for a more optimal resumption 👍

masih: OK added

Kubuxu (approved):

PR #700, opened 2024-10-07
Investigate flaky tests on CI

masih:
Set up a temporary workflow to run all flaky tests repeatedly on CI for further investigation.

### DIFF
--- a/.github/workflows/test-flaky.yml
+++ b/.github/workflows/test-flaky.yml
@@ -1,0 +1,54 @@
+name: Test Repeat Flaky
+
+on:
+  pull_request:
+
+jobs:
+  cache-test-bins:
+    name: Prepare go test cache
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        go:
+          - 1.22
+          - 1.23
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Setup Go ${{ matrix.go }}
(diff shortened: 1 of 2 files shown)


Stebalien:
Well, this looks like it helped a bit.

Stebalien:
Hm. Well, I'm going to try increasing the timeout.

masih:
@Stebalien Tests seem to only fail when run in conjunction with other tests. To me so far this seems like some kind of congestion issue deep in libp2p/pubsub stack.

Stebalien:
> Tests seem to only fail when run in conjunction with other tests. To me so far this seems like some kind of congestion issue deep in libp2p/pubsub stack.

Or just a timing issue? We're using a mock network, so at least it's not some kind of port issue.

But yeah, it sounds like the solution is to simply not run them in parallel.

masih:
Findings:
* Repeated individual tests do not fail (repeated up to 100 times each).
* They do fail when run in conjunction with other tests.
  * More specifically in conjunction with other tests in `f3_test.go`
* These tests are time-based, and use an artificial clock. Having reviewed the artificial clock code nothing immediately stands out.
* The cause of failure is waiting for a condition (e.g. instance number) that is never satisfied regardless of how long we wait (tested up to 30m timeout).
* Tests pass consistently after removing `t.Parallel()` (Repeated 50 times).

PR #701, opened 2024-10-07
Bump version file to 0.6.0

masih:
Prep for release `0.6.0`


masih:
Fuzz failure is with exit code 143 which signals job was killed from outside, in this case most likely by GitHub actions. Re running it.

PR #702, opened 2024-10-07
Apply manifest update immediately after receipt

Stebalien:
That way we:

1. Stop immediately.
2. Expose the manifest to the user immediately so they can see which network we're going to participate in (important for the miner participation API).

### DIFF
--- a/f3.go
+++ b/f3.go
@@ -128,7 +128,8 @@ func (m *F3) GetCert(ctx context.Context, instance uint64) (*certs.FinalityCerti
 
 // Returns the time at which the F3 instance specified by the passed manifest should be started, or
 // 0 if the passed manifest is nil.
-func (m *F3) computeBootstrapDelay(manifest *manifest.Manifest) (time.Duration, error) {
+func (m *F3) computeBootstrapDelay() (time.Duration, error) {
+	manifest := m.manifest.Load()
 	if manifest == nil || manifest.Pause {
 		return 0, nil
 	}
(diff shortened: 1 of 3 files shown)


f3_test.go
```diff
@@ -195,12 +195,13 @@ func TestF3DynamicManifest_WithRebootstrap(t *testing.T) {
 	// check that it rebootstrapped and has a new base epoch.
 	targetBaseEpoch := env.manifest.BootstrapEpoch - env.manifest.EC.Finality
 	env.waitForCondition(func() bool {
+		env.clock.Add(env.manifest.EC.Period)
```
Stebalien, line 198: We now get the manifest immediately so `waitForManifest` will return long before the bootstrap epoch. So we need to wait longer here.

f3.go
```diff
@@ -208,25 +211,34 @@ func (m *F3) Start(startCtx context.Context) (_err error) {
...
-				if err := m.reconfigure(m.runningCtx, pendingManifest); err != nil {
-					return fmt.Errorf("failed to reconfigure F3: %w", err)
+				if err := m.startInternal(m.runningCtx); err != nil {
+					// There's not much we can do here but complain and wait for
+					// a new manifest. This matches the behavior when we
+					// initially start as well.
+					log.Errorw("failed to start F3", "error", err)
 				}
-				pendingManifest = nil
+				hasPendingManifest = false
```
masih, line 241: pending set to false even if start internal fails. That's fine and the expectation is that we will retry eventually. Right?

masih (approved):

f3.go
Stebalien: The expectation is that retrying wouldn't help. We'll either get a new manifest over the network, or we'll stay in the failed state until we restart.

PR #704, opened 2024-10-07
Atomically return manifest + instant

Stebalien:
Otherwise, there's no (nice) atomic way to get both the current progress and the current network.

- I'm renaming this to `Status` because it does more than return the current instant. It also means we have the trifecta: start/stop/status.
- I've removed the separate `Manifest` method because there really wasn't any need for it. Maybe I'm being over-eager to break things but... this gives us a single-atomic way to inspect the state (well, except "is running").

Feel free to tell me this too much...

### DIFF
--- a/f3.go
+++ b/f3.go
@@ -88,10 +88,6 @@ func (m *F3) MessagesToSign() <-chan *gpbft.MessageBuilder {
 	return m.outboundMessages
 }
 
-func (m *F3) Manifest() *manifest.Manifest {
-	return m.manifest.Load()
-}
-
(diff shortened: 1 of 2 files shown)


Stebalien:
Note: this isn't _strictly_ necessary.

Stebalien:
I also kind of don't want to do this.

masih (approved):
This looks correct to me too. But I am not sure about the ergonomics. Hopefully we can revisit all of this once there is no dynamic manifest and make it all more concise.

No blockers from me.

Stebalien:
So, we've merged all the fixes in lotus and, push comes to shove, we'll end up returning an unusable ticket. So I'm not going to merge this change.

PR #708, opened 2024-10-08
Disperse tests in `test` package to where they belong

masih:
Originally, all tests were put in `test` package to avoid circular dependency. This no longer holds true. Therefore:

* Move all simulation tests to `sim`
* Move `sim/signing` package to `internal` since it is used by a number of tests across the repo as well as `cmd/f3`.
* Move signing suite tests to the now existing `internal/signing`

These changes keep with the go convention of keeping tests in the package they are testing.


Stebalien:
IMO, they belong in a separate test package. Maybe "sim/test", but they could also live in some `internal/test` package (really `sim` should live in `internal` as well).

Basically, I'd expect any tests directly in `sim` to be testing the simulation itself (e.g., `message_queue_test.go`).

Stebalien:
If we were to strictly follow the rule "tests live next to the components they test", they'd live in `gpbft` itself. But it's kind of nice to have these in a separate package given how slow they are.

PR #710, opened 2024-10-08
Run cbor gen in parallel

masih:
Run cbor gen in parallel for faster generation.


Stebalien (approved):

Kubuxu:
I'm not sure this is safe as these packages have cross-dependencies on each other, and cbor-gen will use go toolkit to resolve symbols, but if it works, it works.

Stebalien:
It's fine because we've already compiled so reflection uses the pre-compiled code.

Kubuxu:
Ahh, right.

Stebalien:
(I had the same "how does this even work" reaction)

PR #713, opened 2024-10-09
Attempt publication with circuit breaker

### LINKED ISSUE #632: Consider backing off if RequestBroadcast errors
The GPBFT implementation here silently logs errors that occur as a result of requesting broadcast.

In practice, if such situations occur it may be better to backoff from sending consecutive requests, e.g. that may be caused by pubsub congestion.

_Originally posted by @Stebalien in https://github.com/filecoin-project/go-f3/pull/627#discussion_r1745760404_

masih:
GPBFT silently ignores broadcast and rebroadcast requests as it is beyond its boundaries of responsibility to do something about. In practice such failures may be a sign that pubsub is overwhelmed with messages. Therefore, ideally the system should avoid aggravating the situation by requesting further broadcasts. This is specially important in re-broadcast requests because it often involves batch message publication.

The changes here wrap the pubsub publication calls with a circuit breaker that will open on consecutive errors (set to `5`), and will not attempt to actually publish messages until a reset timeout (set to `3s`) has passed.

Fixes #632

### DIFF
--- a/host.go
+++ b/host.go
@@ -13,6 +13,7 @@ import (
 	"github.com/filecoin-project/go-f3/certstore"
 	"github.com/filecoin-project/go-f3/ec"
 	"github.com/filecoin-project/go-f3/gpbft"
+	"github.com/filecoin-project/go-f3/internal/circuitbreaker"
 	"github.com/filecoin-project/go-f3/internal/clock"
 	"github.com/filecoin-project/go-f3/internal/psutil"
 	"github.com/filecoin-project/go-f3/internal/writeaheadlog"
@@ -41,6 +42,7 @@ type gpbftRunner struct {
 
 	participant *gpbft.Participant
 	topic       *pubsub.Topic
+	cb          *circuitbreaker.CircuitBreaker
 
 	alertTimer *clock.Timer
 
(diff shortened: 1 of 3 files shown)


host.go
```diff
@@ -86,6 +88,7 @@ func newRunner(
 		ctxCancel:    ctxCancel,
 		equivFilter:  newEquivocationFilter(pID),
 		selfMessages: make(map[uint64]map[roundPhase][]*gpbft.GMessage),
+		cb:           circuitbreaker.New(5, 3*time.Second),
```
masih, line 91: Thoughts on defaults?

Stebalien:
I'm not really sure how this will help. We'll drop messages either way, this just means we'll drop messages when we don't actually need to do so. Maybe it'll help, I just can't think of how.

My original thinking was to have some form of retry setup. E.g.:

1. Put all messages in a queue.
2. Try sending messages out of the queue, backing off and retrying as necessary.
3. Drop all messages currently in the queue when the rebroadcast timer fires.

masih:
I wish this feedback was added to the issue sooner/when it was captured. 

Thanks for sharing this. I'll take another shot at this.

Stebalien:
Yeah, sorry, I saw that the issue was filed and didn't pay much attention. Also, I'm not 100% sure either way. I feel like we should do _something_ about failures but it's unclear exactly what.

masih:
Nah, i think what you suggested makes sense. For some reason I assumed the retry is built into pubsub and if it errors it means things are not recoverable or congested sufficiently such that the only thing that would help is to drop requests and conservatively retry. Otherwise, if the cause of error is congestion backoff alone could be worse, in that it could cause a prolonged slow publication flow. 

I'd like to take a closer look at pubsub implementation for a more educated attempt.

Stebalien:
IIRC, it's caused by the queue filling up. Which means the main issue is bursting, so backing off and retrying _should_ help as long as we don't let our queue/retry grow indefinitely.

PR #723, opened 2024-10-25
Don't use GetTipset(cert.Head()) to compute startup delay

### LINKED ISSUE #720: Use available information when computing the next instance start
We currently try to get the recently finalized head tipset:

https://github.com/filecoin-project/go-f3/blob/5fc96af6b505faa6cc254bcf1de772378edf6aeb/host.go#L349-L354

However, if it's on a fork, we may not have it. Instead, we should get our current head and use the finalized head's epoch, our head epoch, our head timestamp, and the EC period to compute the correct delay.

Kubuxu:
Resolves #720


host.go
```diff
@@ -346,14 +346,22 @@ func (h *gpbftRunner) startInstanceAt(instance uint64, at time.Time) error {
...
-	base := ts.Timestamp()
+	// the head of the cert becomes the new base
+	baseTipSet := cert.ECChain.Head()
+	// we are not trying to fetch the new base tipset from EC as it might not be available
+	// instead we compute the relative time from the EC.Head
+	// timestamp(head) = genesis + epoch(head) * period
+	// timestamp(base) = genesis + epoch(base) * period + timestamp(head) - timestamp(head)
+	// timestamp(base) = timestamp(head) + genesis + epoch(base) * period - genesis - epoch(head) * period
+	// timestamp(base) = timestamp(head) + (epoch(base) - epoch(head)) * period
+	baseTimestamp := head.Timestamp().Add(time.Duration(baseTipSet.Epoch-head.Epoch()) * h.manifest.EC.Period)
```
Stebalien, line 364: I'm going to be hypocritical here but... it's easy to have hard-to-notice off-by-one errors here. Can we extract this into a separate function and write a simple unit test?

Stebalien (approved):

PR #731, opened 2024-10-29
Test parquet files are readable

masih:
Test should panic with:

```
panic: opening columns of parquet file: ParquetEnvelope: Message: Justification: Vote: Value: row group at index 0 does not have enough columns
```


masih:
prquet export and observer is redone in #744. Failing tests here now no longer matter. Closing.

PR #734, opened 2024-11-06
Fix chain comparison bug and add tests

masih:
Return false if chains of same length are not equal and assert assert it in unit tests.


gpbft/chain.go
```diff
@@ -83,7 +83,7 @@ func (ts *TipSet) Equal(b *TipSet) bool {
 	}
 	return ts.Epoch == b.Epoch &&
 		bytes.Equal(ts.Key, b.Key) &&
-		ts.PowerTable == b.PowerTable &&
+		ts.PowerTable.Equals(b.PowerTable) &&
```
masih, line 86: This is one is a cosmetic change; just to use `cid.Equals` because it exists.

Kubuxu (approved):

PR #739, opened 2024-11-07
run on a faster machine

Stebalien:


masih:
I think for custom runners to work the repo needs to be white listed by whatever running the runners. For example, the runners used by Lotus repo are (i believe) only white listed to for that repo.

PR #740, opened 2024-11-08
Rename OhShitStore™ to `powerstore`

masih:
Cosmetic renaming of the powerstore.

### DIFF
--- a/internal/powerstore/powerstore.go
+++ b/internal/powerstore/powerstore.go
@@ -22,7 +22,7 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
-var log = logging.Logger("f3/ohshitstore")
+var log = logging.Logger("f3/powerstore")
 
 var _ ec.Backend = (*Store)(nil)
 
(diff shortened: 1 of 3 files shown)


masih:
@rjan90 I am not sure if the logs from OhShitStore are mentioned anywhere in F3 SP guides. We probably want to rename them if there is any.

internal/powerstore/powerstore.go
```diff
@@ -50,7 +50,7 @@ func New(ctx context.Context, ec ec.Backend, ds datastore.Datastore, cs *certsto
 	return &Store{
 		Backend: ec,
 
-		ds:         namespace.Wrap(ds, datastore.NewKey("/ohshitstore")),
+		ds:         namespace.Wrap(ds, datastore.NewKey("/powerstore")),
```
masih, line 53: This is the most significant change here, where it effectively clears the store.

rjan90:
> @rjan90 I am not sure if the logs from OhShitStore are mentioned anywhere in F3 SP guides. We probably want to rename them if there is any.

No, we currently do not have them mentioned in the guide. But I plan to cover the meaning of these two log lines:
- 2024-10-24T17:15:03.321Z	WARN	f3/ohshitstore	powerstore/powerstore.go:252	Uh-oh, F3 has fallen behind EC by more than 1350 epochs! Engaging the OhShitStore™.	{"ecHead": 2082124, "f3Base": 2080774}
- 2024-10-24T21:40:03.320Z	INFO	f3/ohshitstore	powerstore/powerstore.go:259	Stopping the OhShitStore™ because we're caught-up	{"ecHead": 2082654, "f3Base": 2082270}

I will change the naming for those.

PR #741, opened 2024-11-08
Bump version file to `0.7.3` in prep for release

masih:
Bump version in prep for release to bubble up refactorings to Lotus.


masih:
Pending sanity check on archioz before cutting the tag

Stebalien:
@Kubuxu says that the store isn't corrupted and fixing the logic "just works", so we're leaving this as-is.

PR #744, opened 2024-11-12
Implement observer with SQL query support and parquet export

masih:
Implement an F3 observer that:
* listens to F3 gossipsub messages
* collects them as long as they can be decoded as CBOR
* offers a SQL interface to query them over a HTTP
* rotates them into Parquet files
* deletes any rotated files based on a configured maximum age


cmd/f3/msgdump/envelope.go
```diff
@@ -1,75 +0,0 @@
-package msgdump
```
masih, line 1: Rewrote the stuff here, because:
* there seem to be an [issue with decoding parquet files previously generated](https://github.com/filecoin-project/go-f3/pull/731).  So far we have just been using the ndjson dump to query things.
* I wanted to have a SQL interface that: 1) can query the latest messages and 2) can query the rotated messages all in one convenient view.
* These changes are significant enough that it deserved its own package etc.

cmd/f3/pubsub.go
```diff
@@ -0,0 +1,22 @@
+package main
```
masih, line 1: Renamed the file due to typo in file name and moved common pubsub utility stuff from here to internal/psutil package.

observer/query.go
```diff
@@ -0,0 +1,111 @@
+package observer
```
masih, line 1: This allows us to run the observer under F3 passive testing infra as a HTTP server and CURL it with SQL statements, e.g.:

```
$ curl -X POST -d 'select distinct sender from messages'  https://where-it-is-gonna-run/query
```

observer/query.go
```diff
@@ -0,0 +1,111 @@
...
+			http.Error(w, fmt.Sprintf("Cannot read request: %s", err), http.StatusBadRequest)
+			return
+		}
+		dbQuery = string(body)
+	default:
+		http.Error(w, fmt.Sprintf("%s: %s", http.StatusText(http.StatusUnsupportedMediaType), requestMediaType), http.StatusUnsupportedMediaType)
+		return
+	}
+
+	rows, err := o.db.QueryContext(r.Context(), dbQuery)
```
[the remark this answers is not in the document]
masih: This is OK because the server is considered an admin endpoint. We do this on purpose.

observer/model.go
```diff
@@ -0,0 +1,104 @@
...
+		// Currently, all Commitments are always empty. For completeness and reducing
+		// future schema changes include them anyway when they are non-empty.
+		commitments = gp.SupplementalData.Commitments[:]
+	}
+
+	value := make([]tipSet, len(gp.Value))
+	for i, v := range gp.Value {
+		value[i] = tipSet{
+			// The TipSet Commitments and PowerTable CID always match the Payload's
+			// SupplementalData for valid messages. Therefore, dropped for brevity.
```
Kubuxu, line 89: That is not true, SupplementalData carries PowerTableCID for verification of next instance, these values are ones that are getting finalized

Kubuxu (approved):

observer/model.go
masih: oo I will include them in the data in that case. Thanks Kuba

Kubuxu:
My only nit would be maybe support for BasicAuth, but I don't think it is immediately necessary as it can be reverse proxied onto it.

masih:
> support for BasicAuth

That's easily configurable at infra. I'd like to avoid having bespoke code for it here.

I'll add basic auth to our deployment.

PR #775, opened 2024-12-03
Implement cert chain generator/verifier from EC

### LINKED ISSUE #743: Have a canonical way to take a FakeEC and build certs from it and/or compare that certs are corresponding to given chain
Would prevent the bug in OhShitStore.
Provide a way to form certs from FakeEC and/or compare FakeEC with external certs.
Provides an extra assertion for testing GPBFT correctness (which we don't really do right now a lot) as well probably simplifies some tests (provide certs to nodes instead of running GPBFT).

masih:
Implement testing utilities to:
* generate random finality certificate chain given an `ec.Backend`
* verify conformity of an existing chain with a given  `ec.Backend`

Fixes #743


Stebalien:
Nit: put this into an internal package?

masih:
> Nit: put this into an internal package?

I need to use it in lotus to verify the cert chain of a network, e.g. calibnet.

Stebalien:
Ah, I see. We're trying to unify our testing code and our actual code. Makes sense.

PR #784, opened 2024-12-10
Introduce caching of PowerTable CIDs

Kubuxu:
First commit is a refactor, second one is the change.

### DIFF
--- a/consensus_inputs.go
+++ b/consensus_inputs.go
@@ -1,0 +1,230 @@
+package f3
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"slices"
+	"time"
+
+	"github.com/filecoin-project/go-f3/certs"
+	"github.com/filecoin-project/go-f3/certstore"
+	"github.com/filecoin-project/go-f3/ec"
+	"github.com/filecoin-project/go-f3/gpbft"
+	"github.com/filecoin-project/go-f3/internal/clock"
+	"github.com/filecoin-project/go-f3/manifest"
+	"go.opentelemetry.io/otel/metric"
+)
+
+type gpbftInputs struct {
+	manifest  *manifest.Manifest
+	certStore *certstore.Store
+	ec        ec.Backend
+	verifier  gpbft.Verifier
+	clock     clock.Clock
+}
+
+//ptCache lru.Cache[string, cid.Cid]
+//}
+//func (h *gpbftInputs) getPowerTableCIDForTipset(ctx context.Context, tsk gpbft.TipSetKey) {
+
+//}
+
+func (h *gpbftInputs) collectChain(ctx context.Context, base ec.TipSet, head ec.TipSet) ([]ec.TipSet, error) {
+	// TODO: optimize when head is way beyond base
+	res := make([]ec.TipSet, 0, 2*gpbft.ChainMaxLen)
+	res = append(res, head)
+
+	current := head
+	for !bytes.Equal(current.Key(), base.Key()) {
+		if current.Epoch() < base.Epoch() {
+			metrics.headDiverged.Add(ctx, 1)
+			log.Infow("reorg-ed away from base, proposing just base",
+				"head", head.String(), "base", base.String())
+			return nil, nil
+		}
+		var err error
+		current, err = h.ec.GetParent(ctx, current)
+		if err != nil {
+			return nil, fmt.Errorf("walking back the chain: %w", err)
+		}
+		res = append(res, current)
+	}
+	slices.Reverse(res)
+	return res[1:], nil
+}
+
+// Returns inputs to the next GPBFT instance.
+// These are:
+// - the supplemental data.
+// - the EC chain to propose.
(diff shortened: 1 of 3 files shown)


consensus_inputs.go
```diff
@@ -0,0 +1,258 @@
...
+
+	ptCache *lru.Cache[string, cid.Cid]
+}
+
+func newInputs(manifest *manifest.Manifest, certStore *certstore.Store, ec ec.Backend,
+	verifier gpbft.Verifier, clk clock.Clock) gpbftInputs {
+	cache, err := lru.New[string, cid.Cid](256) // keep a bit more than 2x max ECChain size
+	if err != nil {
+		// panic as it only depends on the size
+		panic(fmt.Errorf("could not create cache: %w", err))
```
masih, line 36: I would return an error based on the principle that panics should be used rarely. I understand there are pre-existing practices in this repo on use of panic. So not a blocker for me.

consensus_inputs.go
```diff
@@ -0,0 +1,258 @@
...
+	"github.com/filecoin-project/go-f3/ec"
+	"github.com/filecoin-project/go-f3/gpbft"
+	"github.com/filecoin-project/go-f3/internal/clock"
+	"github.com/filecoin-project/go-f3/manifest"
+	lru "github.com/hashicorp/golang-lru/v2"
+	"github.com/ipfs/go-cid"
+	"go.opentelemetry.io/otel/metric"
+)
+
+type gpbftInputs struct {
```
masih, line 21: Tests would be great for this, now that the logic committee and proposal providers are refactored out. OK to add those in separate PRs.

consensus_inputs.go
```diff
@@ -0,0 +1,258 @@
...
+		certStore: certStore,
+		ec:        ec,
+		verifier:  verifier,
+		clock:     clk,
+		ptCache:   cache,
+	}
+}
+
+func (h *gpbftInputs) getPowerTableCIDForTipset(ctx context.Context, tsk gpbft.TipSetKey) (cid.Cid, error) {
+	sTSK := string(tsk)
```
masih, line 50: How big do tipset keys get? do we want to hash these?

consensus_inputs.go
```diff
@@ -0,0 +1,258 @@
...
+	ptCid, err = certs.MakePowerTableCID(pt)
+	if err != nil {
+		return cid.Undef, fmt.Errorf("computing power table CID: %w", err)
+	}
+
+	h.ptCache.Add(sTSK, ptCid)
+	return ptCid, nil
+}
+
+func (h *gpbftInputs) collectChain(ctx context.Context, base ec.TipSet, head ec.TipSet) ([]ec.TipSet, error) {
```
masih, line 69: This function would collect the entire chain that _could_ be finalised irrespective of max chain length of 100. Right?

Specifically, during bootstrap this function would, at best, collect a chain of 900, then a chain of 800, 700, and so on even though each instance would only use at most 100 tipsets of the returned chain as proposal.

Is that the most efficient we can be here? How fast/expensive is `GetParent`?

consensus_inputs.go
```diff
@@ -0,0 +1,258 @@
...
+// Returns inputs to the next GPBFT instance.
+// These are:
+// - the supplemental data.
+// - the EC chain to propose.
+// These will be used as input to a subsequent instance of the protocol.
+// The chain should be a suffix of the last chain notified to the host via
+// ReceiveDecision (or known to be final via some other channel).
+func (h *gpbftInputs) GetProposal(ctx context.Context, instance uint64) (_ *gpbft.SupplementalData, _ gpbft.ECChain, _err error) {
+	defer func(start time.Time) {
+		metrics.proposalFetchTime.Record(context.TODO(), time.Since(start).Seconds(), metric.WithAttributes(attrStatusFromErr(_err)))
```
masih, line 102:
```suggestion
		metrics.proposalFetchTime.Record(ctx, time.Since(start).Seconds(), metric.WithAttributes(attrStatusFromErr(_err)))
```

masih (approved):
No blockers 👍  Logic looks correct, tests pass.

Now that the committee and proposal provider is refactored, it would be nice to get the simulation tests to use the same logic instead if a total re-implementation of them.

consensus_inputs.go
Kubuxu: It is a pet peeve of mine not to introduce errors, especially in constructors, when they cannot happen.

consensus_inputs.go
Kubuxu: on expectation ~5x the hash size

consensus_inputs.go
Kubuxu: It isn't that expensive in comparison to getting the power table, it could be more efficient but it is tricky given null rounds.

consensus_inputs.go
Kubuxu: I agree that tests would be nice, but this is a small step; more refactoring might be required to remove the certstore from it.

PR #788, opened 2024-12-11
feat: add ChainProposedLength into the manifest

### LINKED ISSUE #772: Make maximum chain length configurable in manifest
F3 implementation has a hard limit on chain length: 100 tipsets. This limit is baked into the code in several places. This limit is also hit every time the network bootstraps for at least 9 times (due to 900 epoch lookback at bootstrap time).

One of the key factors affecting the bandwidth usage of F3, specially during catch up/bootstrap, is the size of the messages which is currently heavily influenced by the chain length. 

Ideally, we want to make the 100 length configurable in the manifest so that we can experiment with different configurations to see impact on bandwidth usage during catch up.

Kubuxu:
Resolves #772

### DIFF
--- a/gpbft/chain.go
+++ b/gpbft/chain.go
@@ -22,6 +22,9 @@ const (
 	CidMaxLen = 38
 	// ChainMaxLen specifies the maximum length of a chain value.
 	ChainMaxLen = 100
+
+	// ChainDefaultLen specifies the default length of chain value.
+	ChainDefaultLen = 100
 	// TipsetKeyMaxLen specifies the maximum length of a tipset. The max size is
 	// chosen such that it allows ample space for an impossibly-unlikely number of
 	// blocks in a tipset, while maintaining a practical limit to prevent abuse.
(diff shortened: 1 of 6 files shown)


gpbft/chain.go
```diff
@@ -22,6 +22,9 @@ const (
 	CidMaxLen = 38
 	// ChainMaxLen specifies the maximum length of a chain value.
 	ChainMaxLen = 100
+
+	// ChainDefaultLen specifies the default length of chain value.
+	ChainDefaultLen = 100
```
masih, line 27: Why introduce a new constant?

masih: Got it. Next question: Why the increase in max chain length from 100 to 128?

Kubuxu: done

Kubuxu: Because:
1. it aligns with the size of the merkle tree
2. if in future we want to increase it, the context was fresh in my head, so removing the techdebt of Max = Default was easy

But with intellectual honesty, it was first 1. and 2. is a bit more of post factum justification

masih: That makes sense to me. Thank you for documenting that in this comment. I updated the issue with it.

masih (approved):
We might as well increment the manifest protocol version. BUT later once other changes have also landed in manifest.

PR #791, opened 2024-12-12
Fix a bug in pubsub message ID hash function

masih:
The prior implementation I believe intended to include additional fields into account when calculating the message ID from data. But was only returning hash of data for both manifest and GPBFT messages.

Fix this by writing the data to the hasher, _then_ computing the hash with no suffix.

### DIFF
--- a/internal/psutil/psutil.go
+++ b/internal/psutil/psutil.go
@@ -28,8 +28,10 @@ func pubsubMsgIdHashData(m *pubsub_pb.Message) string {
 	if _, err := hasher.Write(topic); err != nil {
 		panic(err)
 	}
-
-	hash := blake2b.Sum256(m.Data)
+	if _, err := hasher.Write(m.Data); err != nil {
+		panic(err)
+	}
+	hash := hasher.Sum(nil)
 	return string(hash[:])
 }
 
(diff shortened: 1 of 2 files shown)


internal/psutil/psutil.go
```diff
@@ -53,8 +55,10 @@ func pubsubMsgIdHashDataAndSender(m *pubsub_pb.Message) string {
 	if _, err := hasher.Write(m.From); err != nil {
 		panic(err)
 	}
-
-	hash := blake2b.Sum256(m.Data)
+	if _, err := hasher.Write(m.Data); err != nil {
+		panic(err)
+	}
+	hash := hasher.Sum(nil)
```
Kubuxu, line 61: AFAIK `Sum` is a write followed by hash finalization. I'm going to check.

Kubuxu: Ohhh, it is not. It appends to it

Kubuxu (approved):

Kubuxu:
Great catch!!

Kubuxu:
One note is that this will break the message propagation.

masih:
So if we want to merge this it's best to fix the tag for the container used by manifest server and observer so that they do not pull the latest 👍

Stebalien (approved):
That's... wow.

PR #793, opened 2024-12-12
Implement GPBFT message compression using zstd

### LINKED ISSUE #786: Introduce GPBFT message compression
Based on the experiment performed as part of passive testing network `filecoin/46` compressed GPBFT messages are on average 60% smaller. This makes compression of messages a worthwhile exercise considering the small footprint of compression on resource consumption. See: https://github.com/filecoin-project/go-f3/issues/757#issuecomment-2535918089

Ideally, the compression introduced as part of this work should be backward compatible: 
* the implementation should continue to be able to interpret uncompressed messages in order to facilitate as much participation as possible since not all nodes will be upgraded to the latest.
* Eventually we can remove this backward compatibility entirely and only accept compressed messages (similar to the removal of dynamic manifest).

masih:
Add the ability to compress GPBFT messages controllable via manifest. Implement benchmarks to compare vanilla CBOR and ZSTD encoding.

Basic local run:
```
BenchmarkCborEncoding-12    	   47173	     25491 ns/op	  135409 B/op	      87 allocs/op
BenchmarkCborDecoding-12    	   64550	     18078 ns/op	   61728 B/op	     209 allocs/op
BenchmarkZstdEncoding-12    	   29061	     41489 ns/op	  193455 B/op	      88 allocs/op
BenchmarkZstdDecoding-12    	   66172	     17924 ns/op	  176517 B/op	     211 allocs/op
```

Fixes #786

### DIFF
--- a/msg_encoding.go
+++ b/msg_encoding.go
@@ -1,0 +1,75 @@
+package f3
+
+import (
+	"bytes"
+
+	"github.com/filecoin-project/go-f3/gpbft"
+	"github.com/klauspost/compress/zstd"
+)
+
+var (
+	_ gMessageEncoding = (*cborGMessageEncoding)(nil)
+	_ gMessageEncoding = (*zstdGMessageEncoding)(nil)
+)
+
+type gMessageEncoding interface {
+	Encode(*gpbft.GMessage) ([]byte, error)
+	Decode([]byte) (*gpbft.GMessage, error)
+}
+
+type cborGMessageEncoding struct{}
+
+func (c *cborGMessageEncoding) Encode(m *gpbft.GMessage) ([]byte, error) {
+	var buf bytes.Buffer
+	if err := m.MarshalCBOR(&buf); err != nil {
+		return nil, err
+	}
+	return buf.Bytes(), nil
+}
+
+func (c *cborGMessageEncoding) Decode(v []byte) (*gpbft.GMessage, error) {
+	r := bytes.NewReader(v)
+	var msg gpbft.GMessage
+	if err := msg.UnmarshalCBOR(r); err != nil {
+		return nil, err
+	}
+	return &msg, nil
+}
+
+type zstdGMessageEncoding struct {
+	cborEncoding cborGMessageEncoding
+	compressor   *zstd.Encoder
+	decompressor *zstd.Decoder
+}
+
+func newZstdGMessageEncoding() (*zstdGMessageEncoding, error) {
+	writer, err := zstd.NewWriter(nil)
+	if err != nil {
+		return nil, err
+	}
+	reader, err := zstd.NewReader(nil)
+	if err != nil {
+		return nil, err
+	}
+	return &zstdGMessageEncoding{
+		compressor:   writer,
+		decompressor: reader,
+	}, nil
+}
+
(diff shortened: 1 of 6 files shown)


Stebalien:
Lol, those decode benchmarks. Zstd is _fast_.

masih:
We can reduce allocation significantly by pooling the buffers too across both CBOR and ZSTD. but baby steps/small PRs.

msg_encoding_test.go
```diff
@@ -0,0 +1,156 @@
...
+func generateRandomTipSet(b *testing.B, rng *rand.Rand, epoch int64) gpbft.TipSet {
+	return gpbft.TipSet{
+		Epoch:      epoch,
+		Key:        generateRandomTipSetKey(b, rng),
+		PowerTable: generateRandomCID(b, rng),
+	}
+}
+
+func generateRandomTipSetKey(b *testing.B, rng *rand.Rand) gpbft.TipSetKey {
+	key := make([]byte, 20*38)
```
rvagg, line 146: 20? average tipset aims for 5 blocks

msg_encoding.go
```diff
@@ -0,0 +1,75 @@
...
+	cborEncoded, err := c.cborEncoding.Encode(m)
+	if err != nil {
+		return nil, err
+	}
+	compressed := c.compressor.EncodeAll(cborEncoded, make([]byte, 0, len(cborEncoded)))
+	return compressed, err
+}
+
+func (c *zstdGMessageEncoding) Decode(v []byte) (*gpbft.GMessage, error) {
+	cborEncoded, err := c.decompressor.DecodeAll(v, make([]byte, 0, len(v)))
```
rvagg, line 70: out of interest - have you tried variations of the `dst` argument here? `nil` or oversizing by a % given your knowledge of the likely compression ratio?

rvagg (approved):
neat

msg_encoding_test.go
masih: The number 20 comes from the max tipset key length constant in gpbft. I have randomised all fixed values in the generators to pick a size/length at random within configured max.

msg_encoding.go
masih: I have. My plan is to introduce pooling of buffers in a separate PR as it is an optimisation. There are a few places in encoding pipeline that we can benefit from it to reduce allocations.

In terms of impact, this probably doesn't matter as much considering the heap allocation numbers from passive testing. It is just wasteful.

msg_encoding_test.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+	})
+}
+
+func generateRandomGMessage(b *testing.B, rng *rand.Rand) *gpbft.GMessage {
+	return &gpbft.GMessage{
+		Sender:        gpbft.ActorID(rng.Uint64()),
+		Vote:          generateRandomPayload(b, rng),
+		Signature:     generateRandomBytes(b, rng, 96),
+		Ticket:        generateRandomBytes(b, rng, rng.Intn(96)+16),
```
Kubuxu, line 93: Ticket size is constant AFAIK

Kubuxu (approved):

msg_encoding_test.go
masih: Changed to fixed size bimodal presence at random (ignoring message validity for now since it doesn't matter for benchmarking)

PR #794, opened 2024-12-13
Publish containers tagged with version created by uci releaser

masih:
This repository started using unified CI (uci) a while back to make tagged releases. This release mechanism is triggered by editing the `version.json` which in turn triggers a CI workflow to make a tagged release along with ghithub release etc.

The issue with this approach is that because the tag is created by CI it does not trigger other CI jobs that listen to `tags: ['v*']` events. Therefore, since the switch we have only been publishing containers tagged with `main`.

The changes here resolve this issue by triggering the GHCR publish workflow whenever releaser wrokflow succeeds. It then fishes out the ref produced by the releaser workflow to use as tagging. This will then give us containers tagged by whatever release we make using uci releaser workflow.

Additionally, it adds `type=sha,format=long` tagging to containers for easier debugging and deployment of an explicit commit to `main`.

### DIFF
--- a/.github/workflows/publish-ghcr.yml
+++ b/.github/workflows/publish-ghcr.yml
@@ -6,11 +6,29 @@ on:
       - 'main'
     tags:
       - 'v*'
+    workflow_run:
+      workflows: [ Releaser ]
+      types:
+        - completed
   pull_request:
 
 jobs:
+  prepare-checkout:
+    if: github.event_name != 'workflow_run' || github.event.workflow_run.conclusion == 'success'
+    name: Prepare ref
+    runs-on: ubuntu-latest
+    outputs:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-ghcr.yml
```diff
@@ -6,18 +6,38 @@ on:
...
 
 jobs:
+  prepare-checkout:
+    if: github.event_name != 'workflow_run' || github.event.workflow_run.conclusion == 'success'
+    name: Prepare ref
+    runs-on: ubuntu-latest
+    outputs:
+      ref: ${{ github.event_name != 'workflow_run' && github.ref || steps.releaser.outputs.version }}
+    steps:
+      - name: Get Ref from releaser
```
rvagg, line 23:
```suggestion
      - name: Get version tag from releaser
```

rvagg (approved):

PR #798, opened 2024-12-13
Introduce sentinel errors for F3 not running and unknown manifest

### LINKED ISSUE #797: Introduce explicit error type for F3 disabled and no manifest
Additional to the conversation in https://github.com/filecoin-project/lotus/pull/12781#discussion_r1883601248, 
can I also ask that these two particular instances be turned into a publicly accessible error vars? The `"no known network manifest"` would be a good one too. Then we could do some easy checks and pass on our own typed errors in lotus for these common cases.

_Originally posted by @rvagg in https://github.com/filecoin-project/go-f3/issues/795#issuecomment-2541104898_

masih:
For better upstream logic flow introduce sentinel errors to signal:
* when F3 is not running,
* when manifest is not known.

Fixes #797

### DIFF
--- a/f3.go
+++ b/f3.go
@@ -2,6 +2,7 @@ package f3
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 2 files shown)


rvagg (approved):
👌

PR #805, opened 2024-12-18
Add quality duration multiplier

Kubuxu:


gpbft/options.go
```diff
@@ -82,6 +85,16 @@ func WithDeltaBackOffExponent(e float64) Option {
 	}
 }
 
+func WithQualityDurationMultiplier(m float64) Option {
+	return func(o *options) error {
+		if m < 0 {
```
masih, line 90: Probably zero is also an invalid value?

manifest/manifest.go
```diff
@@ -98,9 +99,10 @@ func (c *CxConfig) Validate() error {
 }
 
 type GpbftConfig struct {
-	Delta                time.Duration
-	DeltaBackOffExponent float64
-	MaxLookaheadRounds   uint64
+	Delta                     time.Duration
+	DeltaBackOffExponent      float64
+	QualityDurationMultiplier float64
```
masih, line 104: `QualityDeltaMultiplier` and rename across the code? 

The rationale being: the duration of quality is not exactly governed by that multiplier. It will finish on strong quorum of proposal. Calling it a Delta multiplier is clearer because it is that.

masih (approved):
Comment on terminology, otherwise LGTM.

Issue #792, opened 2024-12-12
Hash common fields for propagation


Kubuxu:
The most significant offender here is ECChain, which is the same (or similar) for many messages and can end up large.

The idea is to hash the ECChain before sending the message and let the receiver resolve it from the hash.
In many scenarios, we will know which ECChain corresponds to the hash because we will have the same one, but for the less common case, we need to propagate these ECChains through pubsub. This can be done on a separate topic (or maybe even on the same topic).

That propagation channel should limit the propagations to only the current instance, so the ECChain message should carry the instance number.

There are a couple of architectural questions:
1. Should GPBFT be aware of it, or should it continue to operate on GMessage as it is today?
2. What should we do with messages that ECChain hashes we cannot resolve?
3. How do we limit the ECChain lookup buffer?
4. Do we need a separate topic, or is it better to use the same one?

Notes:
After receiving an ECChain, we should add all prefix chains to the lookup buffer. This provides more information and improves efficiency. We should also preload the buffer with our own ECChain.

masih:
Summary of discussions in standup: 

* Because the chains advertised in QUALITY phase are the superset of values that could ever be decided in an instance, we could limit publication of ec chains to QUALITY proposals only.
* Each node can then calculate the hash of every prefix of those proposals to then infer other hashes that the instance may be swayed towards. 
* This then means we do not need to change the chain that is being broadcased on the ec chain topic relative to the gpbft progress.

Stebalien:
> what is your take on question No. 4? Do you think it's better to separate the topic or use the same one with a message ID function that's smart enough?

IMO, use a separate topic. Having to decode messages in the message ID function is, IMO, a bad idea in general. But I don't feel strongly about this.

> Should GPBFT be aware of it, or should it continue to operate on GMessage as it is today?

Don't we need quality information to support the ECChains? Otherwise, an attacker can just spam EC Chains and force everyone to store every possible EC chain because they'll have no idea which ones are actually useful.

That is, I _think_ what we need is an extra phase after quality.

Kubuxu:
> Don't we need quality information to support the ECChains? Otherwise, an attacker can just spam EC Chains and force everyone to store every possible EC chain because they'll have no idea which ones are actually useful.

We can limit it and discard chains aggressively after quality.

Stebalien:
I don't really see how to do this securely (e.g., a flood can easily mask all real values). Especially before/during quality.

Are we relying on some form of continuous rebroadcast?

[...]

PR #807, opened 2024-12-19
Implement Chain Exchange protocol over pubsub

masih:
Implement chain exchange protocol over pubsub as a mechanism to propagate `ECChain` across the network with reasonable spam protection.

To protect against spam the mechanism employs two separate caches for chains that are generally discovered across the network and the ones explicitly looked up or broadcasted by the local node. Both caches are capped LRU, where the LRU recent-ness is used as a way to prioritise chains we cache while keeping the total memory footprint fixed. This approach is not the most memory efficient but is simpler to implement as the LRU encapsulates a lot of the complexity.

The code has a lot of TODOs as places to improve or question to the reviewer. To action most of the TODOs further refactoring across the code is needed which is intended to be actioned in separate commits.

The code path introduced here is not integrated into F3 host; future PRs will iteratively integrate the mechanism across F3 host and other places.

Part of #792


chainexchange/pubsub.go
```diff
@@ -0,0 +1,306 @@
...
+	_ ChainExchange      = (*PubSubChainExchange)(nil)
+	_ pubsub.ValidatorEx = (*PubSubChainExchange)(nil).validatePubSubMessage
+
+	chainPortionPlaceHolder = &chainPortion{}
+)
+
+type chainPortion struct {
+	// chain is a pointer to the full chain in order to avoid copying the chain across the cache
+	fullChain *gpbft.ECChain
+	offset    int
```
Kubuxu, line 30: ECChain is a slice, so there is no reason to make it a pointer (I guess it saves a few bytes, but I don't think a few bytes will break us), as the slice is a tuple of `[array_pointer, capacity, length]`
You could also re-slice the fullChain instead of saving offset.

chainexchange/chainexchange.go
```diff
@@ -0,0 +1,27 @@
+package chainexchange
+
+import (
+	"context"
+
+	"github.com/filecoin-project/go-f3/gpbft"
+)
+
+type Key []byte
```
Kubuxu, line 9: maybe make it a string so it can be used in caches and for comparisons?

Kubuxu (approved):
Looks good, some of these TODO-s should get done, but they don't have to get done right now.

chainexchange/chainexchange.go
masih: The reason I haven't is because of avoiding invalid strings once this gets encoded as CBOR (in upcoming partial message that includes it).

PR #808, opened 2024-12-20
Add the ability to listen for discovered chains

masih:
Expand chain exchange to accept a listener which is notified whenever a new chain is discovered. This mechanism is intended to be integrated into F3 host pubsub, whereupon receiving a partial message the host looks up its chain. When known, the chain is returned immediately. Otherwise, the host would buffer the partial message and await notification of its discovering from chain exchange.

Part of #792

### DIFF
--- a/chainexchange/pubsub.go
+++ b/chainexchange/pubsub.go
@@ -107,7 +107,7 @@ func (p *PubSubChainExchange) Key(chain gpbft.ECChain) Key {
 	return rootDigest[:]
 }
 
-func (p *PubSubChainExchange) GetChainByInstance(_ context.Context, instance uint64, key Key) (gpbft.ECChain, bool) {
+func (p *PubSubChainExchange) GetChainByInstance(ctx context.Context, instance uint64, key Key) (gpbft.ECChain, bool) {
 
 	// We do not have to take instance as input, and instead we can just search
 	// through all the instance as they are not expected to be more than 10. The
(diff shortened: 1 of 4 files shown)


chainexchange/pubsub.go
```diff
@@ -270,6 +274,9 @@ func (p *PubSubChainExchange) cacheAsWantedChain(ctx context.Context, cmsg Messa
 				fullChain: &cmsg.Chain,
 				offset:    offset,
 			})
+			if p.listener != nil {
+				p.listener.NotifyChainDiscovered(ctx, key, cmsg.Instance, prefix)
```
Kubuxu, line 278: IDK if the notification under lock is the best idea. It might be better to move it outside the lock.

Kubuxu:
If you are saying we are fine with doing it under lock, we should keep it under lock. Manual interrupted locking like this is a foot gun.

masih:
> Manual interrupted locking like this is a foot gun.

I take this as a blocker? It's either this or notify listener under lock. I have no preference.

Taking a closer look at the lru library we are using, I am sort of underwhelmed that it doesn't expose atomic combined removal operations. if we had those, we could remove the need for lock in most places. I also understand why they haven't , i.e. the auto eviction of lru used for caching.

Kubuxu:
> I take this as a blocker? It's either this or notify listener under lock. I have no preference.

I wouldn't make it a blocker, it is just pattern I try to avoid.

Kubuxu (approved):

masih:
Capturing what discussed in the standup: this code is going to get refactored a whole bunch before it is shipped. Merging for now and will revisit locking pattern in follow up PRs. Thanks

PR #816, opened 2025-01-09
Integrate end to end hashed chain exchange with F3

masih:
Integrate the chain exchange mechanism with F3 host and runner. But without touching the core GPBFT.

The implementation here leaves two major TODOs: 1) chain broadcasting mechanism (currently coupled to GPBFT message broadcast), and 2) partial message validation prior to buffering (currently skipped entirely but with capped buffer sizes and re-validation by core GPBFT once the messages are complete).

The integration introduces the concept of Partial GMessage: a GMessage with chains replaced with the key to the chain. The work introduces a buffer and refill mechanism that listens to the chains discovered, un-buffers the messages having re-constructed their original GMessage and feeds them to the participation using the existing event loop.

Part of #792


Kubuxu:
Conceptually SGTM, will do another dive latter after I think it over.

Kubuxu (approved):

PR #820, opened 2025-01-13
Reduce exit sleep in tests run with race detector to zero

masih:
Reduce exit sleep in tests run with race detector to zero to make the tests run faster.


masih:
Looking at the CI run, this shaved off 1 minute from `go test --race ./...` compared to `main`.

Kubuxu (approved):

PR #830, opened 2025-01-16
Fix potential panic when subscribing to topic fails

masih:
Avoid referencing subscription when constructing an error message as it may be nil when subscription fails.


Kubuxu (approved):
good catch

PR #834, opened 2025-01-20
Consistently use Merkle Tree root as ECChain key

### LINKED ISSUE #825: Use lazily loaded Merkle Tree root as EC Chain Key
The current implementation uses two approaches for generating key for a given EC Chain:

 1) `ChainKey` construction by [hashing each of the tipsets](https://github.com/filecoin-project/go-f3/blob/1b17c6c40971e9ff408afa8b8cd7fb90f79e587c/gpbft/chain.go#L369-L385)
2) Merkle Tree root hash [used in signing the messages](https://github.com/filecoin-project/go-f3/blob/1b17c6c40971e9ff408afa8b8cd7fb90f79e587c/gpbft/gpbft.go#L156-L160).

There are historical reasons for this: e.g. the first approach predates the second. But there is no technical reason for having two separate approaches. 

In terms of computational expense, the first approach is cheaper than the second one. But in practice:
* we have to calculate the root Merkle tree hash once for signing purposes.
* the code repeatedly generates chain key using the first approach.

This means, if merkle tree root is generated once and cached as part of ECChain struct the computational expense becomes a non-issue if not cheaper.

Additionally, this approach would benefit #792, because chainexchange no longer needs to own the concept of generating keys for chains. We can just call `ECChain.Key` and use the same key across the entire implementation.

On approach, the idea is to use `sync.Once` and cache the merkle tree root as key lazily.

masih:
Remove the custom ECChain key generation inside `gpbft` in favor of consistently using Merkle Tree root of chain as key. This key is already used for signature payload generation, and in the new chain exchange as a key to identify a chain. Doing so enables signature validation of messages without having to know the chain.

Consistent use of Merkle Tree root hash as key enables a number of simplifications across the repo:
* unblocks the unification of validation logic between partial and full validators, since caching can use a consistent way to identify messages whether they're partial or not.
* reduce various type gymnastics across root and chain exchange packages by repurposing gpbft ECChain Key.

The work here also introduces lazy-loading for the ECChain, which requires ECChain type to be converted to `struct`, and be passed by pointer. As part of this work, TipSet is also turned into a pointer consistently across various interfaces.

The ECChain receiver functions are adopted to accommodate potential nil values for a chain.

Fixes #825


gpbft/chain.go
```diff
@@ -176,112 +177,144 @@ func tipSetKeyFromCids(cids []cid.Cid) (TipSetKey, error) {
...
 
-// Compares two ECChains for equality.
-func (c ECChain) Eq(other ECChain) bool {
-	if len(c) != len(other) {
+// Eq Compares two ECChains for equality.
+// Note that two zero chains are considered equal. See IsZero.
+func (c *ECChain) Eq(other *ECChain) bool {
+	if oneZero, otherZero := c.IsZero(), other.IsZero(); oneZero || otherZero {
+		return oneZero == otherZero
+	}
```
Kubuxu, line 312: another way to write it:
if c.IsZero() && c.IsZero() == other.IsZero() {
     return true
}
but IDK which one is cleaner

masih: I think the problem with writing it as `if c.IsZero() && c.IsZero() == other.IsZero() {` is that if `c` is not zero but `other` is zero the equality check progresses to the next set of checks. We don't want that because, "zero" also could mean a nil `ECChain` and that could cause a panic. Right?

Kubuxu: aah, you are right

Kubuxu:
I have a slight concern, but it is more general. How do we transition Calibnet to it, given that it lives and breathes the older formats?

masih:
> I have a slight concern, but it is more general. How do we transition Calibnet to it, given that it lives and breathes the older formats?

Good question. If I have not missed anything I think the only wire format change is the `ECChain` CBOR encoding, where it gets encoded as a struct instead of the old array of `TipSet`s. 

To keep compatibility, I can override the CBOR marshalling for it to continue to write in old format since that's all we really need. I actually did this at first then I thought we don't need it having missed the fact that calibnet is live. 

That should be enough right? Is there anything else you can think of that I might have missed?

Kubuxu (approved):
SGTM

Kubuxu:
> That should be enough right? Is there anything else you can think of that I might have missed?

I think the ECChain is the only thing with persistent data. There is a question of how do we switch to chain exchange scheme but I think allowing it to go down for a bit and catch back up is good enough solution for calibnet.

masih:
OK I have added compatibility tests with the old format and handcrafted the CBOR decoding for `ECChain`.

@Kubuxu can I ask you to take a look at [3e5b007](https://github.com/filecoin-project/go-f3/pull/834/commits/3e5b007dfcf1a9e4cb9f63dd89d4d5c905b52298) before I merge this PR. It's a sensitive part of the code and I'd like another pair of eyes on it please. 🙏

masih:
I'll hold off resolving conflicts after Kuba had a chance to review the added CBOR encoding.

Kubuxu (approved):
SGTM, the copy is a bit annoying by oh well

PR #840, opened 2025-01-21
Add a periodic chain re-broadcast mechanism

### LINKED ISSUE #814: Implement a chain rebroadcast mechanism independent of GPBFT
As part of #792 the propagation of EC Chain is separated into its own pubsub topic, managed via `chainexchange`. Initially, the broadcast of chain and its reboradcast is coupled to GPBFT broadcast/re-boradcast: Whenever GPBFT broadcasts a message we broadcast its chain and so on. But we can do better, because: all chain subsets can be inferred from the longest broadcasted chain (i.e. the QUALITY proposal), and depending on the progress it may not be necessary to rebroadcast the chain.

Ideally, the system should have control over the chain broadcast intensity. Separate out the heuristic of chain broadcast from GPBFT.

Ideas:
* Only broadcast a chain a fixed number of times.
* Introduce an incrementing sequence number in messages to work around gossipsub deduplication.
* For safety rebroadcast subsets of the QUALITY chain relative to progress? 
   * Or just stick with always broadcasting the QUALITY proposal. Because from it we can infer all the sub chains that could be stemmed from it during an instance, and it is simpler to implement/maintain.

masih:
Based on a configured rebroadcast interval periodically rebroadcast the longest chain currently being progressed by GPBFT.

Extend the chain exchange broadcast message to include a timestamp rounded down to the rebroadcast interval as a way to work around gossipsub deduplication. This is to help gain better message propagation.

The rebroadcast interval is reset as soon as the first message with a larger instance is observed, signaling the start of the next instance. This helps synchronize the chain rebroadcast intervals relative to the GPBFT instance start.

Fixes #814

### DIFF
--- a/partial_msg.go
+++ b/partial_msg.go
@@ -3,6 +3,7 @@ package f3
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/filecoin-project/go-f3/chainexchange"
 	"github.com/filecoin-project/go-f3/gpbft"
@@ -43,6 +44,10 @@ type partialMessageManager struct {
 	// pendingDiscoveredChains is a channel of chains discovered by chainexchange
 	// that are pending to be processed.
 	pendingDiscoveredChains chan *discoveredChain
+	// pendingChainBroadcasts is a channel of chains that are pending to be broadcasted.
+	pendingChainBroadcasts chan chainexchange.Message
+	// rebroadcastInterval is the interval at which chains are re-broadcasted.
+	rebroadcastInterval time.Duration
 
 	stop func()
 }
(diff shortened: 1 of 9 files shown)


partial_msg.go
```diff
@@ -147,22 +155,82 @@ func (pmm *partialMessageManager) Start(ctx context.Context) (<-chan *PartiallyV
...
+				case current == nil, pending.Instance > current.Instance:
+					// Either there's no prior chain broadcast or a new instance has started.
+					// Broadcast immediately and reset the timer to tick from now onwards. This is to
+					// re-align the chain rebroadcast relative to instance start.
+					current = &pending
+					current.Timestamp = roundDownToUnixTime(time.Now(), pmm.rebroadcastInterval)
+					if err := pmm.chainex.Broadcast(ctx, *current); err != nil {
+						log.Errorw("Failed to immediately re-broadcast chain.", "instance", current.Instance, "chain", current.Chain, "error", err)
+					}
+					ticker.Reset(pmm.rebroadcastInterval)
```
Kubuxu, line 194: Should we set the ticker to the rounded down time + interval instead, to line it up?
I don't think it matters; it just came up while I was reviewing.

Kubuxu (approved):

partial_msg.go
masih: Yup that makes sense. Thanks Kuba

masih: On second thought I think this makes sense for the first firing of ticker but after that we want to use the interval. Right?

This makes the code complicated enough such that it puts me off from doing it. With the default interval of 2 seconds (and most likely sub 10s in any case?) I think it's not worth it. 

Going to merge as is. Please let me know if you think otherwise and I will adjust in a follow up PR 🙏

PR #842, opened 2025-01-22
Integrate zstd compression into chain exchange

### LINKED ISSUE #819: Integrate compression into chainexchange topic
Add compression to messages published on chainexchange topic.

masih:
The GPBFT message exchange over pubsub already uses zstd compression on top of CBOR encoded messages. The work here integrates the same style of compression for chain exchange messages, with additional unification of the encoding mechanism across the two.

The work refactors the root level encoding implementation into a generic encoder decoder that both chain exchange and gpbft used. Tests and benchmarks are updated to reflect this.

The benchmarking of partial gmessage encoding is also adjusted to fix a few redundant statements and bugs in testing.

Fixes #819  #843


Stebalien:
Can we add a compression bomb test? I think we're fine, but it would be nice to have a test.

- The fact that we're stream decoding into the CBOR decoder coupled with the fact that the CBOR decoder has a bunch of limits should protect us against memory issues.
- The fact that our CBOR decoder won't read through unlimited input (e.g., doesn't have logic that skips unknown fields, etc.) means we can't get "stuck" when decoding. But it would be _nice_ if we could wrap the decompressed reader in a limited reader before passing it off to the CBOR decoder, just in case. That way we can guarantee that we'll process at most N bytes of CBOR before giving up.

chainexchange/pubsub.go
```diff
@@ -38,18 +38,24 @@ type PubSubChainExchange struct {
...
 	stop                 func() error
+	encoding             *encoding.ZSTD[*Message]
 }
 
 func NewPubSubChainExchange(o ...Option) (*PubSubChainExchange, error) {
 	opts, err := newOptions(o...)
 	if err != nil {
 		return nil, err
 	}
+	zstd, err := encoding.NewZSTD[*Message]()
```
Kubuxu, line 49: Are we going with ZSTD by default?

masih: For chain exchange yes. For GPBFT it's configurable via manifest.

Happy to make it configurable for chain exchange too if you think it's worth doing.

masih:
> Can we add a compression bomb test?

Sure. Captured https://github.com/filecoin-project/go-f3/issues/843

masih:
@Stebalien the message encoding decoding off Gossipsub does not function in streaming manner. Instead, both the publisher and subscriber hand over `[]byte` slices to the pubsub subsystem to do its thing.

The default maximum message size in pubsub is set to 1 MiB, which as far as I can tell has not been overridden in Lotus. This limit is large enough for the purposes of F3. So, what I have done in https://github.com/filecoin-project/go-f3/pull/842/commits/05bc5784bfe1763eec885dd994148c7979b21890 is to explicitly set the maximum decoded value in zstd to 1 MiB. For sanity, I have also restricted encoder to refuse to encode values that would hit that limit.

Does this cover your concern re expansion attack vector in the context of zstd compression?

Stebalien:
That's good but isn't the _zstd_ coder streaming? Looking at the docs in:

https://pkg.go.dev/github.com/klauspost/compress/zstd#WithDecoderMaxMemory

That restricts the max memory held at any point in time while streaming, not the max that can be read from the stream in total (unless I'm misreading it).

masih:
> That's good but isn't the zstd coder streaming?  

It has two modes: streaming where one needs to allocate a decoder per io.reader, or nonstreaming where one declares a reader with nil io.reader and uses `DecodeAll`. The latter is what the implementation in this repo does. The rationale for this choice was reduced GC.

> (unless I'm misreading it).

The documentation reads: "WithDecoderMaxMemory allows to set a maximum decoded size **for in-memory non-streaming operations** _or_ maximum window size for streaming operations. This can be used to control memory usage of potentially hostile content. ".

Right?

masih:
@Kubuxu @Stebalien Can I get a review on this PR please?

internal/encoding/encoding.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+	if err != nil {
+		return nil, err
+	}
+	compressed := c.compressor.EncodeAll(cborEncoded, make([]byte, 0, len(cborEncoded)))
+	return compressed, nil
+}
+
+func (c *ZSTD[T]) Decode(v []byte, t T) error {
+	cborEncoded, err := c.decompressor.DecodeAll(v, make([]byte, 0, len(v)))
```
Stebalien, line 81: Future change: we should use a buffer pool for these short-lived buffers (https://pkg.go.dev/sync#Pool). If we do that, we can also allocate these buffers with 1MiB capacities and use `WithDecodeAllCapLimit` to avoid any allocations.

Stebalien (approved):
LGTM!

internal/encoding/encoding.go
masih: That's [the plan](https://github.com/filecoin-project/go-f3/pull/793#discussion_r1883631902). Thank you for reminding me. Captured https://github.com/filecoin-project/go-f3/issues/849

PR #844, opened 2025-01-23
Disable docker build summary generation and artifact upload

masih:
The build summary generator is taking a very long time currently. This is most likely due to upstream issues with pulling necessary docker images.

The capability itself is mostly cosmetic, i.e., we don't care about the summary nor artifacts uploaded to GitHub Actions as long as the container itself is pushed. Hence, the decision to disable them for faster build that does just enough.

Example of post-build summary generation taking too long:
 * https://github.com/filecoin-project/go-f3/actions/runs/12926848944/job/36050762119


masih:
Turns out GitHub Actions is in a "toasted" [state](https://www.githubstatus.com/).

Issue #849, opened 2025-01-27
Add pooling to buffers used for message encoding


masih:
See:
* https://github.com/filecoin-project/go-f3/pull/793#discussion_r1883631902
* https://github.com/filecoin-project/go-f3/pull/842#discussion_r1929085482

PR #854, opened 2025-01-27
Pool decode buffers when encoding messages using compression

masih:
Pool the decode buffers and set strict max cap allocation for zstd decompressor.

See Benchamark results below.

Before:

```
BenchmarkCborEncoding
BenchmarkCborEncoding-12    	  467352	      2376 ns/op	   14680 B/op	      10 allocs/op
BenchmarkCborDecoding
BenchmarkCborDecoding-12    	  104410	     11347 ns/op	   14944 B/op	      27 allocs/op
BenchmarkZstdEncoding
BenchmarkZstdEncoding-12    	  286735	      3897 ns/op	   46748 B/op	      12 allocs/op
BenchmarkZstdDecoding
BenchmarkZstdDecoding-12    	  110794	     10783 ns/op	   28512 B/op	      28 allocs/op
```

After:

```
BenchmarkCborEncoding
BenchmarkCborEncoding-12    	  436754	      2383 ns/op	   14680 B/op	      10 allocs/op
BenchmarkCborDecoding
BenchmarkCborDecoding-12    	  106809	     11280 ns/op	   14944 B/op	      27 allocs/op
BenchmarkZstdEncoding
BenchmarkZstdEncoding-12    	  294043	      3918 ns/op	   46746 B/op	      12 allocs/op
BenchmarkZstdDecoding
BenchmarkZstdDecoding-12    	  114854	     10747 ns/op	   18314 B/op	      27 allocs/op
```

Fixes: #849

### DIFF
--- a/encoding_bench_test.go
+++ b/encoding_bench_test.go
@@ -43,7 +43,7 @@ func BenchmarkCborDecoding(b *testing.B) {
 		for pb.Next() {
 			var got PartialGMessage
 			require.NoError(b, encoder.Decode(data, &got))
-			require.Equal(b, msg, &got)
+			requireEqualPartialMessages(b, msg, &got)
 		}
 	})
 }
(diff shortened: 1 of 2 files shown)


rvagg:
> consistent dependency with the one used in Lotus

Not really. https://github.com/filecoin-project/lotus/pull/12015 kind of relevant, there's some pressure to not introduce more cgo lock-in if we can help it, particularly for downstream dependents of those components. I'd say unless the perf improvement here is really dramatic maybe let's just stuck with klauspost/compress.

masih:
> > consistent dependency with the one used in Lotus
> 
> Not really.

Consistent dependency to the same libraries for the same functionality IMO is a desirable. Maybe the answer here is to move Lotus off the DataDog dependency. I know that `klauspost/compress` zstd package has come a long way since its early days in terms of performance.

Maybe the answer is to move Lotus off DataDog dependency?

masih:
> I'm surprised we saved an allocation on encode and nothing on decode (where we added the pool). What happens if we remove the pool? Does the datadog version just have an additional allocation internally?

pooling makes no difference to the encoding part. The difference in allocation comes from the internals of the dependency libraries. Here are the numbers for `DataDog/zstd` _without_ pooling:

```
BenchmarkCborEncoding
BenchmarkCborEncoding-12    	  436611	      2375 ns/op	   14680 B/op	      10 allocs/op
BenchmarkCborDecoding
BenchmarkCborDecoding-12    	  106036	     11302 ns/op	   14944 B/op	      27 allocs/op
BenchmarkZstdEncoding
BenchmarkZstdEncoding-12    	  258430	      4643 ns/op	   28248 B/op	      11 allocs/op
BenchmarkZstdDecoding
BenchmarkZstdDecoding-12    	   27158	     43543 ns/op	 1063526 B/op	      28 allocs/op
```

masih:
To unblock work I am going to separate pooling from dependency change in this repo until we figure out https://github.com/filecoin-project/lotus/issues/12852.

Issue #846, opened 2025-01-24
Investigate docker build issue on linux/arm64 platform


masih:
Initially observed in #834 but it is unrelated to the changes in that PR: confirmed by re-running the docker build on latest commit of `main` [here](https://github.com/filecoin-project/go-f3/actions/runs/12926848944/job/36125151077). The issue:

```
> [linux/arm64 build 6/6] RUN go build -o /go/bin/f3 ./cmd/f3:
87.02 runtime/cgo: gcc: signal: segmentation fault (core dumped)
------
WARNING: No output specified with docker-container driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load
Dockerfile:9
--------------------
   7 |     
   8 |     COPY . .
   9 | >>> RUN go build -o /go/bin/f3 ./cmd/f3
  10 |     
  11 |     FROM gcr.io/distroless/cc
--------------------
ERROR: failed to solve: process "/dev/.buildkit_qemu_emulator /bin/sh -c go build -o /go/bin/f3 ./cmd/f3" did not complete successfully: exit code: 1
```

**Why has it started failing now?** 

Because go-f3 container build is not reproducible:
* the dockerfile uses `latest` tags for the build and base images.
* the GitHub actions use a soft version for the build setup.

Change at any of the above layers could have caused the issue, e.g. gcc linker upgrade.
Also, yesterday GitHub Actions was having a bad day with a flurry of errors specifically in docker build jobs which _may_ be related.

Building multi-platform containers locally (using `podman`) passes for me with no issues.

Unblocking the merge of that PR by capturing this issue to investigate the problem separately.

FacuM:
IMO it looks like the VM is running out of RAM and causing some weird behavior. It looks like I've found a workaround, as I've commented in my issue:

```yml
jobs:

  build:
    runs-on: ubuntu-22.04
```

Please try like that and let me know if it helps. :)

masih:
Thank you @FacuM, I'll give that a shot 👍

masih:
@FacuM In our case using the ubuntu-22.04 runner image made no difference. See: https://github.com/filecoin-project/go-f3/actions/runs/13007480659/job/36277473006?pr=855

masih:
With QEMU setup it still fails: https://github.com/filecoin-project/go-f3/actions/runs/13013844154/job/36297943821

masih:
Update using qemu v8 works.

PR #857, opened 2025-01-28
Set up QEMU prior to image build

masih:
Investigating to see if setting up QEMU would resolve the docker build issues.

Fixes https://github.com/filecoin-project/go-f3/issues/846


masih:
OK looks like building with qemu v8 works. Just re-ran the job to make sure the first success wasn't a fluke.

masih:
OK ✅ Fixes #846

rjan90 (approved):

PR #862, opened 2025-01-30
Rework time dilution in F3 tests and refine assertions

### LINKED ISSUE #818: Review and rewrite more explicit assertion logic for `f3_test.go`
The `f3_test.go` tests the upper level F3 functionality (i.e. including pubub sub-systems, manifest propagation, etc.). It is a valuable suit of tests that have caught a number of issues in the past. 

These tests, however, loosely rely on a fake clock: The EC clock is fake but that does not offer control over the progress of GPBFT itself. This is _OK_ but it means the test should be more explicit about asserting things. For example, it is better to assert that nodes have reached some instance _and a specific phase/round_.  As a result, it has become time consuming to debug issues when these tests fail.

There are adhoc sleeps in a number of places that should ideally be removed once a more precise eventual assertion is made about the state of each node in the test environment.

### LINKED ISSUE #847: `TestF3LateBootstrap` is flaky
Over the course of latest development in this repo `TestF3LateBootstrap` is pupped up a number of times as being flaky. It's timeout has already been bumped to 1m, and it is the slowest test we have in this repo.

One cause that I have observed is that phases time out before messages are delivered which results in repeated base decision and causes assertions that check if a certain epoch is finalised take too long. A root cause is the way mock clock is used in the tests: it is used to grow the chain and gpbft. Another issue is the way clock advancing works relative to assertion checks. As is the test is very CPU hungry as it runs assertions in a hot-loop.

I captured this as a separate issue from #818 to emphasise the need for making this test run faster as it slows down development.

masih:
Progress the clock during tests independently of condition check. Rework the assertions to use built-in eventual assertion. Use consistent names for the asserting functions.

Fixes #818
Fixes #847


masih:
Hold off on reviews until I crack the intermittent failures on CI. Regardless, early feedback is always welcomed 🙏

masih:
@Kubuxu  this is ready for review whenever you get a chance. Thanks

Kubuxu (approved):

PR #882, opened 2025-02-07
Use the latest computed power table instead of always ranging certs

### LINKED ISSUE #881: Cache the latest known power table in `certstore`
A way to significantly speed up `certstore.GetPowerTable` (in exchange for a small memory footprint) is to cache the latest known powertable in `certStore`. This would make calls to `GetPowerTable` for an instance made by [consensus inputs](https://github.com/filecoin-project/go-f3/blob/ded3d042b8f21ccb9e18bff032682fafe8fbc58d/consensus_inputs.go#L221) in steady state very quick. This call makes up the vast majority of `certstore` workload.

See: [#874](https://github.com/filecoin-project/go-f3/issues/874#issuecomment-2639449429)

masih:
Certificate store already keeps track of the latest power table, i.e., the power table of the next expected certificate. Use this pre-computed table whenever the next power table is asked for instead of re-computing it every time.

This results in a much faster response to `GetPowerTable` for the most common case: getting the committee as part of monotonic progress of GPBFT instances.

Fixes #881

### DIFF
--- a/certstore/certstore.go
+++ b/certstore/certstore.go
@@ -287,13 +287,31 @@ func (cs *Store) GetPowerTable(ctx context.Context, instance uint64) (gpbft.Powe
 		return nil, fmt.Errorf("cannot return a power table before the first instance: %d", cs.firstInstance)
 	}
 
-	lastPowerTable := cs.firstInstance
-	if latestCert := cs.Latest(); latestCert != nil {
-		lastPowerTable = latestCert.GPBFTInstance + 1
+	// Copy a reference to both the latest cert and latest power table while holding
+	// the lock to guarantee order in case the latest changes while the requested
+	// power table is being retrieved.
+	cs.mu.RLock()
+	latestCert := cs.latestCertificate
+	latestPowerTable := cs.latestPowerTable
+	cs.mu.RUnlock()
+
+	nextCertInstance := cs.firstInstance
+	if latestCert != nil {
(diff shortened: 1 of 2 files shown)


masih:
Has been running on butterflynet since ~1200 UTC. Get committee times looking much better:
image

certstore/certstore.go
```diff
@@ -287,13 +287,31 @@ func (cs *Store) GetPowerTable(ctx context.Context, instance uint64) (gpbft.Powe
...
+	cs.mu.RUnlock()
+
+	nextCertInstance := cs.firstInstance
+	if latestCert != nil {
+		nextCertInstance = latestCert.GPBFTInstance + 1
+	}
+	if instance > nextCertInstance {
+		return nil, fmt.Errorf("cannot return future power table for instance %d > %d", instance, nextCertInstance)
+	}
+	if instance == nextCertInstance && len(latestPowerTable) != 0 {
```
Stebalien, line 305: Instead of checking `len(latestPowerTable)`, can we check `latestCert != nil`? I guess that won't work for instance 0... meh, this works fine.

Stebalien (approved):

certstore/certstore.go
masih: Yes, it would be problematic for `firstInstance`. An alternative is to explicitly check if the `instance` is the same as `firstInstance`. But that makes up a more complicated logic.

Issue #893, opened 2025-03-03
Run Chaos experiments: Node churn of 33%


masih:
Run a chaos experiment that churns 1/3 of network at random and expect the f3 progress to continue.

masih:
At 33% sudden drop in participation the progress stops. This is across a set of peers all with roughly enqual power.

After making sure all participants are connected, the progress did not resume. 

Repeating the experiment but this time with 20% pod failure.

BigLep:
2025-03-05 standup conversation: 
* the worrisome part is "After making sure all participants are connected, the progress did not resume."
* we think deploying the observer will help a lot

masih:
After reconnecting the nodes and waiting for about 1 hour, F3 resumed to progressing which is great.

I will reset the network and repeat this experiment with failure of 20%.

masih:
At around 0300 UTC F3 progress on a freshly reset stopped. Average connectivity across nodes is in mid 80 peers for a network of 100. This is without any manual connections, and with 1 bootstrap node with pubsub peer exchange enabled.

Investigating why progress stopped before proceeding with introduced chaos.

BigLep:
2025-03-06 verbal conversation on the stalled progress:
Droppage happens in https://github.com/filecoin-project/go-f3/blob/3a9aeca98cabe1f0ca65e2acf6506dc7015e469b/partial_msg.go#L133 per log messge
Suggestion to drop earliest message rather than the latest message
TODO to parameterize the queue
Discussion of 2k network using block times of 4s and manifest uses 30s
(I had to drop)

masih:
Continued discussions:
* I also observe Tag Tracer logs about queue being full. This is deep from the guts of pubsub, a type responsible for message delivery in conjunction with per-peer connection manager.

masih:
During the investigation two issues were identified:
1. the self equivocation filter is not applied to the chain broadcast made when proposal for an instance is fetched. This meant that if a node is stopped mid participation then when it starts back up it would advertise the longest chain it can again for the same instance. Fixed in #899 by checking for existence of any prior messages from self for an instance before broadcasting chain.
2. The nodes used very long backoff in broadcast/rebroadcast to the point where for hours they would not publish any messages. The root cause of this is that the base timestamp for an instance is determined based on the epoch of the latest finalised tipset. The ec time was configured to 30s in manifest but the test network of 2k was using 4s block time. This missmatch meant that the calculated base timestamp was skewed 15X.
[...]

PR #899, opened 2025-03-07
Check for self messages before chain broadcast on get proposal

masih:
A node always attempts to get the proposal, for the instance it has not completed yet. This happens even if a node is restarted mid-instance. This is generally OK because the self-equivocation filter would drop any messages with potentially equivocating vote at the time of startup.

The chain broadcast, however, attempts to broadcast the chain as soon as calls to `GetProposal` are made. At that point, there's no message to be filtered through the self-propagation filter. There's just the chain.

To avoid broadcasting chains that are otherwise not useful to other participants, check if there's been any message from self for the instance for which proposal was retrieved. Then only broadcast the chain if there haven't been any prior messages from self, for instance.

Part of #893

### DIFF
--- a/host.go
+++ b/host.go
@@ -710,8 +710,16 @@ func (h *gpbftHost) RequestRebroadcast(instant gpbft.Instant) error {
 func (h *gpbftHost) GetProposal(instance uint64) (*gpbft.SupplementalData, *gpbft.ECChain, error) {
 	proposal, chain, err := h.inputs.GetProposal(h.runningCtx, instance)
 	if err == nil {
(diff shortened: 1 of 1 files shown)


Kubuxu (approved):

Kubuxu:
One note, this feels like wrong place for this. IMO, this function should be side-effect-free, but we can proceed with this.

masih:
> One note, this feels like wrong place for this. IMO, this function should be side-effect-free, but we can proceed with this.

Yeah I agree, i don't love it. but it's the earliest place where we can possibly publish the chain. open to suggestions :+1:

PR #918, opened 2025-03-14
feat: batch computation of ECChain Keys in chainexchange

### LINKED ISSUE #831: Expose the internal state of Merkle Tree to allow more efficient sub root digest calculation
As part of chain exchange, once we receive a chain the key for every subchain of it is calculated to speed up discovery, implemented [here](https://github.com/filecoin-project/go-f3/blob/d3d80771b0e4fda115eb9be0b25b4239caac4bef/chainexchange/pubsub.go#L233-L237). The current implementation uses the naive/simple approach by calling `merkle.Tree` for every sub chain. This effectively re-hashes the same information over and over for smaller sub-chains.

The digest calculation can be made much cheaper if the `merkle` implementation exposes internal state such that the intermediate state from calculating the longest sub-chain can be used to calculate the smaller sub-chain digests.

Kubuxu:
Adjust the for loop not to repeat the full chain
The EqualExportedValues is required as we never call explicit Key these
values, leaving cached key empty.

Resolves #831

### DIFF
--- a/gpbft/chain.go
+++ b/gpbft/chain.go
@@ -438,7 +438,7 @@ func (c *ECChain) KeysForPrefixes() []ECChainKey {
 
 	batch := merkle.BatchTree(values)
 	res := make([]ECChainKey, c.Len())
-	for i := 0; i < c.Len(); i++ {
+	for i := range c.Len() {
 		res[i] = batch[i]
 	}
 	return res
@@ -457,7 +457,7 @@ func (c *ECChain) AllPrefixes() []*ECChain {
 	batch := merkle.BatchTree(values)
 
 	res := make([]*ECChain, len(c.TipSets))
-	for i := 0; i < len(c.TipSets); i++ {
+	for i := range len(c.TipSets) {
 		var prefix ECChain
 		prefix.TipSets = c.TipSets[: i+1 : i+1]
 
--- a/gpbft/chain.go
+++ b/gpbft/chain.go
@@ -425,6 +425,51 @@ func (c *ECChain) Key() ECChainKey {
 	return c.key
 }
 
+// KeysForPrefixes return batch of keys for all prefixes
+// the indexes to this array correspond to Prefix(i) calls
+func (c *ECChain) KeysForPrefixes() []ECChainKey {
+	if c.IsZero() {
+		return nil
+	}
+	values := make([][]byte, c.Len())
+	for i, ts := range c.TipSets {
+		values[i] = ts.MarshalForSigning()
+	}
+
+	batch := merkle.BatchTree(values)
+	res := make([]ECChainKey, c.Len())
+	for i := 0; i < c.Len(); i++ {
+		res[i] = batch[i]
+	}
+	return res
+}
+
+// AllPrefixes returns an array of all prefix chain including the c itself.
+// It precomputes keys for them as well, populating the key cache.
+func (c *ECChain) AllPrefixes() []*ECChain {
+	if c.IsZero() {
+		return nil
+	}
+	values := make([][]byte, c.Len())
+	for i, ts := range c.TipSets {
+		values[i] = ts.MarshalForSigning()
+	}
+	batch := merkle.BatchTree(values)
+
+	res := make([]*ECChain, len(c.TipSets))
+	for i := 0; i < len(c.TipSets); i++ {
+		var prefix ECChain
+		prefix.TipSets = c.TipSets[: i+1 : i+1]
+
+		copy(prefix.key[:], batch[i][:]) // populate the key cache
+		prefix.keyLazyLoader.Do(func() {})
+
+		res[i] = &prefix
+	}
+
+	return res
+}
+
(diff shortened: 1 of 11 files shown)


gpbft/chain.go
```diff
@@ -425,6 +425,51 @@ func (c *ECChain) Key() ECChainKey {
 	return c.key
 }
 
+// KeysForPrefixes return batch of keys for all prefixes
+// the indexes to this array correspond to Prefix(i) calls
+func (c *ECChain) KeysForPrefixes() []ECChainKey {
```
masih, line 430: I guess no harm in keeping this even though it's no longer used.

gpbft/chain.go
```diff
@@ -425,6 +425,51 @@ func (c *ECChain) Key() ECChainKey {
...
+		return nil
+	}
+	values := make([][]byte, c.Len())
+	for i, ts := range c.TipSets {
+		values[i] = ts.MarshalForSigning()
+	}
+
+	batch := merkle.BatchTree(values)
+	res := make([]ECChainKey, c.Len())
+	for i := 0; i < c.Len(); i++ {
```
masih, line 441:
```suggestion
	for i := range c.Len() {
```

gpbft/chain.go
```diff
@@ -425,6 +425,51 @@ func (c *ECChain) Key() ECChainKey {
...
+		return nil
+	}
+	values := make([][]byte, c.Len())
+	for i, ts := range c.TipSets {
+		values[i] = ts.MarshalForSigning()
+	}
+	batch := merkle.BatchTree(values)
+
+	res := make([]*ECChain, len(c.TipSets))
+	for i := 0; i < len(c.TipSets); i++ {
```
masih, line 460:
```suggestion
	for i := range len(c.TipSets) {
```

gpbft/chain_test.go
```diff
@@ -189,6 +189,54 @@ func TestECChain(t *testing.T) {
...
+	chain := &gpbft.ECChain{TipSets: tipSets}
+	prefixes := chain.AllPrefixes()
+
+	require.Equal(t, len(prefixes), len(tipSets), "AllPrefixes should return a prefix for each tipset")
+
+	for i, prefix := range prefixes {
+		require.Equal(t, prefix.Len(), i+1, "Prefix %d should have length %d", i, i+1)
+		expected := chain.Prefix(i)
+		require.True(t, expected.Eq(prefix))
+		require.Equal(t, expected.Key(), prefix.Key())
```
masih, line 236:
```suggestion
		require.Equal(t, expected.Key(), prefix.Key())
		require.NotSame(t, prefix, expected)
```

host.go
```diff
@@ -500,6 +500,14 @@ func (h *gpbftRunner) BroadcastMessage(ctx context.Context, msg *gpbft.GMessage)
...
 	}
 
+	_, err = h.participant.ValidateMessage(msg)
+	if err != nil {
+		log.Errorf("full message validation failed: %+v", err)
+	}
+	_, err = h.pmv.PartiallyValidateMessage(pmsg)
+	if err != nil {
+		log.Errorf("partial message validation failed: %+v", err)
+	}
```
masih, line 510: We effectively get this from the pubsub right? in that all messages are validated by self before being published. The only thing is the footprint of this which should be minimal anyhow.

Double checking the above is more or less the train of thought here.

merkle/merkle_test.go
```diff
@@ -65,3 +67,88 @@ func TestHashZero(t *testing.T) {
...
+}
+
+func BenchmarkBatchPrefix(b *testing.B) {
+	K := 128
+	inputs := generateInputs("golden", K)
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		runtime.KeepAlive(BatchTree(inputs))
+	}
+}
```
masih, line 154:
```suggestion
Using `b.Run` makes it easier to read the benchmark comparison across the two.

```suggestion
func BenchmarkPrefixes(b *testing.B) {
	const K = 128
	inputs := generateInputs("golden", K)

	b.ResetTimer()
	b.Run("IndividualPrefix", func(b *testing.B) {
		for range b.N {
			for i := range len(inputs) {
				runtime.KeepAlive(Tree(inputs[:i+1]))
			}
		}
	})

	b.ResetTimer()
	b.Run("BatchPrefix", func(b *testing.B) {
		for range b.N {
			runtime.KeepAlive(BatchTree(inputs))
		}
	})
}
```
```

masih (approved):
Nice work @Kubuxu 🍻

gpbft/chain.go
Kubuxu: How did I miss for years that you can do `for i := range <int>`??

host.go
Kubuxu: ahh, this should have been deleted, it is leftover debugging code

gpbft/chain.go
masih: It's new (since 1.22).

Issue #728, opened 2024-10-28
Expose consensus critical metrics directly from gpbft


masih:
This repo provides `f3 observe` command to dump all messages exchanged as part of F3. One then would need to query them to answer questions like:
* Who is currently participating?
* What percentage of power is participating?

This is fine but not ideal. Because, gpbft already has all of this information along with the cache of all the validated messages. For alerting purposes alone it would be useful not to force operators to run an observer to get these metrics.

Kubuxu:
I'm not sure if the metrics I've added are good enough for it. They are useful for debugging and quick looks, but in most scenarios, they will cap out at 66% because when a quorum is reached, we will move on.

PR #924, opened 2025-03-19
Track participation in metrics for quality, commit, prepare and decide

Kubuxu:
Converge is more tricky, might get it done but this is better than what we
had before.

Caveat here is that in many cases we will only see 66% here as we will move on after quorum.

### DIFF
--- a/gpbft/gpbft.go
+++ b/gpbft/gpbft.go
@@ -15,6 +15,7 @@ import (
 	"github.com/filecoin-project/go-bitfield"
 	rlepluslazy "github.com/filecoin-project/go-bitfield/rle"
 	"github.com/ipfs/go-cid"
+	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/metric"
 )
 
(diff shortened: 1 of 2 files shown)


gpbft/gpbft.go
```diff
@@ -1074,7 +1078,7 @@ type chainSupport struct {
 }
 
 // Creates a new, empty quorum state.
-func newQuorumState(powerTable *PowerTable) *quorumState {
+func newQuorumState(powerTable *PowerTable, attributes ...attribute.KeyValue) *quorumState {
```
masih, line 1081: `attributes` is unused.

gpbft/gpbft.go
```diff
@@ -1118,6 +1122,11 @@ func (q *quorumState) receiveSender(sender ActorID) (int64, bool) {
 	q.senders[sender] = struct{}{}
 	senderPower, _ := q.powerTable.Get(sender)
 	q.sendersTotalPower += senderPower
+	if len(q.attributes) != 0 {
```
masih, line 1125: This is always 0 because we are not setting attributes field.

masih:
Fixes #https://github.com/filecoin-project/go-f3/issues/728

gpbft/gpbft.go
Kubuxu: oops

masih (approved):

PR #927, opened 2025-03-24
Remove the remaining mechanism for swappable payload marshal for signing

masih:
Previous versions of go-f3 made marshal for signing swappable during tests to help tests run faster. This has caused much confusion in debugging false negative test failures.

As a result, the fake marshal for signing was changed to use a format consistent with the real network (#923).

As a result we no longer need the types and mechanisms in code to make marshalling of payload for signing swappable. Remove them all.


Kubuxu:
I considered doing this but kept them in the gpbft to allow sim and emulation to run faster. I'm not against it, I'm just not sure about tradeoffs.

Kubuxu (approved):

masih:
> I'm just not sure about tradeoffs

Thanks for pointing that out.

Looking at CI run [on main](https://github.com/filecoin-project/go-f3/actions/runs/14033460708/job/39286082173?pr=927), vs. [this branch](https://github.com/filecoin-project/go-f3/actions/runs/14033460708/job/39286082173?pr=927) there is no difference in time it takes to run tests. This is because the host no longer short-cuts the signing ( [removed in #923](https://github.com/filecoin-project/go-f3/pull/923/files#diff-7717c52956f97079739dd32cd1d2c7a9ff80173a3895ac11d9e845d1b6c5ca41L844-L848)).

Aside from that, comparing CI run before #923  and after it does not show any increase in time to run tests.

So all and all, I think having fake payload marshal for signing didn't add as much value as we expected it to.

PR #963, opened 2025-05-02
Remove the ability to dynamically change F3 Manifest

masih:
Now that F3 is active on mainnet, remove all capabilities related to changing F3 manifest at runtime.

Adjust the code to _always_ demand a non-nil manifest and remove tests that asserted runtime manifest manipulation.

Additionally, remove redundant fields in the manifest that accepted explicit power config, ignoring EC power and pausing F3.

Part of https://github.com/filecoin-project/lotus/issues/12971


f3.go
```diff
@@ -39,13 +38,12 @@ type f3State struct {
 	ps       *powerstore.Store
 	certsub  *certexpoll.Subscriber
 	certserv *certexchange.Server
-	manifest *manifest.Manifest
 }
```
masih, line 41: I left the `f3State` behind to keep the internal pause/resume working - used solely for testing.

PR #969, opened 2025-05-03
Fix missing if not exist in observer table migration

masih:
Without it the db creation fails on service restart if db exists.


masih:
This PR is atop commit `b524461e54897ba0df26e2d5560cdf7508f8c92d` which has the observer fixes but before f3 cleanup

masih:
enabling container publish from it temporarily

Kubuxu (approved):

PR #972, opened 2025-05-03
Move partial message to its own package to avoid circular dependency

masih:
Just to make sure it is not breaking anything, PRed this separately. 

It moves the partial message types into a separate pacakge so that f3 root package can depend on observer otherwise -> circular dependency.


Kubuxu (approved):
SGTM but lint is complaining

PR #978, opened 2025-05-05
Implement an aider utility to help rebroadcast using observer

masih:
Implement a utility that:
* finds the latest instance and round
* reconstructs the list of all known messages that would have been re-broadcasted by the network from observer server.
* broadcasts it to the entire network.

...with configurable interval and degree of parallelism.

The aider additionally uses a list of give lotus nodes to find peers in the network. It then periodically connects to them to maximise its connectivity and therefore message propagation.


Kubuxu (approved):
SGTM apart from the dedup

Issue #989, opened 2025-05-07
Bubble up changes from `Silence of the Certs` to `main`


masih:
As part of investigating the mainnet issue re instance 6017 progressing very slowly we implemented a number of strategic approaches (see #971), most of which is appropriate to be bubbled up to main.

This issue is done when they are bubbled up to main.

The specific things to bubble up are:

- [x] https://github.com/filecoin-project/go-f3/pull/969
- [x] https://github.com/filecoin-project/go-f3/pull/984
- [x] https://github.com/filecoin-project/go-f3/pull/985
- [x] https://github.com/filecoin-project/go-f3/pull/987
- [x] https://github.com/filecoin-project/go-f3/pull/990
- [x] https://github.com/filecoin-project/go-f3/pull/991

Closes #971.

masih:
All the bubble up work now has a corresponding open PR that depend on eachother.
@Kubuxu  To avoid merge conflict please review the PR in the order listed in the issue description (ascending order of PR ID).

PR #990, opened 2025-05-07
Allow observer to be integrated into an existing pubsub

masih:
Allow observer to take pubsub as an option along with flag on whether to register validator or not.

---

* Originally implemented across #974 and #976
* Part of #989 
* **Review #987 first.**

### DIFF
--- a/cmd/f3/aider.go
+++ b/cmd/f3/aider.go
@@ -1,0 +1,440 @@
+package main
+
+import (
+	"bytes"
+	"encoding/json"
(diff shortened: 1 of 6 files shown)


Kubuxu:
This never worked as there can be only one subscription (from error logs on IDA).

masih:
Is there any harm to make it such that it can accept existing pubsub instance now that we have done the work?

Kubuxu (approved):

PR #991, opened 2025-05-07
Integrate DHT and Filecoin.NetPeers peer discovery into observer

masih:
Use Filecoin DHT and NetPeers Filecoin API call to discover new peers with a configurable connectivity threshold for each.

The observer now has three distinct ways to maintain connectivity:
* Via a fixed set of addresses (e.g. bootstrap peers list)
* Via Filecoin DHT bootstrap
* Via calling `Filecoin.NetPeers` on a list of given lotus daemons

Each can now have their own connectivity threshold below which additional peer discovery is attempted.

Additionally, clean up the ability to handle change in network name now that network name is final.

---
* Originally implemented in #981 
* Part of #989
* **Review #990 first.**


go.mod
```diff
@@ -14,7 +14,9 @@ require (
 	github.com/ipfs/go-log/v2 v2.5.1
 	github.com/klauspost/compress v1.17.11
 	github.com/libp2p/go-libp2p v0.37.2
+	github.com/libp2p/go-libp2p-kad-dht v0.25.2
```
Kubuxu, line 17: The deps here make me consider breaking out f3 command into its own sub-module.
On the other hand, dep trimming should solve it for most use cases.

masih: Yeah in later version of go this is a non-issue?

Dependency tree in this repo could be much thinner in anycase. Observer dependency to duckdb is another example.

PR #992, opened 2025-05-08
Implement lag awareness in aider based on current progress

### LINKED ISSUE #988: Make aider lag-aware
Change Aider to only trigger rebroadcast of messages when F3 is actually behind, or has entered CONVERGE. This way we can leave it running in the background and don't need to worry about it adding extra bandwidth to the network

masih:
Determine if F3 is lagging and if so engage rebroadcast aid with resilience against failure cases. The lag determination strategy uses both lotus daemon and observer as sources to determine the current F3 progress. It then decides if F3 is lagging behind based on:
 * number of rounds
 * distance from head
 * elapsed time since instance ID changed.
 * last successful lag determination

The combination of above assures that both in cases of failure or less than perfect information aid is attempted. The only case where aid can't be attempted is when no initial F3 state is determinable.

Fixes #988

### DIFF
--- a/cmd/f3/aider.go
+++ b/cmd/f3/aider.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"net/http"
 	"os"
+	"slices"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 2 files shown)


cmd/f3/aider.go
```diff
@@ -286,11 +285,130 @@ func aid(c *cli.Context, f3Chatter *pubsub.Topic) error {
...
+		// if round is non-zero.
+		return progress.Round > 0, progress
+	}
+
+	// There's live f3 progress fetched from lotus daemon. Conservatively pick the
+	// least progress and use it to determine if F3 is lagging.
+	slices.SortFunc(progresses, compareProgress)
+	leastProgress := progresses[0]
+
+	if leastProgress.Round > 0 {
```
Kubuxu, line 332: I would increase this, as I don't think aider can react fast enough to help with round 1 or 2 (they take under a minute).

Kubuxu (approved):
one comment, otherwise LGTM

Kubuxu (approved):

PR #998, opened 2025-05-20
Batch insert observed messages using Duckdb appender

### LINKED ISSUE #993: Optimise path to storing messages in observer
When the observer db gets large, the write path slows enough to demand increase in the size of subscription buffer size. Optimise it so that writing messages out is as fast as possible: 

* buffer in memory.
* write to some kind of WAL for crash resistence.
* concurrently dump out to parquet when a size threshold is met.

masih:
For fast batch insertion of messages observed into duckdb, use the native appender to batch the messages based on size or max wait time. Whichever happens first.

Fixes #993

### DIFF
--- a/observer/observer.go
+++ b/observer/observer.go
@@ -3,8 +3,8 @@ package observer
 import (
 	"context"
 	"database/sql"
+	"database/sql/driver"
 	_ "embed"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"net/http"
@@ -24,7 +24,7 @@ import (
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
 	record "github.com/libp2p/go-libp2p-record"
 	"github.com/libp2p/go-libp2p/core/peer"
-	_ "github.com/marcboeker/go-duckdb"
+	"github.com/marcboeker/go-duckdb"
 	"go.uber.org/multierr"
 	"golang.org/x/sync/errgroup"
 )
@@ -45,6 +45,12 @@ type Observer struct {
 
 	messageObserved chan *Message
 	msgEncoding     *encoding.ZSTD[*pmsg.PartialGMessage]
+
+	dbConnector           *duckdb.Connector
+	dbAppender            *duckdb.Appender
+	dbConnection          driver.Conn
+	unflushedMessageCount int
+	lastFlushedAt         time.Time
 }
 
 func New(o ...Option) (*Observer, error) {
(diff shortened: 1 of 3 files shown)


observer/observer.go
```diff
@@ -200,30 +222,35 @@ func (o *Observer) observe(ctx context.Context) error {
...
 		om.VoteValueKey,
 	); err != nil {
-		return fmt.Errorf("failed to execute query: %w", err)
+		return fmt.Errorf("failed to append row: %w", err)
+	}
+	o.unflushedMessageCount++
+	batchSizeReached := o.unflushedMessageCount >= o.maxBatchSize
+	batchDelayElapsed := !o.lastFlushedAt.IsZero() && time.Since(o.lastFlushedAt) >= o.maxBatchDelay
+	if batchSizeReached || batchDelayElapsed {
+		if err := o.dbAppender.Flush(); err != nil {
```
Kubuxu, line 248: Is `Flush` allowed to run concurrently?
I'm thinking, what if we run into low message count again, then the messages won't flush until another message lands.

masih: > Is Flush allowed to run concurrently?

I think it can't (godocs on that function are missing) but it is not being called concurrently in this implementation. 

> what if we run into low message count again, then the messages won't flush until another message lands
Yup good point. I'll adjust the channel to trigger flush on timer. This will be syncronous but would decouple flushing from arrival of messages. Thanks Kuba

Kubuxu (approved):

Kubuxu (changes requested):
timer won't work

observer/observer.go
```diff
@@ -170,13 +192,15 @@ func (o *Observer) createOrReplaceMessagesView(ctx context.Context, includeParqu
 
 func (o *Observer) observe(ctx context.Context) error {
 	rotation := time.NewTimer(o.rotateInterval)
+	flush := time.NewTimer(o.maxBatchDelay)
```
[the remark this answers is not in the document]
masih: ah man... I have fat fingered this. Sorry.

observer/observer.go
```diff
@@ -169,14 +191,16 @@ func (o *Observer) createOrReplaceMessagesView(ctx context.Context, includeParqu
 }
 
 func (o *Observer) observe(ctx context.Context) error {
-	rotation := time.NewTimer(o.rotateInterval)
+	rotation := time.NewTicker(o.rotateInterval)
+	flush := time.NewTicker(o.maxBatchDelay)
```
Kubuxu, line 195: Another note, this will lead to a maximum batch time of 2x maxBatchDelay, but I'm fine with that.

Kubuxu (approved):

PR #1007, opened 2025-05-26
Introduce single power table cache

Kubuxu:


ec/caching.go
```diff
@@ -0,0 +1,68 @@
...
+	// take semaphore
+	p.smaphore <- struct{}{}
+	defer func() { <-p.smaphore }()
+
+	res, err := p.Backend.GetPowerTable(ctx, tsk)
+	if err != nil {
+		return nil, fmt.Errorf("getting power table: %w", err)
+	}
+
+	p.cache.Add(string(tsk), res)
```
masih, line 66: What is the maximum tipset key size? I see cache size is fixed to 256. Is it worth hashing it before using as key?

f3.go
```diff
@@ -62,16 +63,19 @@ type F3 struct {
 // New creates and setups f3 with libp2p
 // The context is used for initialization not runtime.
 func New(_ctx context.Context, manifest manifest.Manifest, ds datastore.Datastore, h host.Host,
-	ps *pubsub.PubSub, verif gpbft.Verifier, ec ec.Backend, diskPath string) (*F3, error) {
+	ps *pubsub.PubSub, verif gpbft.Verifier, ecBackend ec.Backend, diskPath string) (*F3, error) {
 	runningCtx, cancel := context.WithCancel(context.WithoutCancel(_ctx))
+
+	// concurrency is limited to half of the number of CPUs, and cache size is set to 256 which should be enough for most use cases
```
masih, line 69: Why 256? Because we don't expect that much diversity in tipset keys within a hand full of instances? 

I can imagine a future reader to question what "most use cases" means.

internal/singleflight/singleflight.go
```diff
@@ -0,0 +1,244 @@
...
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+// (below is the original license)
+//
+// Copyright 2009 The Go Authors.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+
```
masih, line 11:
```suggestion
//
```

internal/singleflight/singleflight.go
```diff
@@ -0,0 +1,244 @@
...
+//    * Redistributions of source code must retain the above copyright
+// notice, this list of conditions and the following disclaimer.
+//    * Redistributions in binary form must reproduce the above
+// copyright notice, this list of conditions and the following disclaimer
+// in the documentation and/or other materials provided with the
+// distribution.
+//    * Neither the name of Google LLC nor the names of its
+// contributors may be used to endorse or promote products derived from
+// this software without specific prior written permission.
+
```
masih, line 21:
```suggestion
//
```

internal/singleflight/singleflight.go
```diff
@@ -0,0 +1,244 @@
...
+// A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+// OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+// LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+// Modified 2025 by Filecoin Project to use Generics.
+//
```
masih, line 34:
```suggestion

```

These suggestions should make the godoc show up correctly.

internal/singleflight/singleflight.go
```diff
@@ -0,0 +1,244 @@
+// Copyright 2013 The Go Authors. All rights reserved.
```
masih, line 1: Thinking out loud:
* Alternatively, move the license to LISENSE.md in root of the package? 
* If BSD is compatible with the go-f3 license I think we could skip this notice entirely maybe?

Not sure what the rules are.

internal/singleflight/singleflight.go
```diff
@@ -0,0 +1,244 @@
...
+// LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+// Modified 2025 by Filecoin Project to use Generics.
+//
+// Package singleflight provides a duplicate function call suppression
+// mechanism.
+package singleflight
```
masih, line 37: I am curious, why fork code instead of using `golang.org/x/sync` as dependency?

masih (commented):
No blockers, left a few questions.

internal/singleflight/singleflight.go
Kubuxu: I've added generics support

masih: I don't have any strong opinions on this one. FWIW i would avoid this in favour of less LOC in this repo considering the generics support added here is only ever used for one type.

Kubuxu: Moved to standard, you are right, it was not worth the mess

masih (approved):
LGTM apart from a question about max tipset key size 👍

Kubuxu:
Addressed in a commend (I think)

ec/caching.go
Kubuxu: I don't think it is worth it, the hard cap is 15*(32+8), expected it 5*(32+8).

PR #1015, opened 2025-06-12
Retract 0.8.5 release due to #1014

Kubuxu:
This commit retracts 0.8.5 (a brand new release) due to anomaly in
metrics caused by #1010 (reverted in #1014).
The 0.8.5 release is not used by any implementation at the time of
retraction.
Follow up release 0.8.6 is on its way.


Kubuxu:
We won't retract the release, it contains a regression but it is not critical.

PR #1017, opened 2025-06-12
fix: signature payload determination for partial messages

masih:
Use the provided value key for signature payload when message is partial.


BigLep:
Should we add a unit/regression test?

PR #1024, opened 2025-06-20
Infer messages from evidence of justification

### LINKED ISSUE #342: Treat reception of justifications as if reception of the messages they evidence
A vote justification carries evidence of a strong quorum of nodes all having sent the same message. If a participant receives a justification but has not received those messages, it can treat the justification as those messages. For example, if a node is stuck in PREPARE but receives a COMMIT message, it should be able to use the justification from that COMMIT to immediately complete the PREPARE phase.

masih:
The justification accompanied by most messages in GPBFT is evidence of a strong quorum of some states across the participants. Treat the receipt of such justifications as the receipt of the messages that justified it.

This allows a faster progress towards quorum in the network, and reduces reliance on re-broadcast and majority network state propagation.

Specifically, there are 2 phases that can benefit from the above optimization:

1. PREPARE phase, using justification of PREPARE in:
  * COMMIT messages at the same round.
  * PREPARE messages at the next round.
  * CONVERGE messages at the next round.
2. COMMIT phase, using justification of COMMIT to bottom in:
  * CONVERGE messages at the next round.
  * PREPARE messages at the next round.

The above optimization doesn't matter in CONVERGE and DECIDE phases. Because, CONVERGE by design awaits the phase timeout, and DECIDE is already optimized by `skipToDecide` where the participant rightly so immediately progresses to the final phase using the justification of COMMIT accompanied by the DECIDE message.

Fixes #342

### DIFF
--- a/gpbft/gpbft.go
+++ b/gpbft/gpbft.go
@@ -272,7 +272,6 @@ func (i *instance) receiveOne(msg *GMessage) (bool, error) {
 		if msg.Justification != nil {
 			msgRound.prepared.ReceiveJustification(msg.Vote.Value, msg.Justification)
 		}
-
 	case COMMIT_PHASE:
 		msgRound.committed.Receive(msg.Sender, msg.Vote.Value, msg.Signature)
 		// The only justifications that need to be stored for future propagation are for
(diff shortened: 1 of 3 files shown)


gpbft/gpbft.go
```diff
@@ -528,9 +548,18 @@ func (i *instance) beginCommit() {
...
 	var justification *Justification
 	if !i.value.IsZero() {
-		if quorum, ok := i.getRound(i.current.Round).prepared.FindStrongQuorumFor(i.value.Key()); ok {
+		valueKey := i.value.Key()
+		currentRound := i.getRound(i.current.Round)
+		nextRound := i.getRound(i.current.Round + 1)
+		if quorum, ok := currentRound.prepared.FindStrongQuorumFor(valueKey); ok {
 			// Found a strong quorum of PREPARE, build the justification for it.
 			justification = i.buildJustification(quorum, i.current.Round, PREPARE_PHASE, i.value)
+		} else if justifiedByCommit := currentRound.committed.GetJustificationOf(PREPARE_PHASE, valueKey); justifiedByCommit != nil {
```
masih, line 557: We are refetching justifications here which is aligned with the current pattern of method call in core gpbft package. 

I ideally we should pass in a justification to `beginCommit` to avoid looking it up twice. (similar story for prepare).

But for now I have kept the diff small in favour of getting the logic correct.

Kubuxu (approved):

PR #1042, opened 2025-07-22
Add observer metrics

Kubuxu:


Kubuxu:
@rvagg I would appriciate a post-review, I'm going to merge this when CI passes.

PR #1071, opened 2026-03-18
chore(deps): bump kyber to a stable version

hanabi1224:
This PR tries to bump kyber to a stable version


Kubuxu:
Gnark is integrated into Kyber now, so we can drop the internal gnark backend.

PR #1080, opened 2026-05-15
fix(f3): throttle delayed bootstrap retries

### LINKED ISSUE #1079: Silence log-spam on new node sync
Logs full of this, ticking by at a rapid rate, when syncing from an older point, in this case from late last year which I think was after bootstrap wasn't it? Can we log less frequently than this?

Image

Kubuxu:
When wall-clock time has passed bootstrap but the observed EC head is still behind, retry bootstrap less frequently while the head remains far from the bootstrap epoch.

Fixes #1079


rvagg (approved):
okie dokie 🤞 I'll admit to not being familiar with this code

